PR #54, opened 2016-04-18
Remove executable permission from non-executable content (docs, json, and js without shebang)

yarikoptic:


isTravis:
Thank you for this! Apologies it took so long to accept the PR. We really appreciate your contribution!

yarikoptic:
The least I could do. Thanks!

PR #92, opened 2016-07-12
Jupyter integrations

hassanshaikley:


isTravis:
This pull request is set to merge into `master`. Could you resubmit it with the base being `architecture2` instead?

PR #141, opened 2016-08-04
List followers on user profiles

hassanshaikley:
This is the way that profiles are styled:

screen shot 2016-08-04 at 2 00 15 pm

I will be pushing more to this branch, so I don't imagine that it matters if you want to pull it in now or later.  : )

Also we must have lost some of your followers upon migration, you may want to look into that.

### DIFF
--- a/api/routes/atom-routes.js
+++ b/api/routes/atom-routes.js
@@ -295,6 +295,15 @@ export function getAtomData(req, res) {
 			}
 		});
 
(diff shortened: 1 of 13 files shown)


hassanshaikley:
screen shot 2016-08-04 at 3 03 49 pm

hassanshaikley:
screen shot 2016-08-04 at 3 29 45 pm

PR #166, opened 2016-08-11
Added Code atom

joeltg:
Code atom! Code splitting coming soon.


isTravis:
Will keep around for future reference of codeMirror in ProseMirror - but architectures are out of date now.

PR #175, opened 2016-08-29
Adding more formatted messages

hassanshaikley:
Let me know that this is okay!


hassanshaikley:
screen shot 2016-08-29 at 4 48 02 pm

The bottom segment are your changes, right? @isTravis

PR #195, opened 2016-10-12
iFrame resize client side snippet added to client.js

hassanshaikley:

PR #196, opened 2016-10-17
Linting Fix: https://github.com/pubpub/pubpub/pull/195 

hassanshaikley:
Error replaced with warning.  : )


isTravis:
I think a cleaner fix might be reverting to the 'error' - and then putting a lint exception at the top of the file. Something like `/* eslint no-unused-expressions: 0 */`

[Eslint](http://eslint.org/docs/rules/no-unused-expressions) documentation here might help.

PR #300, opened 2018-12-18
update: Fix bug that showed manage when logged out.

gabestein:


client/components/Header/Header.js
```diff
@@ -146,9 +146,11 @@ class Header extends Component {
 									{!isBasePubPub &&
 										<a href="/search" role="button" tabIndex="0" className="hide-on-mobile bp3-button bp3-large bp3-minimal">Search</a>
 									}
-									{isAdmin && isPage
-										? <a href={`/dashboard/pages/${this.props.locationData.params.slug || ''}`} className="bp3-button bp3-large bp3-minimal">Manage</a>
-										: <a href="/dashboard" className="bp3-button bp3-large bp3-minimal">Manage</a>
+									{isAdmin && isPage &&
+										<a href={`/dashboard/pages/${this.props.locationData.params.slug || ''}`} className="bp3-button bp3-large bp3-minimal">Manage</a>
```
gabestein: btw -- did you mean to test for the homepage as well? Originally I had it so that while on the homepage it didn't go to manage the page, but the entire community, since it's the default URL people visit. 

I can see it going either way, but lean towards going to base manage rather than homepage.

isTravis: Oh hmm - I did mean to test for the homepage. I was being more general and just saying all Pages bring you to manage their content. I don't have a strong preference though - so happy to go with it taking you to manage/pubs.

PR #306, opened 2019-02-19
fix: update all broken Storybook entries

idreyn:
This PR fixes various React Storybook entries that were previously failing due mostly to missing props. I've added the props where appropriate and I'm looking into a way to automatically test that all fixtures mount and render, either as a pre-push hook or a CI step, to make sure this doesn't happen again.

_Test plan:_

Run `npm run storybook` and visit `localhost:9001`. Check that every story renders as expected.

### DIFF
--- a/client/components/PubHeader/PubHeader.js
+++ b/client/components/PubHeader/PubHeader.js
@@ -57,7 +57,7 @@ class PubHeader extends Component {
 	}
 
 	componentWillUnmount() {
-		this.cleanupSticky();
+		this.stickyInstance.cleanup();
 		window.removeEventListener('resize', this.handleResize);
 	}
 
@@ -65,14 +65,8 @@ class PubHeader extends Component {
 		const nextOffsetHeight = this.headerRef.current.offsetHeight;
 		if (nextOffsetHeight !== this.offsetHeight) {
 			this.offsetHeight = nextOffsetHeight;
-			this.cleanupSticky();
-			this.stickyInstance = stickybits('.pub-header-component', { stickyBitStickyOffset: 35 - this.offsetHeight, useStickyClasses: true });
-		}
-	}
-
-	cleanupSticky() {
-		if (this.stickyInstance && typeof this.stickyInstance.cleanup === "function") {
 			this.stickyInstance.cleanup();
+			this.stickyInstance = stickybits('.pub-header-component', { stickyBitStickyOffset: 35 - this.offsetHeight, useStickyClasses: true });
 		}
 	}
 
--- a/client/components/PubDraftHeader/PubDraftHeader.js
+++ b/client/components/PubDraftHeader/PubDraftHeader.js
@@ -32,7 +32,7 @@ class PubDraftHeader extends Component {
 	}
 
 	componentWillUnmount() {
-		this.stickyInstance.cleanUp();
+		this.stickyInstance.cleanup();
 	}
 
 	render() {
--- a/searchSync/init.js
+++ b/searchSync/init.js
@@ -1,3 +1,3 @@
-require('babel-register');
+require('@babel/register');
 // require('./migrate-v5.js');
 require('./searchSync.js');
--- a/stories/components/footerStories.js
+++ b/stories/components/footerStories.js
@@ -10,9 +10,9 @@ import { accentDataDark, accentDataLight, communityData } from '../data';
 const wrapperStyle = { margin: '1em 0em' };
 
 const customSocialItems = [
-	{ id: 'si-0', icon: <Icon icon="vimeo" />, title: "Website", value: "custom", url: "https://vimeo.com/custom" },
-	{ id: 'si-1', icon: <Icon icon="soundcloud" />, title: "Check out my SoundCloud", value: "custom", url: "https://twitter.com/custom" },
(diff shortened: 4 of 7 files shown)


client/components/PubHeader/PubHeader.js
```diff
@@ -57,19 +57,25 @@ class PubHeader extends Component {
...
 		const nextOffsetHeight = this.headerRef.current.offsetHeight;
 		if (nextOffsetHeight !== this.offsetHeight) {
 			this.offsetHeight = nextOffsetHeight;
-			this.stickyInstance.cleanup();
+			this.cleanupSticky();
 			this.stickyInstance = stickybits('.pub-header-component', { stickyBitStickyOffset: 35 - this.offsetHeight, useStickyClasses: true });
 		}
 	}
 
+	cleanupSticky() {
```
isTravis: For simple existence bugs like this, I tend to fix them by setting a default rather than a check. So, in the constructor, instead of `this.stickyInstance = undefined`, we could simply do `this.stickyInstance = { cleanup: ()=>{} }`. Obviously for more complex or deeply-nested existence issues, a check is simpler. In this case though, the one-line of default-setting feels simpler to reason about to me. Do you have thoughts or best-practice insight otherwise? (In the early days, I think there will be a lot of figuring out what is my personal quirk, and what is actually a good technical idea - so always feel free to give counter arguments!)

Also, there is a bit of storybook weirdness going on. I don't think what you did here actually solved the bug - but it did solve the problem. In storybook, pubHeader does indeed render with these changes - but the bug itself seems to come from pubDraftHeader unmounting.

screen shot 2019-02-20 at 8 08 58 am

The bug is actually a typo in pubDraftHeader.js. `.cleanUp()` should be `.cleanup()`.

One last note on this: if we did use the `cleanupSticky()` function, we would have to initialize that from within the constructor: `this.cleanupSticky = this.cleanupSticky.bind(this)`. Initializing these functions is a bit verbose - but last I read up on it, the alternatives felt messier and more easily bug-ridden. Again - open to feedback on alternatives though.

isTravis:
For commit messages, we try to prepend the message with `fix`, `feature`, `update`, or `dev`. More context is [in the readme](https://github.com/pubpub/pubpub#commits)

isTravis:
I pushed up [a commit](https://github.com/pubpub/pubpub/pull/306/commits/e3fffd2096f72b62c0cb175ab2de7608f970d9dd) with some linting and a change to pubDraftHeader - feel free to take a look at that.

client/components/PubHeader/PubHeader.js
idreyn: >  I pushed up [a commit](https://github.com/pubpub/pubpub/pull/306/commits/e3fffd2096f72b62c0cb175ab2de7608f970d9dd) with some linting and a change to pubDraftHeader - feel free to take a look at that.

That seems good to me. Sorry about the `cleanup`/`cleanUp` confusion! A few thoughts I have based on your comments:

- It would be nice to make this kind of bug impossible by using a static typechecker — either Flow or Typescript.
- I'm ±0 on the idea of default dummy objects like `{cleanup: () => {}}`; I find them helpful, but often difficult to maintain in the long term!
- I don't think it should be necessary to manually `bind` a class method as described unless you're planning on calling it outside of the class, e.g. by passing it as a callback prop to a React element in `render`.

idreyn: As an aside, it looks like certain code in `pubpub` is already using the [`@babel/plugin-proposal-class-properties`](https://babeljs.io/docs/en/next/babel-plugin-proposal-class-properties.html) transformer. In that case, I think we can start using arrow functions to replace the use of `bind` entirely, i.e. instead of:

```
class Test extends React.Component {
    constructor(props) {
        super(props);
        this.handleClick = this.handleClick.bind(this);
    }

    handleClick() {
        console.log("oh no, I am clicked");
    }

    render() {
        return <button onClick={this.handleClick} />;
    }
}
```

we could use arrow functions to represent event handlers:

```
class Test extends React.Component {
    handleClick = () => {
        console.log("oh no, I am clicked");
    }

    render() {
        return <button onClick={this.handleClick} />;
    }
}
```

since arrow functions use lexical scoping for `this`, it will always refer to the class instance and we can avoid doing the binding. Last I checked (this morning) this feature isn't standard ES, but it's pretty far down the standards track and I feel pretty comfortable making use of it.

isTravis: Right - you only need to bind in the constructor if you're going to be calling `this` within the function and want it to refer to the Component scope. 

We used to use arrow functions as you described above, but we switched to constructor binding because:
1) I thought the arrow functions looked a bit ugly in component functions, especially when they had props
```
class Test extends React.Component {
    handleClick = (event, index, scope) => {
        console.log("oh no, I am clicked");
    }
}
```
In other contexts, I'm 100% on the arrow function train and our linter even enforces them over `function() {`.
2) something felt nice about having all the functions declared in the constructor as a means of communicating what functions exist within a component. This is less of an issue for small, simple components - but I do find nice with larger components like the Pub.js container.

Both these reasons feel a bit hand wavy - so I'm not against moving towards arrow binding for component functions. Let's continue offline - errr.... online on Slack :) - since I'm going to merge this PR now.

PR #307, opened 2019-02-20
dev: add Jest testing framework to automatically check Storybook fixtures

idreyn:
This PR adds the frontend testing framework Jest as a mechanism to automatically check that our Storybook entries render as a sort of unit test. In the future, we'll be able to use Jest to write more frontend tests. We'll also be able to use this mechanism to perform snapshot testing on components, if that's something we decide we want. For now, this is simple and lightweight — just run `npm test` and it'll tell you whether all the stories render correctly or not.

In the future I'd like to create a git pre-push hook or a CI step to run this automatically; I'll save that for another PR.

_Test plan:_
Run `npm test` and verify that all stories render. Now try messing with the render function of a component that we test with Storybook — maybe throw an error from `render()` or something, and verify that `npm test` complains about that component.


idreyn:
Excuse the force-pushing, I'm still getting used to a git-based review workflow.

isTravis:
Given the huge number of changes from prettier yesterday, could you update this PR to play nicely with the current state of `dev`?

idreyn:
> Given the huge number of changes from prettier yesterday, could you update this PR to play nicely with the current state of `dev`?

Done!

isTravis:
Just to recap from the end of day Friday, the two pending things on this PR are:

1. Moving defaultProps out of PubOptionsSaveVersion component
2. Testing if enzyme and enzyme-adapter-react-16 can be moved into devDependencies

idreyn:
> Testing if enzyme and enzyme-adapter-react-16 can be moved into devDependencies

Looks like I'm not getting that lint error anymore, which is likely due to the upstream version bump of eslint or the changes to the enabled rules. So these can be moved into `devDependencies` — thanks for pushing on this.

PR #314, opened 2019-02-28
update: Move cursor to end of selection after creating a link

idreyn:
This PR modifies the way the link creation flow works in the pub editor: after selecting a block of text, adding a link to it, and hitting enter, your cursor will be moved to the end of the new link. Previously, the link would remain highlighted, necessitating an extra key press.

I'd like all of this link manipulation logic to be moved out of `Pub.js` and into something a little more specific. But I'm going to wait until after our impending refactor of that area to tackle this.

_Test plan:_
Add some text to a pub. Highlight part of it, add a link, and verify that after dismissing the link popup, your cursor is at the end of the new link.

### DIFF
--- a/client/containers/Pub/Pub.js
+++ b/client/containers/Pub/Pub.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 /* eslint-disable-next-line import/no-extraneous-dependencies */
 import firebase from '@firebase/app';
 import checkIfMobile from 'is-mobile';
+import { TextSelection } from 'prosemirror-state';
 // import applyDevTools from 'prosemirror-dev-tools';
 import PageWrapper from 'components/PageWrapper/PageWrapper';
 import PubHeader from 'components/PubHeader/PubHeader';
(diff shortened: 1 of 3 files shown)


isTravis:
I agree that this should be outside of Pub.js. One separation I've kept is that anything that needs to use ProseMirror should happen in pubpub-editor. I think I feel the same about this functionality. For example, in addition to `<Editor />`, `@pubpub/editor` exports a series of utility functions. These functions are [defined here](https://github.com/pubpub/pubpub-editor/blob/master/src/utilities.js).

I think I'd prefer to see this link-management happen by `Pub.js` calling a function like

```
import Editor, { moveCursor} from '@pubpub/editor`
...

this.setState({ linkPopupIsOpen: false }, () => {	
	const { view } = this.state.editorChangeObject;
	moveCursor(view, 'end');
});

```
Perhaps `moveCursor` takes params to move the cursor the the start or end of selection, forward 1, backward 1, forward n, start of doc, end of doc, etc.

idreyn:
I was thinking along different lines — I think it's okay to do cursor manipulation outside of `pubpub-editor`, but I would like the manipulation here (and really everything related to the link popup and its state) to be handled by something other than `Pub` — probably closer to `PubBody`.

If you want, I can take a swing at your proposed abstraction here, but in any case we don't always want to move the cursor to the end of the doc, since the highlighted text might not be there.

isTravis:
++

To clarify though, I meant the `moveCursor` function could take options that determine how the cursor moves. Options could include:
- Start/End of doc
- Start/End of selection
- Forward/Backward N

PR #318, opened 2019-03-04
fix: Correctly position remove button on tags

idreyn:
This PR fixes an issue where the remove/"x" button on tags was not being correctly positioned due to what looks like an extraneous `.bp3-` Blueprint CSS class that was being applied to one of its parents. AFAIK the parent element with that class is safe to remove entirely, and I think the general takeaway is that we should leave the use of Blueprint CSS classes to the Blueprint React API wherever possible.

_Before:_
image
_After:_
image

_Test plan:_
Find a place where we use a `TagMultiSelect` component — the "Filter by tags" entry in the layout editor at `/dashboard/pages/` is as good as any — and verify that added tags have their close button positioned correctly.

### DIFF
--- a/client/components/LayoutEditorPubs/LayoutEditorPubs.js
+++ b/client/components/LayoutEditorPubs/LayoutEditorPubs.js
@@ -209,25 +209,23 @@ class LayoutEditorPubs extends Component {
 						onChange={this.changeTitle}
 					/>
 					<InputField label="Filter by Tag">
-						<div className="bp3-button-group bp3-select">
-							<TagMultiSelect
-								allTags={this.props.communityData.tags}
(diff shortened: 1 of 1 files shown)


isTravis:
Awesome - nice find. I spent some time searching through the Chrome debugger when you said that it was a bit hairy trying to find the source of this CSS bug, and couldn't find it. 

100% agree that we should avoid using `bp3-` in all cases. A lot of the use of it is legacy where blueprint used to require `pt-` classes for a lot of things, and then when they upgraded to 3.x, they had a migration script that converted `pt-` to `bp3-`. So, it should be mostly legacy code that we should move from when we spot it.

PR #345, opened 2019-05-24
feature: Attribution "affiliation" field

idreyn:
This pull request adds the `affiliation` field to the `PubAttribution` and `CollectionAttribution` models, allowing us to set an affiliation for contributors to pubs and collections, and submit that information to Crossref. To users, we expose a field like this:

image

Right now it's just a freeform string — so you could affiliate yourself with the Chattanooga School of Clowning Arts if you want. Eventually we'll want to start collecting affiliations at the `User` level, linking to ORCID, etc. — but it's a start.

_Test plan:_
After running `npm migrate && npm start` I observed the following:
- Both the collection and pub-level attribution editors let you see and update affiliation for each author
- Submitting DOI information for both a collection and a pub causes the affiliation to appear for each author in the deposit generated at `test.crossref.org`.

(In theory we could update our CitationJS code to use this as well, but from a quick skim of the documentation I'm not sure if they support this/this is something we'd include here — @isTravis feel free to weigh in)

### DIFF
--- a/client/components/AttributionEditor/AttributionRow.js
+++ b/client/components/AttributionEditor/AttributionRow.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-import { Checkbox, Icon, Position } from '@blueprintjs/core';
+import { Checkbox, Icon, InputGroup, Position } from '@blueprintjs/core';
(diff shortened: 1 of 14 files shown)


isTravis:
Great - this is looking good and will get us what we need to start on the v6 PubMeta Details panel. 👍

PR #372, opened 2019-05-30
update: Restore collab avatars in v6

idreyn:
As in v5 we would like to show the avatars of other collaborators currently editing the same pub branch. Most of the code is in place to do this is already left over from v5 — I've just had to make some tweaks to make it understand branches, and do a bit of cleanup along the way. Here it is in action:
Screen Shot 2019-05-30 at 5 35 49 PM

_Test plan:_
Log in as two different PubPub users and open the same pub. Then verify that:
- Users do not see themselves among the collaborators
- Collaborator avatars appear in the draft editing bar as expected
- Cursor colors in the editor match the color rings around the avatars
- Custom profile images are shown for users that have them
- Closing a user's editor causes their bubble to disappear

### DIFF
--- a/client/utils/firebaseClient.js
+++ b/client/utils/firebaseClient.js
@@ -26,5 +26,3 @@ export const initFirebase = (rootKey, authToken) => {
 			console.error('Error authenticating firebase', err);
 		});
 };
-
-export const y = 6;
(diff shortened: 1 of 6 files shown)


client/utils/firebaseClient.js
```diff
@@ -26,5 +26,3 @@ export const initFirebase = (rootKey, authToken) => {
 			console.error('Error authenticating firebase', err);
 		});
 };
-
-export const y = 6;
```
isTravis: 😭

Issue #385, opened 2019-06-06
Include community title or primary collection title in page title


gabestein:
**Who requested**
https://twitter.com/judell/status/1136381509932240896

**Stories**
- As a scraper (spider, annotation tool, etc.), I want the community or collection title to appear in the page title, so I can give more context to users about what collection the pub belongs in.

**Implementation Notes & Context**
- For example, [this pub](https://bookbook.pubpub.org/pub/926svib6) just says `Chapter 1` in the title. There's no way to know what community or collection the Pub is in just from looking at it. This has implications for SEO, for social sharing, annotation tools, and for users looking for Pubs in browser tabs.
- There are probably other places we need to provide this, such as RSS feeds.

**Technical Specification**
- If the pub has a primary collection set that's not a journal, use the title of that collection; else, use the title of the community.

**Mockups or reference images**

judell:
Hi Gabe, we meet again!

From a Hypothesis perspective (but also broadly relevant) a critical thing would be to include a work-specific ID in the URL.

Thanks for picking up on this.

gabestein:
👋 again.

Thanks, that's really helpful. Does the pub slug at the end of the url work as a unique id? Right now it can be overridden with a pretty permalink, but we're ultimately planning to move to a world where that slug is always in the url.

ie: [https://bookbook.pubpub.org/pub/**926svib6**](https://bookbook.pubpub.org/pub/926svib6)

judell:
That appears to be a per-chapter ID. If you're Remi, and you want to review all the annotations on your book, you'd need to do a per-chapter search. You'd rather do a book-wide search to find annotations on all chapters.

xolotl:
Also, if that ID ever changes (eg, versioning?), it would be good to think through how annotations would aggregate around/despite those changes.

gabestein:
> That appears to be a per-chapter ID. If you're Remi, and you want to review all the annotations on your book, you'd need to do a per-chapter search. You'd rather do a book-wide search to find annotations on all chapters.

Ah, I see, thanks. I'm not sure exactly how we'll approach that, but worth thinking about -- potentially we could append an id for the collection to the URL. We'll add it to the hopper.

> Also, if that ID ever changes (eg, versioning?), it would be good to think through how annotations would aggregate around/despite those changes.

[...]

PR #403, opened 2019-06-10
Pub page metadata and titles

gabestein:
Addresses #385.

If the pub is not in a primary collection or the pub's primary collection is of kind 'issue', the page title becomes: `{pubTitle} • {communityTitle}`

Otherwise, the pub is in a primary collection that is a book or conference, in which case the page title becomes: `{pubTitle} • {primaryCollection.title}`

Additionally, it appears that page metadata (title, og, twitter card, etc.) had regressed in V6. This attempts to restore this metadata, and makes modifications such that citation titles do not include community or collection titles.

One open question to resolve: what to do with unlisted in generateMetaComponents in the new pub model (server/routes/pub.js#99).

### DIFF
--- a/server/routes/pub.js
+++ b/server/routes/pub.js
@@ -73,6 +73,14 @@ app.get(
 						mode: mode,
 					},
 				};
+				let primaryCollection = false;
+				pubData.collectionPubs.forEach((collection) => {
+					if (collection.isPrimary && collection.collection.kind !== 'issue') {
+						primaryCollection = collection.collection;
+					}
+				});
+				const collectionTitle = primaryCollection.title || initialData.communityData.title;
+				const pubTitle = `${pubData.title} · ${collectionTitle}`;
 				return renderToNodeStream(
 					res,
 					<Html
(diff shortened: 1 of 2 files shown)


server/routes/pub.js
```diff
@@ -73,13 +73,30 @@ app.get(
...
 							initialData: initialData,
+							title: pubTitle,
+							citationTitle: pubData.title,
+							collectionTitle: collectionTitle,
+							description: pubData.description,
+							image: pubData.avatar,
+							attributions: pubData.attributions,
+							publishedAt: pubData.firstPublishedAt,
+							doi: pubData.doi,
+							//unlisted: isUnlistedDraft,
```
gabestein: Was unsure how unlisted interacts with the V6 pub model.

gabestein:
It occurs to me that at some point, we may want to store data about what is the pub's canonical title for citations and other purposes, and what is its display title higher up in the pub model for consistency across search, deposits, etc. For now, this solves the issue just at the pub level.

server/routes/pub.js
```diff
@@ -73,13 +73,30 @@ app.get(
 						mode: mode,
 					},
 				};
+				let primaryCollection = false;
+				pubData.collectionPubs.forEach((collection) => {
+					if (collection.isPrimary && collection.collection.kind !== 'issue') {
+						primaryCollection = collection.collection;
+					}
+				});
+				const collectionTitle = primaryCollection.title || initialData.communityData.title;
```
isTravis: In the case where there is no primaryCollection, calling `primaryCollection.title` (i.e. `false.title`) will throw an error. I'll add a commit where we use a reduce instead to handle this case.

PR #404, opened 2019-06-10
update: Add link to Pub from collection DND pane

gabestein:
This is just a thing I remembered a few folks have requested. Seems to play just fine with DND. Feel free to tell me to get lost if now's not a good time for it, but I was there, remembered it, had a second, so why the heck not.

### DIFF
--- a/client/containers/Dashboard/DashboardContent/Collection/CollectionEditor/PubRow.js
+++ b/client/containers/Dashboard/DashboardContent/Collection/CollectionEditor/PubRow.js
@@ -35,7 +35,9 @@ const PubRow = (props) => {
 			)}
 			<div className="contents">
 				<div className="info">
-					<div>{pub.title}</div>
+					<div>
+						<a href={`/pub/${pub.slug}`}>{pub.title}</a>
+					</div>
 					<div className="bp3-text-muted">
 						<em>{authorsNamesFromPub(pub).join(', ')}</em>
 					</div>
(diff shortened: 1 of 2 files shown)


isTravis:
I pushed up a commit to only show the links on the pubs that have been selected. Before this commit, we would have an interface like so:

Screen Shot 2019-06-10 at 5 26 16 PM


It doesn't screenshot my cursor (drats) - but when hovering a pubRow on the left column, it gives you a pointer cursor rather than a drag cursor. After the commit, only the pubs in the right hand side are links. I think this makes the `text-decoration: underline` of an <a /> more appropriately used too.

Screen Shot 2019-06-10 at 5 27 10 PM

gabestein:
I thought about this, but having the links in the left filter pane is more important than having it on the right. The need for this is in deciding what pub to add to the collection. I usually filter with the search, then want to click into the pub to make sure it's the one I want to add to the collection before moving it over, especially because saving is immediate.

What about: overriding the cursor in css or adding a link icon instead of linking from the title?

isTravis:
Hmm - good point. In that case, a link icon/button feels like the right solution. My natural response was to mouse over to the title to click-and-drag, and having that accidentally send me to another page feels like it could be disorienting.

gabestein:
Ok, a few design thoughts/Qs. I like the idea of the link icon going away and being replaced with the title when you move it to the right side. Otherwise, it starts to look like a ton of icons -- and this provides some nice visual separation between pubs not in the collection and pubs in the collection which was previously missing.

But the icon...I cannot figure out what I want to do here. It's just so much icon. So, four options in this screenshot: 

1. small minimal anchor button
1. regular link normal text color
1. regular link #666
1. regular link lightskyblue

Screen Shot 2019-06-10 at 6 41 50 PM

Opinions? Other options to consider?

idreyn:
Ahh, this is super useful. Of the link icon options, I think I like #2 the best.

gabestein:
That seems to be the consensus, so I've pushed that version up. Seems good-enough-for-now, but if anyone wants to take a stab, go for it.

isTravis:
Looking good! I went in and made a couple small tweaks to give it hover effects and a larger click target.

Screen Shot 2019-06-12 at 12 22 02 PM

PR #407, opened 2019-06-11
feature: GDPR infrastructure and UI

idreyn:
This PR includes some things that will help make PubPub fully GDPR compliant:

- A new Privacy settings page that enables users to toggle their election to use third-party data collection services (currently just Keen). This lives at `/privacy/settings`  —  additionally, the `/privacy/policy` page now hosts the privacy policy, and there's room for a GDPR compliance page there as well.

Screen Shot 2019-06-11 at 12 52 29 PM

- A banner that appears on top of the screen allowing users to opt into or out of cookies not essential to the user experience (again, just Keen for now). This sticks to the header and disappears, hopefully as permanently as possible, once an election is made.

Screen Shot 2019-06-11 at 12 53 09 PM

- Finally, a new `gdprConsent` field on `User` that takes `true | false | null` as values. The logic around this is a little delicate: when a user makes a GDPR election and then later signs up for an account, that election is passed into their `User` model. But when _they_ log out, we invalidate that election for future users that are created, and ask them to re-elect once their new account is created. So the election sticks to the device, but not to every user that creates an account on the device.

_Test plan:_
- Visit `/privacy/settings` and verify that the third-party analytics toggle successfully modifies the `gdpr-consent` cookie. If you're logged in, this should also update the `gdprConsent` field on your `User` model. When the toggle is disabled, it should immediately delete the `keen` cookie.
- Remove all cookies and verify that on any page with the `Header` you see the big GDPR banner. The "Accept" and "Decline" buttons should serve exactly the same purpose as turning the aformentioned toggle on and off, respectively.
- Log out, clear cookies, and consent to cookies. Then make an account and verify that the acceptance is passed into the new user's `gdprConsent` field. Repeat the same thing by declining the cookies. Now log out and ensure that the banner is still not shown. Now create a new user account and ensure that the `gdprConsent ` value has _not_ been forwarded to this new user's account, and they are asked to make an election again.
- [To check on V6 dev] Make sure that the Keen SDK starts up iff the user has given GDPR consent.


client/utils/isProd.js
```diff
@@ -0,0 +1,9 @@
+let isPubPubProduction;
+
+export const setIsProd = (val) => {
```
isTravis: This doesn't seem to be called anywhere. Previously, we set isPubPubProduction within hydrateWrapper right after we parsed the initialData.

idreyn: Oh, great catch.

PR #432, opened 2019-06-24
feature: in-Pub collection browsing

idreyn:
This PR adds some UI that adds collection-level context to the Pub reading experience. We need this for Frankenbook, but I'm hoping that it'll be useful more generally.

In the header, we have a new dropdown that shows all the other pubs available in a pub's primary collection, along with arrow buttons leading to the previous and next pubs in the collection. If there's no primary collection, this UI won't be shown. I've also added a little overflow menu to the collection tags that prevent them from needing to scroll horizontally across the header.

image

At the end of the pub, we'll also show a large `PubPreview` for the next pub in the primary collection, if it exists:

image

The links in these new elements add contain a `readCollection` URL parameter that takes precedence over the primary collection of a pub. That means that if we're on pub A and it has collection X as its primary, navigating to pub B will keep us in collection X even if pub B's primary is collection Y (hopefully that makes sense!)

_Test plan:_
- Visit a pub with a primary collection and make sure the primary collection browser works as expected. Try clicking through pubs in the dropdown menu and using the arrows to navigate from beginning to end.

- Check the bottom of these pubs (right below the footer) for the Read Next UI and verify that its links work.

- Verify that a pub with no primary collection loads as expected, but doesn't have either the collection browser or Read Next elements.


package.json
```diff
@@ -162,6 +162,7 @@
 		"terser-webpack-plugin": "^1.2.3",
 		"throng": "^4.0.0",
 		"tmp-promise": "^2.0.1",
+		"url-polyfill": "^1.1.5",
```
isTravis: So far we've been using polyfill.io and using a script tag within `server/Html.js`. This comes with the benefit that we don't load the polyfill into the bundle, and only fetch what's needed (polyfill.io returns a custom js based on the browser requesting it). Looks like polyfill.io has support for a URL polyfill.

Any reason you prefer using an npm import?

server/branch/permissions.js
```diff
@@ -1,6 +1,33 @@
 import { Branch, BranchPermission, PubManager, CommunityAdmin } from '../models';
 import { checkIfSuperAdmin } from '../utils';
-import calculateBranchAccess from './calculateBranchAccess';
```
isTravis: Just for clarity, is this and the related diffs below simply refactoring `calculateBranchAccess` into `/permissions` and renaming it `getBranchAccess`? I'm supportive of that, just want to make sure I'm not missing any logic updates.

server/collectionPub/api.js
```diff
@@ -5,17 +5,45 @@ import {
...
 	return {
 		userId: user.id,
-		collectionId: req.body.collectionId,
-		communityId: req.body.communityId,
+		collectionId: argsFrom.collectionId,
+		communityId: argsFrom.communityId,
 	};
 };
 
+app.get('/api/collectionPubs', async (req, res) => {
```
isTravis: This seems to be identical and duplicative of the `app.post()` directly below.

server/server.js
```diff
@@ -106,6 +106,7 @@ app.use((req, res, next) => {
 		req.headers.host = req.headers.communityhostname;
 	}
 	if (req.hostname.indexOf('localhost') > -1) {
+		req.headers.localhost = req.headers.host;
```
isTravis: Could you explain what this is for? Looks like something for storybook, but I can't totally follow.

server/branch/permissions.js
idreyn: Yep, that's the only thing!

server/collectionPub/api.js
idreyn: Huh, not sure how that happened!

server/server.js
idreyn: This is towards making links on the devserver keep you within your devserver rather than leading outbound to `x.pubpub.org`. I could break this out into another PR, since it is a kind of surprising thing to include here.

package.json
idreyn: Nope! I'll use the polyfill.io one instead.

Issue #396, opened 2019-06-07
Ability for community admins to request community deletion from the settings page


gabestein:
See #393, but for communities from the community settings page.

Note this one we may never want to automate. There are fairly strong arguments we can make that academic content falls under the GDPR [freedom of information exception](https://gdpr-info.eu/art-85-gdpr/). But of course there are cases where someone just plays around and wants to delete, so maybe there's a way we can say: if the community has been around for X time and has X content that's public, we disable the automatic option? Idk, out of scope for now but just to get it out there.

gabestein:
Hey @idreyn and @isTravis. I'm just not gonna get to this this cycle, and I do think we should have it before V6 if possible. It's super simple, so would one of you mind taking it? I think we can use basically identical language from the privacy settings page.

Issue #395, opened 2019-06-07
Ability for community admins to request data export from the settings page.


gabestein:
See #393, but for communities, on the community settings page.

PR #440, opened 2019-07-03
New/update: Community and User Export & Delete

gabestein:
- Adds export and delete section to community settings page
- Tweaks user export and delete mailto formatting and language

Addresses #395 and #396.

Screen Shot 2019-07-03 at 12 02 09 PM


client/containers/Dashboard/DashboardContent/Settings/Settings.js
```diff
@@ -23,6 +23,25 @@ const propTypes = {
 class Settings extends Component {
 	constructor(props) {
 		super(props);
+
+		/* Export & Delete Mailto Props */
+		this.exportEmailBody = `
```
gabestein: I did some research and I think adding props to the constructor is the best way to do this -- I can't just make it a constant in the file because I need to access props. And of course it shouldn't be state. But if there's a better way, lmk and I'll refactor.

Issue #459, opened 2019-07-22
Lost steps should be stored locally so they can be re-merged once user is back online


gabestein:

isTravis:
The changes introduced in #472 make it impossible to make further steps once an error has been fired. So, I do not believe new steps could be lost due to an error. In combination with the 'Unsaved changes' dialog popup, I believe we have most bases covered that lead to this situation.

The remaining case would be a situation where someone without internet connectivity quit their browser, ignoring messages - or had an unexpected computer shutdown. In that situation, the cleanest path to me feels like creating a new 'restored' branch that the user can decide how to merge (rather than forcing steps to merge into the current branch). 

Given those points, I'd like to close this for now, and re-open if we find this is still a significant issue when we have branch merge/diffing available.

PR #472, opened 2019-07-30
update: Add warning message when the Editor throws errors

isTravis:
This PR responds to #456. The bulk of functionality updates are in [pubpub-editor](https://github.com/pubpub/pubpub-editor/commit/68468e91876bb5b408bc172305c97f5edc6fc67a). Both local Prosemirror errors and asynchronous Firebase errors are caught, and passed to a new `onError()` param on the `<Editor />` component. 

This onError function is used in `<PubBody />` to trigger an alert that blocks further typing and forces a page reload.
Screen Shot 2019-07-30 at 12 02 03 PM

When `onError()` is triggered, we also fire a Sentry event and add a `{ error_source: 'editor'}` tag which will allow filtering from the Sentry end.
Screen Shot 2019-07-30 at 12 11 44 PM

Feedback/commits on the language of the error is especially welcome.

### DIFF
--- a/client/containers/Pub/PubDocument/PubBody.js
+++ b/client/containers/Pub/PubDocument/PubBody.js
@@ -4,8 +4,10 @@ import { useBeforeUnload } from 'react-use';
 import PropTypes from 'prop-types';
 import * as Sentry from '@sentry/browser';
 import { Card, Alert } from '@blueprintjs/core';
-import Editor from '@pubpub/editor';
+import Editor, { getJSON } from '@pubpub/editor';
 import { getResizedUrl } from 'utils';
+import TimeAgo from 'react-timeago';
+import { saveAs } from 'file-saver';
 import { PageContext } from 'components/PageWrapper/PageWrapper';
 import { PubSuspendWhileTypingContext } from '../PubSuspendWhileTyping';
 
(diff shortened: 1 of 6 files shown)


gabestein:
Looks great. Only thing I'm wondering is if there's a sane way to message that any edits made before the error will be saved, so the user doesn't fear refresh. Reading your comment on #459, looks like that's generally true -- but presumably some small amount of edits could be lost, right?

So could language like: "Don't worry, changes made before the error occurred have been saved." work and also be accurate?

Alternately, could we link out to the last step that was saved (ie, the link button on the history slider) with something like: "Don't worry, changes up to [this point] have been saved."

client/containers/Pub/PubDocument/PubBody.js
```diff
@@ -115,6 +117,15 @@ const PubBody = (props) => {
 						updateLocalData('collab', { editorChangeObject: editorChangeObject });
 					}
 				}}
+				onError={(err) => {
+					setEditorError(err);
+					if (window.sentryIsActive) {
```
idreyn: I hope this never happens in practice, but if this callback were triggered during a server-side render, I believe it would crash the request with a 500 because `window` is undefined. We probably want `typeof window !== 'undefined' && window.sentryIsActive` here.

isTravis: I thought about that and figured any Editor errors would've been caught when calling `getBranchDoc` to build the doc from Firebase steps in the first place. But, better to add the extra safety check. Will add a commit in a moment.

isTravis:
Good points, Gabe. The edge case where an error occurs during offline editing is the trickiest here. I updated the language and functionality to 1) specify how long it has been since the changes were last saved, and 2) to offer a download option if that time is greater than 0 seconds.

The download option will give them a file that contains the prosemirror docJSON. Folks will have to reach out to us to import that at the moment, but we could eventually add an 'import from backup' option. Since this backup is hopefully a rare last-resort solution, let's cross that bridge if/when we find a handful of folks having issues with it.

## Time Since last save = 0 seconds
Screen Shot 2019-07-31 at 5 24 56 PM

## Time Since last save > 0 seconds
Screen Shot 2019-07-31 at 5 25 36 PM

PR #478, opened 2019-08-01
dev: Some new testing infrastructure

idreyn:
This PR contains some code that we'll use to make it easier to write unit tests against our server-side code. There are no tests in this PR, but I've already written quite a few and gotten a sense that this setup is pretty ergonomic for doing so. Here is a summary of the changes:

1. We introduce the following dependencies:

- `supertest`: makes it easy to call Express routes in a test environment
- `supertest-as-promised`: wraps the above in a promise-based library
- `sinon`: contains a bunch of helpers for stubbing and spying code in tests

2. There are provisions for generating a new test database against the schema defined by Sequelize, and running tests against that database. This is still a bit shaky as it requires you to manually start a Postgres server — I'm working on automating the setup and teardown of that so it works on Heroku CI.

3. I am doing some surgery on `server.js`, which defines most of our Express logic. In summary, writing unit tests against our API means that we'll want to stop responding with HTTP `500` as a catch-all for both permissions failures and actual execution errors; otherwise, we won't be able to tell whether the server really rejected a request or it just broke. Ultimately, it seems like the idiomatic thing to do is throw errors relating to non-`2XX` status codes and let the error middleware handle those by responding with the appropriate status code. What _that_ implies is we need a way to catch errors propagated from our mostly promise-based route handlers. Hence, we'll need to wrap any asynchronous routes in `wrap()` until we migrate to Express 5.X, which has built-in support for such routes. Whew. 

4. Finally there's a new package of utilities called `stubstub` that I wrote to make writing tests easier. It has the following notable exports:

- `setup` and `teardown` are functions that we'll want to wrap around the [Jest-provided `(before|after)(Each|All)` functions](https://jestjs.io/docs/en/setup-teardown.html) that are called to setup and teardown tests. We want a place to put shared logic for getting the database in a good state for every test — there's not much in there now, but we reserve the right to put all kinds of things in them. So you should call `setup` and `teardown` at least once in every test suite that interacts with the database, like so:

```
// This will give you a clean slate before all the tests run
// If you want a clean slate before _each_ test runs, use beforeEach/afterEach instead
setup(beforeAll, async () => { ... })
// ...some tests, I hope
teardown(afterAll, async () => { ... })
```

If you `setup` and `teardown` with the `each` functions, that doesn't preclude you from also writing your own custom login in the `all` functions (or vice-versa). Use your best judgement on when this should occur — you can speed tests up by using `beforeAll`/`afterAll` and defining your models just once per suite, but you may have to do more manual housekeeping in each test to restore database state.

- `stub` is a simpler wrapper around `sinon.stub` that lets you stub out many functions from a module with a single function call.

- `makeUser` and `makeCommunity` create a user and a community, respectively. `makeCommunity` actually returns a `{community, admin}` pair, since almost every test involving a community will also need a reference to a community admin.

- `login` provides a [Supertest agent](https://github.com/visionmedia/supertest) for a given PubPub user. Typically you'll use it like this:
```
// You can also (await (await login(user)).post(...) if you're feeling evil
const agent = await login(user);
await agent.post('/some/route').send({foo: bar}).expect(200)
```
You can call it without any arguments to get an agent representing a logged-out user.

Putting it all together, a typical test file might look like this:

```js
// We are in server/whatever/__tests__/api.test.js here
// Jest just puts this stuff in the global scope, but we have to tell ESLint not to panic about it.
/* global describe, it, expect, beforeAll, afterAll, beforeEach afterEach */

// Imports from stubstub, and possibly some other query helpers.
import {
    makeUser,
    makeCommunity,
    setup,
    teardown,
    login,
    stub,
} from '../../../stubstub';
import { createPub } from '../../pub/queries';

// This module contains functions that we don't actually want the tests to call, often those that
// make requests to Firebase or other third-party services -- so we'll stub them out.
import * as moduleToStub from '../../wherever';

// Define some variables we'll reuse throughout our test suite. We have to use `let` here because
// they won't actually be assigned until the `setup` block. ¯\_(ツ)_/¯
let moduleStub;
let communityAndAdmin;
let pubManager;
let pub;

// Assign our models. If you don't properly async/await here, they won't be ready before the tests
// start running, and you're gonna have a bad time.
setup(beforeAll, async () => {
    communityAndAdmin = await makeCommunity();
    pubManager = await makeUser();
    pub = await createPub(
        { communityId: communityAndAdmin.community.id },
        pubManager
    );
});

// We're doing this in a beforeEach block in case we want to have a fresh mock for every test case.
// Otherwise we may get some false positives if stub state (information about whether a function was
// called or not, etc) persists across cases.
beforeEach(() => {
    moduleStub = stub(moduleToStub, ['methodToStub', 'otherMethodToStub']);
});

describe('/api/whatever', () => {
    it('is just a test', async () => {
        const { admin, community } = communityAndAdmin;
        // Login as a Supertest agent -- don't forget to await here!
        const agent = await login(admin);
        // Likewise, don't forget to await here too. Noticing a pattern?
        const { body } = await agent
            .post('/my/route')
            .send({
                pubId: pub.id,
                communityId: community.id,
                // ...some stuff relevant to your test
            })
            .expect(200);
        expect(body.someField).toEqual('something');
        expect(moduleStub.stubs.methodToStub.called).toEqual(true);
    });
});

// Restore the stubbed module in case we want the real thing in other tests.
afterEach(() => {
    moduleStub.restore();
});

// Call teardown.
teardown(afterAll);
```

_Test plan:_
Ironically, we don't really have one because there are no new tests in here! You can try running `stubstub/test-db.sh --create` to get a database to connect to, if you want.


server/utils/citations.js
```diff
@@ -33,6 +32,8 @@ const getCollectionLevelData = (primaryCollectionPub) => {
 };
 
 export const generateCitationHTML = (pubData, communityData) => {
+	// eslint-disable-next-line global-require
```
idreyn: This is my attempt to prevent `citation-js` from loading during tests, because it creates a subprocess that doesn't terminate (why???) and Jest complains that it can't exit.

isTravis:
It'd be great to break out the description you've written for this PR into README files in appropriate locations (e.g. the `/stubstub` folder).

PR #482, opened 2019-08-01
dev/update: Tests and updates to discussions-related code

idreyn:
This PR adds tests for discussions-related code. It also makes some fairly significant changes to the way discussion permissions are handled, so they now respect branch permissions. This is adequately covered by the new tests, but I'll do some manual testing before merging this as well.

_Test plan:_
- `npm run test server/discussion`
- Also manually poke around a pub and make sure you can still add discussions as a user with `discuss` branch permissions

### DIFF
--- a/server/discussion/permissions.js
+++ b/server/discussion/permissions.js
@@ -12,9 +12,7 @@ export const getPermissions = async ({
 	discussHash,
 }) => {
 	if (!userId) {
-		return new Promise((resolve) => {
-			resolve({});
-		});
+		return {};
 	}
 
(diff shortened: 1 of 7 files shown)


server/discussion/permissions.js
```diff
@@ -1,55 +1,61 @@
...
+export const getPermissions = async ({
+	branchId,
+	discussionId,
+	pubId,
+	userId,
+	communityId,
+	discussHash,
+}) => {
 	if (!userId) {
 		return new Promise((resolve) => {
```
isTravis: Because every async function returns a promise - can't we simply say `return {};` here?

idreyn: Yeah, you're right, that's much better.

PR #498, opened 2019-08-08
fix: Update keen to run without personal data before consent

isTravis:


gabestein:
Thanks y'all! 

Just to note for folks browsing and wondering about our tracking policies, I think we're on solid ground here in terms of balancing our desires to minimizing tracking as much as we can, eliminate any tracking that has a potential for abuse, while at the same time getting the data we need to make the product (and the academy!) better. 

Thus, without consent, we'll record pageviews, time on page, and some click events, but it will be completely anonymous -- no IP/geo -- the cookie will be first-party, and we won't even create a faux user id, making it impossible for us to track individual users across sessions.

Ultimately, we'd like to move this server-side to be even more compliant and make it even harder for us to track things we shouldn't, unintentionally or not.

https://brianclifton.com/blog/2011/05/20/google-analytics-and-the-new-eu-privacy-law/
https://brianclifton.com/blog/2011/06/16/google-analytics-and-the-new-eu-privacy-law-2/
https://www.hobo-web.co.uk/eu-privacy-directive-google-analytics-minimally-intrusive-unlikely-to-get-you-into-trouble/
https://gds.blog.gov.uk/2012/03/19/its-not-about-cookies-its-about-privacy/

PR #508, opened 2019-08-19
update: Make /collection/:id requests redirect to collection page, where applicable

idreyn:
This PR updates the behavior of the `/collection/:id` route (notable for being referenced in DOI deposits) so that it resolves to the Page associated with a Collection, where one exists.

_Test plan:_
- Create a collection and copy its URL (from the Metadata tab of the collection editor) into the search bar. Verify that it redirects to a `/search` URL.
- Now link your collection to a Page (from the Details tab of same). Verify that the same collection URL now redirects to the Page.
- `npm test server/routes`


coahearn:
:)

PR #529, opened 2019-09-04
fix: Update page title when Pub title changes

idreyn:
Does what it say on the tin — makes the page title change when we edit the Pub title in real time.

_Test plan:_
Open a pub, edit the title by clicking on it in the header, make sure the page title changes.

### DIFF
--- a/client/containers/Pub/PubSyncManager.js
+++ b/client/containers/Pub/PubSyncManager.js
@@ -1,15 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import queryString from 'query-string';
 
 import { loginDataProps } from 'types/base';
 import { apiFetch, getRandomColor } from 'utils';
-import queryString from 'query-string';
 import { initFirebase } from 'utils/firebaseClient';
+import { getPubPageTitle } from 'shared/utils/pubPageTitle';
 
 const propTypes = {
 	pubData: PropTypes.object.isRequired,
 	children: PropTypes.func.isRequired,
 	locationData: PropTypes.object.isRequired,
+	communityData: PropTypes.object.isRequired,
 	loginData: loginDataProps.isRequired,
 };
 
@@ -169,10 +171,11 @@ class PubSyncManager extends React.Component {
 	}
 
 	updatePubData(newPubData) {
-		/* First, set the local state. */
+		/* Firs t, set the local state. */
 		/* Then, sync appropriate data to firebase. */
 		/* Other clients will receive updates which */
 		/* triggers the syncMetadata function. */
+		document.title = getPubPageTitle(this.state.pubData, this.props.communityData);
 		this.setState(
 			(prevState) => {
 				const nextData =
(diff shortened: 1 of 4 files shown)


client/containers/Pub/PubSyncManager.js
```diff
@@ -169,10 +171,11 @@ class PubSyncManager extends React.Component {
 	}
 
 	updatePubData(newPubData) {
-		/* First, set the local state. */
+		/* Firs t, set the local state. */
```
isTravis: Typo?

client/containers/Pub/PubSyncManager.js
```diff
@@ -212,6 +215,7 @@ class PubSyncManager extends React.Component {
 				if (this.state.firebaseRootRef && hasUpdates) {
 					this.state.firebaseRootRef.child('metadata').update(firebaseSyncData);
 				}
+				document.title = getPubPageTitle(this.state.pubData, this.props.communityData);
```
isTravis: What is the need for having the `document.title = get...` line both on line 178 and line 218? Is there a case where those two are setting different values?

client/containers/Pub/PubSyncManager.js
```diff
@@ -1,15 +1,17 @@
...
 import { apiFetch, getRandomColor } from 'utils';
-import queryString from 'query-string';
 import { initFirebase } from 'utils/firebaseClient';
+import { getPubPageTitle } from 'shared/utils/pubPageTitle';
 
 const propTypes = {
 	pubData: PropTypes.object.isRequired,
 	children: PropTypes.func.isRequired,
 	locationData: PropTypes.object.isRequired,
+	communityData: PropTypes.object.isRequired,
```
isTravis: Should this be pulling from PageWrapper context rather than being passed as a prop?

idreyn: The Context API is not ergonomic enough to use in a callback of a class-based React component. We'd either have to use a `<Consumer>` component and bend over backwards to capture the context passed into its function-child so that it can be referenced outside of `render()`, or rewrite the component in Hooks.

We really do need a `usePageContext()` hook though. It's like four lines of code and the next time I come up against this problem I'm going to implement it.

client/containers/Pub/PubSyncManager.js
idreyn: Ah, this was a mistake. My bad!

client/containers/Pub/PubSyncManager.js
idreyn: I wish I had a cat to blame this one on

isTravis: 🐈

client/containers/Pub/PubSyncManager.js
isTravis: Ah! Right - from the git diff view, I didn't see/remember this was still using React.Component.

PR #573, opened 2019-10-10
Feature: Improved Pub import

idreyn:
This PR contains our new file-to-Pub import mechanism. Most of the guts of the converter are in the [prosemirror-pandoc](https://github.com/pubpub/prosemirror-pandoc) repository — what this PR contains is supporting code for the worker task, and the frontend UI. Some new features include:

- Multi-file import — bring in a `.tex` or JATS `.xml` file and its image dependencies at the same time.
- Integration of `pandoc-citeproc` to parse standalone bibliography files (e.g. `.bib`) into first-class PubPub citation objects
- More reliable footnote handling
- Support for importing LaTeX math blocks directly into a Pub instead of copy-pasting.

There is more work to be done here which we will defer to future cycles:
- Inline image conversion support for quasi-exotic formats like `.tif` that we see in the wild
- Support for JATS bibliography objects
- Import preview functionality 

_Sorta cruddy gif:_
Screen Recording 2019-10-10 at 1 51 27 PM 2019-10-10 13_55_58

_Test plan:_
Create a new Pub and then click the large Import button in the center of the Pub content to begin the import. Try .docx files, .tex files and their associated media, etc. There are all sorts of minutiae that we can check for here:
- The import button should be visible iff the Pub body is empty
- The "use as" tags on files in the uploader should correctly indicate to the worker which files are play what role in the import — document, bibliography, or supplement (secondary files that we pass to Pandoc, like `.tex` dependencies)
- The converter should provide useful warnings about missing files/citations and should fail gracefully on malformed documents.


client/containers/Pub/PubDocument/FileImport/useFileManager.js
```diff
@@ -0,0 +1,100 @@
+import { useState } from 'react';
+
+import { s3Upload } from '../../../../utils';
```
isTravis: I believe this can just be `from 'utils'` as it is aliased in our webpack config.

isTravis:
This is great - so excited to see this go live. Really nice experience for multi-file, and glad to see the expandable details part.

The only UX bug I think we need to fix before pushing this live is that [cancelling does not seem to clear the uploaded file and import status](http://g.recordit.co/1112Nmm4ou.gif). Unless, this is the intended behavior? I would expect Cancel to reset the whole process though.

I pushed up a commit with some language and style changes (to use Blueprint components rather than custom layout) - before and after below.

changes

PR #582, opened 2019-10-25
feature: New accessible menu component

idreyn:
While Blueprint serves us well as a UI framework, it's unfortunately not very good for keyboard and screenreader users, especially in places where complex interactions and portals are involved. In particular, dropdown menus are practically unusable with a keyboard. [This appears to be a known issue](https://github.com/palantir/blueprint/issues/3079) with a `help wanted` tag, but given the amount of work that goes into correctly implementing accessible UI elements _and_ merging that code into a complicated foreign codebase, I decided against trying to submit a PR to Blueprint to fix this.

Instead, I've used a very promising new UI framework called [Reakit](https://reakit.io/) to build a new menu component, which can be navigated using the keyboard and provides useful screenreader announcements. We use the Blueprint CSS classes in the component, so the look and feel matches the rest of PubPub. I'm very excited about Reakit, and I'd like to push us to use it to "a11y-ize" a few more questionable Blueprint components such as `Dialog`, and as a basis for our custom components such as `Discussion` as well. Using [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) correctly is incredibly difficult, and it makes sense to build off this developer's efforts and put them on the shortlist of projects to eventually consider funding.

Anyway, back to the menu. As a proof-of-concept, I've added it to the global nav user menu, and to dropdowns in the site nav, which makes these items newly keyboard/screenreader accessible. In the medium term I'll be replacing Blueprint menus as I come across them with this component, and fleshing out its props as necessary.

_Actually, Frankenstein was the doctor. The real monster is this mashup of two UI frameworks:_
Screen Recording 2019-10-25 at 1 45 04 PM 2019-10-25 14_01_23

_Test plan:_
Visit a community and make sure that the global nav user menu in the top right, as well as dropdowns in the site nav, are usable with mouse, keyboard, and keyboard + screenreader.

### DIFF
--- a/client/components/Header/Header.js
+++ b/client/components/Header/Header.js
@@ -1,17 +1,9 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import {
-	Popover,
-	PopoverInteractionKind,
-	Position,
-	Menu,
-	MenuItem,
-	MenuDivider,
-	Button,
-	AnchorButton,
-	Intent,
-} from '@blueprintjs/core';
+import { Button, AnchorButton, Intent } from '@blueprintjs/core';
+
 import { GridWrapper } from 'components';
+import { Menu, MenuItem } from 'components/Menu';
 import Avatar from 'components/Avatar/Avatar';
 import DropdownButton from 'components/DropdownButton/DropdownButton';
 import { apiFetch, getResizedUrl } from 'utils';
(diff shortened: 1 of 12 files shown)


client/components/Header/Header.js
```diff
@@ -217,51 +207,6 @@ class Header extends Component {
 							)}
 						</div>
 						<div className="buttons-wrapper">
-							{headerLinks.map((linkItem, index) => {
```
idreyn: Rather than adapting it to use the new `Menu`, I removed this code referencing `headerLinks` because this feature seems to be obsolete — there is no way to add them from the UI and no communities in prod currently make use of them.

isTravis:
> Actually, Frankenstein was the doctor. The real monster is this mashup of two UI frameworks:

😆

PR #634, opened 2019-11-22
update: pricing page

gabestein:
Updates the pricing page to reflect our latest model ideas. Also updates the email address to partnerships@ (partners@ also works), which goes to select team members as well as automatically to freshdesk to record a lead.

_Test plan_
- Read it and see what you think?
- Make sure the email links work and that sending emails to partnerships@ work

Screen Shot 2019-11-21 at 7 28 09 PM

### DIFF
--- a/client/containers/Pricing/Pricing.js
+++ b/client/containers/Pricing/Pricing.js
@@ -28,11 +28,9 @@ const Pricing = (props) => {
 					<h1>Pricing</h1>
 					<p className="description">
 						PubPub's goal is to provide a high-quality, open-source, institution-owned,
-						hosted alternative to commercial publishing software. We're committed to
-						giving anyone the ability to start a community and publish without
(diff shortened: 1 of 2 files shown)


client/containers/Pricing/Pricing.js
```diff
@@ -155,15 +125,26 @@ const Pricing = (props) => {
...
-						<a href="mailto:hello@pubpub.org?subject=Additional%20Services">
-							feel free to get in touch
-						</a>
-						.
-					</p>
+					<div className="pricing-footer">
+						<p className="description">
+							Need advanced features, but unsure you can afford PubPub Organization?
+							We offer substantial discounts for small, mission-driven communities.
+							information,{' '}
```
isTravis: Looks great overall. Seems like a word is missing here though.

PR #645, opened 2019-12-11
update: pub frontend metadata

gabestein:
This PR adds a bunch of metadata to Pub pages to be better conforming for Zotero, Google Scholar, and other services using the [Google Scholar / Highwire](https://scholar.google.com/intl/en/scholar/inclusion.html#indexing) schema. Included:
- A text representation of an abstract, if present
- PDF downloads
- Specific metadata for books, conferences, and issues

_Test plan_
- Create a Zotero account and Install Zotero and the Zotero chrome plugin (https://www.zotero.org/)
- Visit a pub in a book (https://hdsr.duqduq.org/pub/d1z0dryx), conference (https://hdsr.duqduq.org/pub/3mgfrtt1), and journal issue (https://hdsr.duqduq.org/pub/hx3pgxjc) and ensure that Zotero properly identifies their type and pulls in correct metadata fields for the types (book: ISBN, book title, chapter title; conference: conference title; journal issue: ISSN, volume, issue, journal title).
- Visit a pub that has a generated pdf download and ensure Zotero picks that up as the full text PDF (https://hdsr.duqduq.org/pub/hx3pgxjc)
- Visit a pub that has a formatted PDF download and ensure Zotero picks that up as the full text PDF (https://hdsr.duqduq.org/pub/iu26pfw1)
- Visit a pub with a complex abstract (https://hdsr.duqduq.org/pub/hx3pgxjc) with multiple inline things and ensure that it is: adding link URLs to the text where appropriate, and outputting latex math where appropriate. 
- Ensure that pubs with and without abstracts merge properly and can go through review properly.


server/utils/index.js
```diff
@@ -356,20 +404,16 @@ export const generateMetaComponents = ({
...
+			<meta key="pub1" name="citation_publisher" content="PubPub" />,
 			<meta key="pub2" property="dc.publisher" content="PubPub" />,
 		];
 	}
 
 	if (doi) {
 		outputComponents = [
 			...outputComponents,
-			<meta key="doi1" property="citation_doi" content={`doi:${doi}`} />,
+			<meta key="doi1" name="citation_doi" content={`doi:${doi}`} />,
```
idreyn: Out of curiosity, any particular reason this one is `name` while the others are `property`?

shared/pub/metadata.js
```diff
@@ -0,0 +1,54 @@
+export const getPDFDownloads = (pub) => {
+	const downloads = pub.downloads;
+	const exports = pub.activeBranch.exports;
+	if (downloads && downloads.length > 0) {
+		return downloads.filter((download) => {
+			if (download.url.substring(download.url.length - 4) === '.pdf') {
+				return download;
```
idreyn: For `filter` this should just return `true` if we want to keep the item.

shared/pub/metadata.js
```diff
@@ -0,0 +1,54 @@
+export const getPDFDownloads = (pub) => {
+	const downloads = pub.downloads;
+	const exports = pub.activeBranch.exports;
+	if (downloads && downloads.length > 0) {
+		return downloads.filter((download) => {
+			if (download.url.substring(download.url.length - 4) === '.pdf') {
```
idreyn: Or `download.url.endsWith('.pdf')`

shared/pub/metadata.js
```diff
@@ -0,0 +1,54 @@
...
+				return download;
+			}
+			return false;
+		});
+	}
+	if (exports && exports.length > 0) {
```
idreyn: ⚠️ I think this logic needs to be reworked so that if `downloads.length > 0` but we don't find a `.pdf` download, we fall through to the exports array here. Something like:

```js
if (downloads) {
    const matchingDownload = downloads.find(dl => dl.url.endsWith('.pdf');
    if (matchingDownload) return matchingDownload;
}
if (exports) {
   // ...
}
```

shared/pub/metadata.js
```diff
@@ -0,0 +1,54 @@
...
+	}
+	return false;
+};
+
+export const getTextAbstract = (content) => {
+	let abstract = '';
+	if (!content) {
+		return abstract;
+	}
+	if (
```
idreyn: ⚠️ This is missing an existence check for `content.content[...]`. The document may have only one (or maybe even zero?) entries in the `content` array.

shared/pub/metadata.js
```diff
@@ -0,0 +1,54 @@
...
+			if (exportFile.format === 'pdf') {
+				return exportFile;
+			}
+			return false;
+		});
+	}
+	return false;
+};
+
+export const getTextAbstract = (content) => {
```
idreyn: Let me suggest making this a bit more readable by destructing the document a bit and adding clarifying names:
```js
// Names like 'doc' or 'docJson' are conventional in our codebase to suggest
// that the variable is a Prosemirror doc
export const getTextAbstract = docJson => {
	const {content} = docJson;
	const [firstChild, secondChild] = content;
	// See whether the first child is the header we expect
	const firstChildIsAbstractHeader =
		firstChild &&
		firstChild.type === 'heading' &&
		firstChild.level === 1 &&
		firstChild.attrs.id === 'abstract';
	if (firstChildIsAbstractHeader && secondChild) {
		// Start extracting content from the second paragraph
		const {content: abstractContent} = secondChild;
		if (abstractContent) {
			abstractContent.forEach(/* ... */);
		}
	}
}
```

shared/pub/metadata.js
```diff
@@ -0,0 +1,54 @@
...
+	let abstract = '';
+	if (!content) {
+		return abstract;
+	}
+	if (
+		content.content[0].type === 'heading' &&
+		content.content[0].attrs.level === 1 &&
+		content.content[0].attrs.id === 'abstract'
+	) {
+		content.content[1].content.forEach((item) => {
```
idreyn: ⚠️ Likewise here. In general, there's no guarantee that a Prosemirror node will have a `content` member. (It took me months to remember all these rules and I still sometimes call it `children` by accident)

shared/pub/metadata.js
```diff
@@ -0,0 +1,54 @@
...
+
+export const getTextAbstract = (content) => {
+	let abstract = '';
+	if (!content) {
+		return abstract;
+	}
+	if (
+		content.content[0].type === 'heading' &&
+		content.content[0].attrs.level === 1 &&
+		content.content[0].attrs.id === 'abstract'
```
idreyn: This is probably fine, but I think it might be more robust to check the h1 content than its `attrs.id`, which we might repurpose for intra-document linking in the future.

server/utils/index.js
gabestein: I found out that, technically, all of the Highwire/Google Scholar (`citation_`) tags should be `name`, while all of the Dublin Core (`dc:`) tags should be `property`. Zotero was failing to grab some of the tags because of that (I [suggested](https://forums.zotero.org/discussion/80424/zotero-not-pulling-in-a-few-highwire-fields-for-pubpub#latest) they parse both fields like some other platforms do).

So, all the `citation_` tags should now be name, all of the `dc:` tags should now be `property`.

shared/pub/metadata.js
gabestein: what a useful function!!!

shared/pub/metadata.js
gabestein: Thank you. This is the type of code I would like to learn to write. I'm real rusty on destructuring and never think of stuff like creating named bools out of various properties.

PR #648, opened 2019-12-12
Paged branding

gabestein:
All this does is makes the title color of the PDF the community dark color. We'll do more branding later, but this is a small thing that will make folks happy for now.

### DIFF
--- a/workers/tasks/export/metadata.js
+++ b/workers/tasks/export/metadata.js
@@ -50,7 +50,7 @@ export const getPubMetadata = async (pubId) => {
 		publishedDateString: publishedDateString,
 		updatedDateString: updatedDateString,
 		communityTitle: pubData.community.title,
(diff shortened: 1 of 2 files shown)


workers/tasks/export/metadata.js
```diff
@@ -50,6 +50,7 @@ export const getPubMetadata = async (pubId) => {
 		publishedDateString: publishedDateString,
 		updatedDateString: updatedDateString,
 		communityTitle: pubData.community.title,
+		communityAccentColorDark: pubData.community.accentColorDark,
```
idreyn: My weak suggestion here is that you could call this field `accentColor` or even `color`, to decouple the rendering code from the idea of where specifically the color came from.

gabestein: 👍 
Going with `accentColor` because I can see other colors being added in the future.

PR #658, opened 2019-12-19
new: DOIs

gabestein:
Adding some DOIs.


gabestein:
Thanks. It probably will and we probably should next time we make any additions.

PR #659, opened 2019-12-19
new: Google Scholar/Highwire metadata

gabestein:
This generates Google/Highwire meta tags for references on the frontend for pubs that use built-in footnotes or citations. It's a pretty naive generator; on the other hand, the tags themselves are [not well specified](https://citation.js.org/blog/?post=3781582023912504297). Eventually, we should probably write a CSL to produce these, but that seems like a multi-week project not particularly worth our time at the moment.

Features of this particular PR:
- De-duplicates among footnotes and tags
- Supports journal, conference, and book (defaulting to journal)
- Filters out unstructured inputs

_Test plan_
1. Visit pub with lots of citations and verify they are represented in meta tags 
2. Visit pub with duplicate tags (both within footnote/citation and across footnote/citation) and verify they are _not_ replicated in meta tags. (ex: https://hdsr.duqduq.org/pub/hx3pgxjc)
3. Visit pub with journal, book, and conference citations and verify they are represented properly with `citation_journal_title`, `citation_book_title`, and `citation_conference_title` respectively.
4. Create new pub. Publish with no references. Then add some references and publish. Make sure meta tags are not there before, there after, and that merging works as expected.

### DIFF
--- a/shared/pub/metadata.js
+++ b/shared/pub/metadata.js
@@ -50,43 +50,41 @@ export const getTextAbstract = (docJson) => {
 	return abstract;
 };
 
+const noteTypes = {
+	chapter: 'book',
+	book: 'book',
+	proceedings: 'conference',
+	'paper-conference': 'conference',
+};
+
 export const getGoogleScholarNotes = (notes) => {
 	return notes.reduce((unique, note) => {
 		const noteArray = [];
 		const noteType = note.json[0].type;
-		const noteTypes = {
-			chapter: 'book',
-			book: 'book',
-			proceedings: 'conference',
-			'paper-conference': 'conference',
-		};
 		const noteTypeString = noteTypes[noteType] || 'journal';
 
-		const noteArr = Object.entries(note.json[0]);
-		noteArr.forEach((entry) => {
-			switch (entry[0]) {
+		Object.entries(note.json[0]).forEach(([key, value]) => {
+			switch (key) {
 				case 'title':
-					noteArray.push(`citation_title=${entry[1]}`);
+					noteArray.push(`citation_title=${value}`);
 					break;
 				case 'author':
-					entry[1].forEach((author) => {
+					value.forEach((author) => {
 						noteArray.push(`citation_author=${author.given} ${author.family}`);
 					});
 					break;
 				case 'container-title':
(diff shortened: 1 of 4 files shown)


shared/pub/metadata.js
```diff
@@ -49,3 +49,49 @@ export const getTextAbstract = (docJson) => {
 	}
 	return abstract;
 };
+
+export const getGSNotes = (notes) => {
```
idreyn: Could we call this `getGoogleScholarNotes` instead? `GS` is a little opaque for someone reading this code out of context.

shared/pub/metadata.js
```diff
@@ -49,3 +49,49 @@ export const getTextAbstract = (docJson) => {
...
+	return notes.reduce((unique, note) => {
+		const noteArray = [];
+		const noteType = note.json[0].type;
+		const noteTypes = {
```
idreyn: Nit: this can be factored out of the function.

shared/pub/metadata.js
```diff
@@ -49,3 +49,49 @@ export const getTextAbstract = (docJson) => {
...
+			chapter: 'book',
+			book: 'book',
+			proceedings: 'conference',
+			'paper-conference': 'conference',
+		};
+		const noteTypeString = noteTypes[noteType] || 'journal';
+
+		const noteArr = Object.entries(note.json[0]);
+		noteArr.forEach((entry) => {
```
idreyn: Personally I'd find this a bit easier to read as `Object.entries(...).forEach(([key, value]) => { ... })`. Something about `entry[0]` or `entry[1]` just adds cognitive overhead in remembering what it is I'm looking at. And having `noteArray` and `noteArr` in the same scope is a little confusing too.

gabestein: Done, thanks. Had no idea you could do that in forEach (is that...destructuring, what is that?). Anyhow, super helpful.

shared/pub/metadata.js
gabestein: +1

shared/pub/metadata.js
gabestein: I think this is what you meant?

PR #667, opened 2020-01-11
dev: Refactor PageWrapper to App

isTravis:
This PR refactors our code structure and build hierarchy to have a single React component as the root of all views. This is in contrast to splitting up each view as a separate build chunk. Rather than [`vendor.js` + `about.js`, `vendor.js` + `pub.js`, ...] we now have [`vendor.js` + `main.js`]. 

This simplifies the build, but more importantly, greatly simplifies the code structure. We introduce a single `App.js` folder that implements headers, footers, pageContext, etc. 

In the process of doing this, we needed to refactor many containers to use a React function component rather than a class component, so we could use the `usePageContext` hook. 

Many files had to be touched for this PR, but it's mostly boilerplate changes. 

To get an overview of meaningful changes, I would suggest reviewing the following files:
```
client/containers/App/App.js
client/containers/App/paths.js
client/webpack/webpackConfig.prod.js
client/utils/hydrateWrapper.js
server/Html.js
server/routes/explore.js (example of route-level changes)
client/containers/Explore/Explore.js (example of container-level changes)
```


client/containers/App/paths.js
```diff
@@ -0,0 +1,117 @@
...
+	Privacy,
+	Pub,
+	Search,
+	Signup,
+	Terms,
+	User,
+	UserCreate,
+} from 'containers';
+
+export default (viewData, locationData, chunkName) => {
```
idreyn: This feels good. My only suggestion is to note that this function doesn't depend on `viewData` in any meaningful way...it just applies it to the `activeComponent` and sends that back to the caller, and means that we have to write `<Component {...viewData} />` over and over here. Better perhaps to do e.g.:

```js
const paths = {
    About: {
         Component: About,
         // ...
    },
    // ...
}
```
and then in `App.js` just render `<Component {...viewData} />` the one time.

client/containers/App/App.js
```diff
@@ -0,0 +1,74 @@
...
+	const pageContextProps = {
+		communityData: communityData,
+		loginData: loginData,
+		locationData: locationData,
+		scopeData: scopeData,
+	};
+
+	const showNav = !hideNav && !communityData.hideNav && !isDashboard;
+	const showFooter = !hideFooter && !isDashboard;
+	return (
```
idreyn: Completely subjective, but I think most of the spaces between components in the JSX here are unnecessary.

client/containers/CommunityCreate/CommunityCreate.js
```diff
@@ -1,177 +1,144 @@
...
-			createIsLoading: false,
-			createError: undefined,
-		};
-		this.onCreateSubmit = this.onCreateSubmit.bind(this);
-		this.onSubdomainChange = this.onSubdomainChange.bind(this);
-		this.onTitleChange = this.onTitleChange.bind(this);
-		this.onDescriptionChange = this.onDescriptionChange.bind(this);
-		this.onHeroHeaderLogoChange = this.onHeroHeaderLogoChange.bind(this);
-	}
+const CommunityCreate = () => {
```
idreyn: A future refactor like this could make use of the `<PageContext.Consumer>` component in the render function and skip the hooks conversion altogether. I'm happy to see our codebase gradually drop React class components, but I think it pays off to be deliberate about it — there have been several very subtle errors that have crept in when we've tried to hookify things on our way through other changes, especially when (as is not the case here) `useEffect` is involved.

client/containers/CommunityCreate/CommunityCreate.js
```diff
@@ -1,177 +1,144 @@
...
+	const { loginData } = usePageContext();
+	const [subdomain, subdomainSetter] = useState('');
```
idreyn: I think we should stick with the `[val, setVal]` convention for `useState` bindings here.

client/containers/App/paths.js
isTravis: Definitely - good catch. A copy-paste holdover that didn't get enough thought. I'll make that change.

client/containers/App/App.js
isTravis: ++

client/containers/CommunityCreate/CommunityCreate.js
isTravis: Ya, you're right. I did this because it was easier to handle the parallel conversion of 10+ state fields in components like `<UserCreate />`. Having to change the capitalization of the `v` in `[val, setVal]` always took a bunch of manual editing, but duh - there's a shortcut to do that, so my original concern is irrelevant.
```
cmd + KU    Transform to Uppercase
cmd + KL    Transform to Lowercase
```

Issue #539, opened 2019-09-16
Implement Editor toolbar and formatting designs


isTravis:
Deepak's new designs for footnotes and citations are available on Figma. Implement these to be functional in code.

gabestein:
While we're implementing here, there does appear to be some instability with footnotes and citations, including:

- Frequent failures to parse DOIs and fetch results
- Errors when adding structured text
- That recurring issue where selecting a footnote/citation and then another of the same type doesn't update the toolbar (I think the plan was to fix that with this implementation)

gabestein:
Good suggestion from Hanna:

"When I write/edit a pub and insert a footnote link, I have to go back to the number inserted in the text and click on it in order to be able to insert a footnote. It would be helpful if, when inserting a footnote, I would automatically be in the editor and could directly type in the footnote text."

IE, clicking footnote in the editor bar would ideally focus on the created footnote node and bring you into the footnote text editor in the editor bar.

PR #673, opened 2020-01-14
feature: Formatting bar rewrite

idreyn:
This PR contains a complete rewrite of significant parts of the editor UI, clustered around the "formatting bar" that lets users apply formatting to text as well as add and modify the properties of elements in the document like images, citations, etc. Some notable UX improvements include:

- Better overall design (thanks, Deepak!)
- Improved keyboard usability
- Ability to preview changes to node attributes that might have undesirable intermediate states (like bibtex citations or latex equations) before committing them to the document
- Support for nesting toolbars, enabling better rich-text capabilities for document nodes. For instance, it is now possible to add a link or an equation to a footnote.
- Better contextual clues indicating which toolbar items correspond to which nodes
- Resolution for a class of bugs clustered around formatting bar and document state. For instance, it is currently common for the formatting bar to not update its displayed attributes when the user changes their selection between two different nodes of the same type, or to apply the attributes of the previously selected node to the newly selected one. This kind of thing should no longer occur.

Tucked inside of this PR is an update for editor performance that could be separated out if desired. For context: I develop on a 2013 Macbook Pro which still has pretty good specs, but shows its age on some web applications. PubPub has traditionally been one of them: long Pubs or Pubs with lots of comments or concurrent editors would make the editor unusably slow for me. The root of the issue is that even small changes within Prosemirror can generate dozens of `editorChangeObject` instances representing the current editor state. This is not a performance issue for the editor itself, but each of these updates triggers React lifecycle events and can lead to tasks that take >500ms.

To address this I've decoupled the Prosemirror `editorChangeObject` updates from React `setState` updates in `PubSyncManager` — we store the editor updates in the order we receive them, and batch-apply them to the React UI when _not_ processing editor input events using `requestIdleCallback()`. There's some fine-tuning left to be done here, but I've tested the editor for many cumulative hours and the behavior appears completely correct and not that startling from a latency perspective; occasionally there will be lag between the editor and the formatting bar.

Finally, there are some bits and pieces that I've left out of this PR in the interest of shipping something this week: 
- Remove the old formatting bar from other Prosemirror editors, like the Page editor and the discussion widgets. This is worth doing, but requires attention to detail and I'd rather focus on it separately.
- Replace the stock Blueprint formatting bar icons with a few custom ones
- Remove `PubSuspendWhileTyping` if our new performance solution appears sufficiently one-size-fits-all
- Better empty states, onboarding, guidance for how to use the editor
- Auto-selecting newly created nodes
- A table editor that we can all be proud of
- Various small look-and-feel things that aren't 100% there

**Test plan:**
This is too much stuff to write down a comprehensive test plan for, but let's make sure all of the formatting bar buttons _work_, either applying marks or creating nodes. All of the node editors (equation, footnote, citation, media, table) and the special-casey link editor should work too. Let's also test what happens when you click around in the document while node editors are open — we really don't want to transfer attributes from one image to another or anything like that.

**Screenshots:**
Again, a little too much to be comprehensive here, but here's a representative pair:

_old:_
Screen Shot 2020-01-14 at 4 12 53 PM

_new:_
Screen Shot 2020-01-14 at 4 13 30 PM


client/components/FormattingBar/positioning.js
```diff
@@ -0,0 +1,39 @@
+import { Selection, NodeSelection } from 'prosemirror-state';
```
isTravis: I've preferred keeping all prosemirror imports within the `@pubpub/editor` package. For use as we have below, I'd typically create a helper function in the editor package and `import { createNodeSelection } from '@pubpub/editor'`. I can understand that feeling verbose, but it does keep us from having different prosemirror package versions within the code base. 

What are your thoughts on that layer of abstraction?

client/components/FormattingBar/Controls/ControlsEquation.js
```diff
@@ -1,140 +1,104 @@
...
-
-const propTypes = {
-	attrs: PropTypes.object.isRequired,
-	updateAttrs: PropTypes.func.isRequired,
-	changeNode: PropTypes.func.isRequired,
-	selectedNode: PropTypes.object.isRequired,
-	editorChangeObject: PropTypes.object.isRequired,
-	isSmall: PropTypes.bool.isRequired,
-};
+import { useDebounce } from 'use-debounce';
```
isTravis: The `react-use` package that we already include has a [`useDebounce`](https://github.com/streamich/react-use/blob/master/docs/useDebounce.md) hook. If the interfaces are compatible, let's use that instead.

client/components/FormattingBar/FormattingBar.js
```diff
@@ -1,395 +1,287 @@
...
-	PopoverInteractionKind,
-} from '@blueprintjs/core';
-import { Overlay, DropdownButton } from 'components';
-import Controls from './Controls';
-import Media from './Media';
+import classNames from 'classnames';
+import { Toolbar, ToolbarItem, useToolbarState } from 'reakit';
+
+import { usePageContext } from 'containers/Pub/pubHooks';
+import { useRefMap } from '../../utils/useRefMap';
```
isTravis: Can simply use `from 'utils/useRefMap'`.

client/containers/Pub/PubDocument/PubBody.js
```diff
@@ -206,11 +213,11 @@ const PubBody = (props) => {
 								firebaseRef: firebaseBranchRef,
 								clientData: props.collabData.localCollabUser,
 								initialDocKey: pubData.initialDocKey,
-								onStatusChange: debounce((status) => {
+								onStatusChange: (status) => {
 									getNextStatus(status, (nextStatus) => {
 										props.updateLocalData('collab', nextStatus);
 									});
-								}, 250),
+								},
```
isTravis: The purpose of this debounce was to prevent the status from ever flickering as it quickly toggles between saved and saving states. With it removed, this seems to happen frequently:

flicker

If we do add a debounce back, best to use the `useDebounce` hook from `react-use` in a more appropriate place.

isTravis:
So many improvements! Thank you for diving so deep on this update. The update on close (for captions, equations, etc) works great now. Switching between items (e.g. two citations) feels clean and is finally bug free! Working through some testing and code comments now.

isTravis:
Media settings other than caption do not seem to persist on close: 

qYv28ImLAl

isTravis:
The legacy formatting bar on discussions seems to have added elements that are not intended for the discussion bar. Specifically, the block-type dropdown and ellipsis dropdown. The vertical line separators also appear to have been added in.

**Old**
Screen Shot 2020-01-15 at 11 23 41 AM

**New**
Screen Shot 2020-01-15 at 11 22 16 AM

client/components/FormattingBar/positioning.js
idreyn: Overall I'd like to find a way to write Prosemirror-oriented code inside of `pubpub` when it doesn't feel like the code is core to the functionality of `@pubpub/editor`. I hadn't considered the version matching thing, which is interesting — it seems like what I really want is a way to enforce that `pubpub` grabs the same version of all the Prosemirror packages as `@pubpub/editor`. For the time being, I'll move this to the editor package.

client/components/FormattingBar/Controls/ControlsEquation.js
idreyn: Unfortunately they're not; the `use-debounce` version does by default what I almost always want — provides a debounced version of an input value directly — while the `react-use` version just lets you pass in a callback from which you might update a state value manually. I think the API of the former is useful enough to justify having around, though arguably we could also implement it ourselves in a homegrown hook in just a few lines of code.

client/components/FormattingBar/FormattingBar.js
idreyn: Tell that to my editor!

client/containers/Pub/PubDocument/PubBody.js
idreyn: Huh, this is not happening for me, but best to restore it!

idreyn:
Thanks for doing some QA — I have fixed those two glitches!

idreyn:
I think the tests are failing because I renamed `/Controls` to `/controls` and `/Media` to `/media`, which is confusing some case-sensitive part of the system. Presumably we need to clear some sort of cache on Heroku to fix this, but since the tests are passing locally I'm going to merge and try again.

isTravis:
Closes #539

PR #684, opened 2020-01-27
generalize the GDPR banner to show other banners as well

### LINKED ISSUE #681: TOS Banner
Banner for all users that notifies them of updated terms once, drops a cookie and then never does it again. Should expire at a certain date (perhaps there's a way to do that via env vars or the db so we don't have to bump code to do it).

idreyn:
Here we press the GDPR banner into service for sequentially showing other legal notifications as well — in this case, for a terms of service update. When users accept the TOS change, we store a `tos-update` cookie with the current date and then use that to make sure they don't see it again (on this device).

I have made some styling changes to this banner so it has `position: fixed` and lives at the bottom of the screen. I think this is a little better than living at the top and reflowing the page content, but I have almost no opinion about this and am happy to defer to others.

Managing this banner will be done with a code change for the time being — overall easier than mucking with environment variables or whatnot.

Closes #681 

_Test plan:_
- Open a new browser session and make sure you see the GDPR banner, and that accepting or denying the third-party cookies causes the TOS banner to show.
- Click the "See terms" button and verify that it opens a link to the TOS in a new window.
- Close the banner. On the next page reload, you shouldn't see it, or the GDPR banner, again.

_Screenshot:_
image


gabestein:
Thank you! Looks great, except I don't think we want an `accept` button. They don't really have a choice if they want to use the website. So we should show it once with see terms and just have a `close` button or something like that.

idreyn:
image

We talked about this for a bit on the call today and messed around with styles and positioning. Taking everyone's input into account, this is what I've come up with.

The font is smaller and the buttons are bigger, the background is less obtrusively black and I've killed the PubPub logo. We now try to make it clear up front that there are multiple steps with the "breadcrumbs" at the top and the "Next" CTA on the TOS screen.

Finally, I've decided to give the popup a maximum width because it's easier on the eye and keeping the buttons at the bottom makes their position between screens more predictable.

Issue #678, opened 2020-01-23
Toolbar disappears in drafts


KFGBot:
_We use GitHub Issues for tracking active code development. If you have general questions or feedback about PubPub, we invite you to use our [Discourse forum](https://discourse.knowledgefutures.org)._

### Required
---

**What went wrong, step-by-step?**
1. At different points and not in every pub, the toolbar will disappear when scrolling in edit mode
2. Does not reappear when you stop scrolling
3. Can still select content, say an image, that appears after the "disappearing point," scroll back up until the toolbar reappears, and edit that content (such as add a caption). 
...

**What URL can we use to see the issue?**
Try this HDSR draft: https://hdsr.mitpress.mit.edu/pub/fxz7kr65/branch/2
The toolbar disappears for me around this paragraph: https://hdsr.mitpress.mit.edu/pub/fxz7kr65/branch/2?from=1278&to=1281

**What browser/operating system were you using?**
Chrome

### Optional

---

**Have any screenshots? (Attach them by dragging the image here)**

**Who reported this? (If not you)**

KFGBot:
Still running into this and not sure how to recreate it. It seems to always be at a different point in the doc. and not in all docs. Some links to other examples (linking to the approx point in the text where, for me, the toolbar disappears):
https://kfg.mit.edu/pub/teclkjnl/branch/2/1?from=1432&to=1445
https://kfg.mit.edu/pub/q2i554i5/branch/2/2840?from=18357&to=18381
https://hdsr.mitpress.mit.edu/pub/pfji5iif/branch/2/55?from=1271&to=1291

gabestein:
Think it is related to stickybits plugin. I'm gonna take a look.

PR #708, opened 2020-02-19
fix: Make sticky headers work when page size changes

idreyn:
This addresses #678 and also Deepak's report that scrolling to the bottom of a Pub will sometimes cause large amounts of unpleasant jittering. Hoping to get a few pairs of eyes on this to verify that the problem is fixed, since I had some problem reproducing it to begin with.

_Technical mumbo-jumbo:_
It seems that Stickybits doesn't automatically recompute certain values when the total height of the page content changes, such as when a lot of text is added to a Pub — instead, you need to explicitly call the `update` method on the instance it provides. Here we observe changes using the `ResizeObserver` API where available, and using a timeout otherwise (in Safari and IE/legacy Edge). There are polyfills available for this API, but they appear not to work in this case. I factored the correct behavior into a `useSticky` hook that we should use from here on out.

_Test plan:_
Open a Pub and verify that the header works in read-only mode. Switch to an editable branch and add enough text that the total height of the page content is significantly increased, and verify that both the pub header and formatting bar stick as expected, and that there is no jitter when reaching the bottom of the page.

### DIFF
--- a/client/containers/Pub/PubDocument/PubHeaderFormatting.js
+++ b/client/containers/Pub/PubDocument/PubHeaderFormatting.js
@@ -1,7 +1,7 @@
-import React, { useEffect, useRef } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import stickybits from 'stickybits';
 
+import { useSticky } from 'utils/useSticky';
 import { FormattingBar, buttons } from 'components/FormattingBar';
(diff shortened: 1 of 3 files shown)


idreyn:
Looking at the timings, the 100ms timeout tasks take about 35µs on my 2013mbp. If we had a few dozen sticky elements on a page I'd advocate for doing something smarter, but the current performance impact is undetectable.

PR #716, opened 2020-03-02
fix: Add a few additional redirects neeed by BAAS

pkgw:
The Bulletin of the American Astronomical Society (BAAS) just replaced their old WordPress site with PubPub, and there are a few URLs out in the wild that we want to preserve by establishing redirects. A couple of the new entries that we need are because PubPub doesn't accept URLs with trailing slashes for its established names. One of them is for a URL slug that apparently got changed at some point.


pkgw:
Ping?

gabestein:
Hey Peter – apologies, but we're not yet totally in the habit of accepting third-party contributions. I'll push the `/ten-years` redirect today (#723), but I want to wait for @isTravis on the trailing slashes for `/obituaries/` and `/obituaries-listing/`, because he just updated this last week, and I'm not sure if this is the right way to do it, or if we need to make some updates to how we handle wildcards. He's been out this week, but will be back Monday. We'll have more info then.

PR #719, opened 2020-03-06
new: Adds DMCA info without editing text of TOS

gabestein:
Pretty straightforward. Adds a dialog with DMCA information to the TOS that was missing in our initial push. Since this does not edit the TOS, I don't believe sending an update to users is required.

Screen Shot 2020-03-06 at 12 06 37

Screen Shot 2020-03-06 at 12 06 42

Only weird-ish thing is I used a button, restyled to look like a link, because jsxa11y complained about using an anchor to launch the dialog.

_Test plan_
1. Visit Terms (/legal/terms)
2. Click "please click here for more information"
3. Verify that dialog opens when the line is clicked
4. Close dialog by clicking on the `x` in the header
5. Verify that dialog can be closed with escape and by clicking outside the dialog

### DIFF
--- a/client/containers/Legal/Terms.js
+++ b/client/containers/Legal/Terms.js
@@ -385,8 +385,7 @@ const Terms = function(props) {
 				>
 					click here for more information
 				</Button>
-				<Dialog isOpen={isOpen} onClose={toggleOverlay}>
-					<p className={Classes.DIALOG_HEADER}>DMCA Agent Information</p>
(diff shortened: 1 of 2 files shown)


gabestein:
Thanks. I don't disagree, but if we change the text materially I think we have to alert all our users. Another option is just to link to somewhere else that has the information, but that feels less good from a UX perspective.

PR #727, opened 2020-03-17
update: First pass at implementing Overview UI design

isTravis:
This PR implements Deepak's Overview design for Collection and Community scopes. Some design elements are still missing as we build out capabilities for them.


client/containers/DashboardOverview/ContentOverview.js
```diff
@@ -14,19 +18,35 @@ const propTypes = {
 
 const ContentOverview = (props) => {
 	const { overviewData } = props;
-	const {
-		scopeData: {
-			elements: { activeTargetType },
-		},
-	} = usePageContext();
+	const { scopeData } = usePageContext();
+	const useCollectionStateObject = useCollectionState(scopeData);
```
idreyn: We should avoid naming the return value of these hooks with values starting with `use`:

- At a glance, the variables look like they themselves are hooks
- I don't think it's important to know that these values come from a hook. In modern React, the sources of truth for most stateful values _do_ come from a hook, and we should feel free to swap hooks in and out without the variable names bound to these values having to change as well!

Instead I would suggest:
- Destructuring at the hook's callsite, e.g. `const {xyz, updateXyz} = useXyz()`, and passing the object state and the updater functions as separate props to child components that consume them. This feels like 'idiomatic React' in the sense that data flows in as one prop, and the functions that update the data are a second prop — it also signals intent about whether a child component ought to be able to update state or merely consume it. This can be pattern-matched from the APIs of the built-in hooks, i.e. we get `[state, updateState]` from `useState()` and `[state, dispatch]` from `useReducer()`, and we're expected to destructure the return value. Most libraries that expose hooks do something similar. 
- (Less preferred) giving the returned state values a name like `xyzState` and retrieving the values as `xyzState.xyz` and `xyzState.updateXyz`.

client/containers/DashboardOverview/OverviewBlocks.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon } from 'components';
+
+require('./overviewBlocks.scss');
+
+const propTypes = {
+	overviewData: PropTypes.object.isRequired,
+	useCollectionPubsObject: PropTypes.object.isRequired,
```
idreyn: Another place where these kinds of objects are sort of obscuring intent — this signature should read:

```
useCollectionPubsObject: PropTypes.shape({
    collectionPubs: PropTypes.arrayOf(PropTypes.shape({ ... })),
}).isRequired
```

`PropTypes.object` should only be used when the shape of the object is truly irrelevant to the component whose signature is being specified — e.g. it is being passed directly to a child.

client/containers/DashboardOverview/OverviewBlocks.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const OverviewBlocks = (props) => {
+	const { overviewData, useCollectionPubsObject } = props;
+	const { collectionPubs } = useCollectionPubsObject;
+
+	const countItems = (type) => {
+		const activeList = collectionPubs ? collectionPubs.map((cp) => cp.pub) : overviewData.pubs;
+		return activeList.reduce((prev, curr) => {
```
idreyn: One thing I find improves readability of `reduce` calls is naming the arguments to the reducer something other than `(prev, curr)` so we can tell what they represent — in this case I'd suggest `(count, pub)`.

client/containers/DashboardOverview/OverviewControls.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
```
idreyn: I am wondering about the benefit of unifying the controls for community- and collection-level scopes into a single component, when the different scopes have very different needs. Right now there are no community-level controls and that scope is best served by passing in a single "Create Pub" button — I've been imagining developing a `<ScopedCreatePubButton>` component for this purpose. Over time, though, we could add some — and this component would be full of conditionals to accomodate two mostly disjoint sets of controls. Really, the biggest thing "community overview controls" and "collection overview controls" have in common is that they happen to exist in the same place in the UI, and that's not something this component itself is aware of — so I'd advocate for making them two separate components.

Zooming out a little, I see these two overview screens being best served by two different components, rather than one component with a lot of conditionals, and achieving reuse through composition.

client/containers/DashboardOverview/collectionState.js
```diff
@@ -12,6 +12,9 @@ const findRankForSelection = (selections, index) =>
 	);
 
 const linkCollectionPubs = (overviewData, collection) => {
+	if (!collection) {
```
idreyn: I'm confused about how we could arrive in a state where the collection being passed to this component is not defined, but it seems like an upstream concern — if there's no collection in scope, we should not be calling the `useCollectionState()` hook.

client/containers/DashboardOverview/PubSelect.js
```diff
@@ -23,16 +25,22 @@ const defaultProps = {
...
-
+	const renderPubItem = (pub, { handleClick, modifiers: { active } }) => {
+		const authorString = generateAuthorString(pub);
+		return (
+			<MenuItem
+				className={classNames({ item: true, active: active })}
+				key={pub.id}
+				onClick={handleClick}
+				text={
+					<React.Fragment>
```
idreyn: Decoupling this from `ContentRow` seems like the right decision, thank you!

isTravis:
Awesome - thanks for the thoroughness!

- Blocks have been made smaller for the moment, but I agree they don't feel like a great use of space. The numbers are available in the SideMenu, so ideally we'd have something more interesting/active/informative in this space eventually.
- Row misalignment has been fixed (missing a `flex: 0 0 auto`). Unfortunately, `<table>` and css grid become unwieldy at best, non-functioning at worst, when trying to implement drag-and-drog (which require `<div>` wrappers around each row).
- I share some angst about having a single `ContentOverview` component for Collection and Community overviews. I'm not wedded to the current solution - so open to refactoring if it begins to get in our way.

I'll reply directly to the remaining inline items. Thanks again!

client/components/DashboardFrame/dashboardFrame.scss
```diff
@@ -0,0 +1,33 @@
+.dashboard-content-header {
+    display: flex;
+    align-items: flex-start;
+    margin-bottom: 20px;
+    .dashboard-header-left {
+        .title {
+            span {
```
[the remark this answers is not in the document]
isTravis: 👍 Done.

client/containers/DashboardOverview/ContentOverview.js
```diff
@@ -2,9 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { usePageContext } from 'utils/hooks';
+import { DashboardFrame } from 'components';
 
-import CollectionOverview from './CollectionOverview';
-import CommunityOverview from './CommunityOverview';
+import { useCollectionState, useCollectionPubs } from './collectionState';
+import OverviewDetails from './OverviewDetails';
```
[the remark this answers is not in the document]
isTravis: 🎉

client/containers/DashboardOverview/OverviewBlocks.js
isTravis: Happy to try that. I find with reducers, keeping the logic of how the loop is iterating can be really helpful. Sometimes seeing `prev` is the most useful name for me because I know it's the thing that got passed on from the last loop. In more complex reducers with many variables, `count` may not stand out as loudly. That could just be how I've become used to writing reducers more than sound logic - and for this simple case I agree it reads cleaner with `count, pub`.

client/components/DashboardFrame/DashboardFrame.js
```diff
@@ -15,12 +19,20 @@ const defaultProps = {
...
 
 const DashboardFrame = (props) => {
-	const { className, children, controls, details, title } = props;
+	const { className, children, controls, details } = props;
+	const { scopeData, locationData } = usePageContext();
+	const {
+		elements: { activeTargetType },
+	} = scopeData;
+	const { path } = locationData;
+	const title = capitalize(path.split('/').pop());
```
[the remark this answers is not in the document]
isTravis: Ya - good point. Agreed and changed.

client/containers/DashboardOverview/ContentOverview.js
isTravis: Agreed. Fixed with refactor in latest commit.

client/containers/DashboardOverview/OverviewControls.js
isTravis: Agreed. Fixed with refactor in latest commit.

client/containers/DashboardOverview/collectionState.js
isTravis: Yup - confusing indeed. Fixed with refactor in latest commit.

client/containers/DashboardOverview/OverviewBlocks.js
isTravis: Agreed. Fixed with refactor in latest commit.

PR #759, opened 2020-04-22
new: Heap tracking

gabestein:
Adds Heap tracking for both dev and prod, including the ability to opt out from tracking and small updates to the privacy policy.

The way Heap works is it assigns a relatively persistent anonymous `user_id` to every visitor. We can add an additional `identity` field from our side, and it will automatically merge sessions from the anonymous `user_id` when we supply an `identity` – useful for tracking logged in vs. logged out, and across devices. Also nice: we can strip or further anonymize the `identity` from downstream analytics in redshit etc. without losing the ability to match sessions.

So, this PR sets the PubPub user id as the Heap `identity` if a user is logged in and consents to tracking. 

Heap also allows you to set event properties. I've added an event property called `loggedIn` that shows true when users are logged in and consented and does not appear when users are logged out (or not consented), so we can differentiate across logged out and logged in events.

While logged in, opting out of tracking has the effect of removing the `identity` field and resetting your `user_id`, effectively breaking the tracking chain forever. 

When logged out, however, the anonymous `user_id` that Heap sets is hard to unset. So, if you're logged out and you opt out of tracking, I re-identify you with a number generated by Math.random and then tell Heap to unset that `identity`. This gives you a new anonymous `user_id`, breaking the tracking chain.

I've also included passing the same IDs (community, collection, pub, branch) we pass for Keen as event properties so we can eventually match to the product db in analytics tools.

_Test plan_
- Move `setupHeap();` from line 34 to line 31 in `client/utils/hydrateWrapper.js` to enable logging on local environments
- Load site in dev environment. View Define -> Live tracking on Heap for dev environment.
- Click around on dev and verify that Heap is tracking events
- Click into a pub and verify that communityId, pubId and branchId are set in events logged
- Click to the homepage and verify that pubId and branchId are unset
- Accept tracking cookies and login.
- Verify that logged in is set to true in the event. 
- Visit the Users tab and verify that a user with an identity that looks like a PubPub ID has visited.
- Visit `/legal/settings`
- Toggle settings to off.
- Refresh the page
- Heap live view should show a new, anonymous user id for subsequent refreshes. Logged in should not be set on the event.
- Toggle settings to on
- Refresh the page.
- Visit the Users tab and verify that the user with the same PubPub ID identity has now made more visits.
- Load in the production environment and verify that events are now logged on production.

### DIFF
--- a/client/containers/Legal/PrivacyPolicy.js
+++ b/client/containers/Legal/PrivacyPolicy.js
@@ -91,20 +91,22 @@ const PrivacyPolicy = function() {
 				users. Any such internal analytics will be conducted on an anonymized set of
 				user-generated content.
 			</p>
-			<h3>Keen</h3>
+			<h3>Keen & Heap</h3>
 			<p>
-				We use Keen software to perform Site usage analytics. Keen collects anonymous
-				information from users to help us track Site usage and referrals from other
-				websites. These data are used primarily to optimize the website experience for our
-				visitors, but we may use the data as well to assist us in our marketing of the Site.
+				We use Keen and Heap software to perform Site usage analytics. Keen and Heap collect
+				anonymous information from users to help us track Site usage and referrals from
+				other websites. These data are used primarily to optimize the website experience for
+				our visitors, but we may use the data as well to assist us in our marketing of the
+				Site.
 			</p>
 			<p>
-				Information collected and processed by Keen include the user’s IP address, network
-				location, and geographic location. Keen acquires all its information directly from
-				the user, by installing a cookie (see below) on JavaScribp3-enabled computers. The
-				Site does not share any information it collects with Keen, and Keen does not collect
-				any personal identifying information such as names, contact information, social
-				security numbers or financial information.
+				Information collected and processed by Keen and Heap include the user’s IP address,
+				network location, and geographic location. Keen and Heap acquires all its
+				information directly from the user, by installing a cookie (see below) on
+				JavaScript-enabled computers. The Site does not share any information it collects
+				with Keen and Heap, and Keen and Heap do not collect any personal identifying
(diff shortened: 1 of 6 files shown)


client/containers/Legal/PrivacyPolicy.js
```diff
@@ -91,20 +91,22 @@ const PrivacyPolicy = function() {
...
+				anonymous information from users to help us track Site usage and referrals from
+				other websites. These data are used primarily to optimize the website experience for
+				our visitors, but we may use the data as well to assist us in our marketing of the
+				Site.
+			</p>
+			<p>
+				Information collected and processed by Keen and Heap include the user’s IP address,
+				network location, and geographic location. Keen and Heap acquires all its
+				information directly from the user, by installing a cookie (see below) on
+				JavaScript-enabled computers. The Site does not share any information it collects
```
idreyn, line 106: lol at "JavaScript-enabled computers"

client/containers/Legal/PrivacyPolicy.js
```diff
@@ -91,20 +91,22 @@ const PrivacyPolicy = function() {
...
-				visitors, but we may use the data as well to assist us in our marketing of the Site.
-			</p>
-			<p>
-				Information collected and processed by Keen include the user’s IP address, network
-				location, and geographic location. Keen acquires all its information directly from
-				the user, by installing a cookie (see below) on JavaScribp3-enabled computers. The
-				Site does not share any information it collects with Keen, and Keen does not collect
-				any personal identifying information such as names, contact information, social
-				security numbers or financial information.
+			<h3>Keen & Heap</h3>
```
idreyn, line 94: A new buddy comedy, coming summer 2020

PR #766, opened 2020-04-27
update: Make searchSync and searchUtils work with new Member model

isTravis:
This PR makes several changes to update our search permissioning based on the new Members model. The Members model simplifies the handling of private data by simply including an `accessIds` array on each algolia object which is filtered against. 

This does not yet account for Pubs who have their draft set to be publicly visible, as the PublicPermissions model is still in development.

### DIFF
--- a/client/components/PubPreview/pubPreviewUtils.js
+++ b/client/components/PubPreview/pubPreviewUtils.js
@@ -41,9 +41,22 @@ export const generatePlainAuthorString = (pubData) => {
 	if (!pubData.attributions) {
 		return '';
 	}
-	const authors = pubData.attributions.filter((attribution) => {
-		return attribution.isAuthor;
-	});
+	const authors = pubData.attributions
+		.filter((attribution) => {
+			return attribution.isAuthor;
+		})
+		.map((attribution) => {
+			if (attribution.user) {
+				return attribution;
+			}
+			return {
+				...attribution,
+				user: {
+					id: attribution.id,
+					fullName: attribution.name,
+				},
+			};
+		});
 	return authors
 		.sort((foo, bar) => {
 			if (foo.order < bar.order) {
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
 		"@storybook/addon-storyshots": "^5.0.11",
 		"@storybook/addon-viewport": "^5.0.11",
 		"@storybook/react": "^5.0.11",
-		"algoliasearch": "^3.32.1",
+		"algoliasearch": "^4.2.0",
 		"amqplib": "^0.5.3",
 		"autoprefixer": "^9.5.0",
 		"aws-sdk": "^2.444.0",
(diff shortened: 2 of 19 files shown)


client/components/PubPreview/pubPreviewUtils.js
```diff
@@ -41,9 +41,22 @@ export const generatePlainAuthorString = (pubData) => {
...
 		return '';
 	}
-	const authors = pubData.attributions.filter((attribution) => {
-		return attribution.isAuthor;
-	});
+	const authors = pubData.attributions
+		.filter((attribution) => {
+			return attribution.isAuthor;
+		})
+		.map((attribution) => {
```
idreyn, line 48: or `.map(ensureUserForAttribution)`

package.json
```diff
@@ -44,7 +44,7 @@
 		"@storybook/addon-storyshots": "^5.0.11",
 		"@storybook/addon-viewport": "^5.0.11",
 		"@storybook/react": "^5.0.11",
-		"algoliasearch": "^3.32.1",
+		"algoliasearch": "^4.2.0",
```
idreyn, line 47: Have you been able to test this? I think I would caution against a major version bump of a dependency at this time, unless there's a specific reason for it.

workers/utils/searchUtils.js
```diff
@@ -79,115 +80,67 @@ export const getPubSearchData = (pubIds) => {
...
+				pubId: pub.id,
+				title: pub.title,
+				slug: pub.slug,
+				avatar: pub.avatar,
+				description: pub.description,
+				byline: authorByline ? `by ${authorByline}` : '',
+				communityId: pub.community.id,
+				communityDomain: pub.community.domain || `${pub.community.subdomain}.pubpub.org`,
+				communityTitle: pub.community.title,
+				communityAccentColorLight: pub.community.accentColorLight,
```
idreyn, line 108: I think we can drop the properties from here that aren't slugs or natural language strings.

workers/utils/searchUtils.js
```diff
@@ -197,46 +150,55 @@ export const getPageSearchData = (pageIds) => {
...
+		// eslint-disable-next-line no-await-in-loop
+		const { members } = await getMembers({ scopeData: scopeData });
+		const accessIds = members.map((member) => {
+			return member.userId;
+		});
+		const data = {
+			pageId: page.id,
+			title: page.title,
+			slug: page.slug,
+			avatar: page.avatar,
```
idreyn, line 168: likewise here re. non-NL strings.

package.json
isTravis: The reason for the bump was because I was getting really odd behavior using the old package. Algolia records with `isPublic: false` (as confirmed in the web dashboard) were coming down with `isPublic: true`. I wiped the Algolia indexes, re-uploaded records, tested in new browsers, and spent an hour or so debugging the strange behavior. It was a further problem because private information was leaking into results for non-logged-in users (the filters on isPublic were failing). 

Upgrading the package fixed the issue (though there was a migration guide that took some time to follow). I've tested the upload and search with the version bump and it seems to be good to go.

workers/utils/searchUtils.js
isTravis: Our search cycle doesn't hit our postgres DB at all. The client directly queries algolia for search results.  Anything that we want to render in the results has to be stored in algolia records.

client/components/PubPreview/pubPreviewUtils.js
isTravis: 👍 Fixed in 60911d1

PR #767, opened 2020-04-27
fix: Discussion Visibility and syncing bugs

isTravis:
This PR sets the Visibility of new Discussion objects based on their branch of origin. If created on a 'public' branch (i.e. a Release), the Visibility is set to 'public'. In all other cases, the Visibility is set to 'members'.

This PR also disables the live syncing of discussion objects over the discussionsContentLive object on firebase. A new live-syncing approach will be implemented shortly.

### DIFF
--- a/server/discussion/queries.js
+++ b/server/discussion/queries.js
@@ -7,7 +7,6 @@ import {
 	ThreadComment,
 	ThreadEvent,
 	Visibility,
-	Branch,
 } from '../models';
 
 const findDiscussionWithUser = (id) =>
@@ -121,12 +120,8 @@ export const createDiscussion = async (inputValues, user) => {
 		threadId: newThread.id,
 	});
 
-	const branchData = await Branch.findOne({
-		where: { id: inputValues.branchId || null },
-		attributes: ['id', 'title'],
-	});
 	const newVisibility = await Visibility.create({
-		access: (branchData || {}).title === 'public' ? 'public' : 'members',
+		access: inputValues.visibilityAccess,
 	});
 	const newDiscussion = await DiscussionNew.create({
 		id: inputValues.discussionId,
--- a/server/discussion/queries.js
+++ b/server/discussion/queries.js
@@ -9,7 +9,6 @@ import {
 	Visibility,
 	Branch,
 } from '../models';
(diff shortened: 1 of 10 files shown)


server/discussion/queries.js
```diff
@@ -120,8 +121,13 @@ export const createDiscussion = async (inputValues, user) => {
 		userId: user.id,
 		threadId: newThread.id,
 	});
+
+	const branchData = await Branch.findOne({
```
idreyn, line 125: Without a corresponding update to `permissions.js` this will allow anyone who knows the ID of the draft branch to write a members-only comment. I confess this is unlikely, but I would propose the following structure:

On the client, send `visibility: pubData.isRelease ? 'public' : 'members'` as part of the payload to the API request. Then pass `visibility` into `getPermissions()`, and check whether the logged-in user can create a discussion on the Pub with the given visibility.

This is closer to our new mental model of PubPub and the code we'll need to write to support user-controlled discussion visibility anyway. Branches are sort of a vestigial part of our IA now so we should try to avoid passing around references to them from here on out.

server/utils/firebaseAdmin.js
```diff
@@ -109,17 +109,21 @@ export const mergeFirebaseBranch = (pubId, sourceBranchId, destinationBranchId)
...
-	const existingDiscussion = await discussionsRef
-		.orderByChild('id')
-		.equalTo(discussionId)
-		.once('value');
-	if (existingDiscussion.exists()) {
-		const childKey = Object.keys(existingDiscussion.val())[0];
-		return existingDiscussion.ref.child(childKey).update(discussion);
-	}
-	return discussionsRef.push(discussion);
+export const updateFirebaseDiscussion = async () => {
```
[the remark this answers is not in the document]
isTravis: ++ Removed in 7826f4d

server/discussion/queries.js
isTravis: Good catch. I made changes in line with your suggestion in e28b77d. Testing shows it setting the right `access` value and rejecting those who try to set `members` without authorization. 

I'm going to resolve and merge this, but feel free to continue the conversation if you recommend additional refactors to this.

PR #772, opened 2020-04-29
fix: access-hash redirect bug

isTravis:
This PR fixes an issue with pub redirects that are inbound with an access hash. 

Access hashes used to live on Branches, and now live on Pubs. If the inbound accessHash matches the `viewHash`, `discussHash`, or `editHash`, the Pub's respective accessHash is placed in the redirect route. 

`discussHash` has been deprecated as `viewHash` on Pub confers discuss privileges.

**Test**
Test the functionality (on the auto-generated review app if you wish!) by navigating to `/pub/plf0syc4/branch/2?access=7mawlz01`. The `7maw...` hash was the `editHash` on the draft branch before migration. You should land on the Pub draft with edit permissions.

### DIFF
--- a/server/routes/pubRedirects.js
+++ b/server/routes/pubRedirects.js
@@ -82,7 +82,7 @@ app.get(
 			const prefix = `${domain}/pub/${slug}`;
 			const pubData = await Pub.findOne({
 				where: { slug: slug },
-				attributes: ['id'],
+				attributes: ['id', 'slug', 'viewHash', 'editHash'],
 				include: [
 					{ model: Branch, as: 'branches' },
 					{ model: Release, as: 'releases' },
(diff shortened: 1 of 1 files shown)


server/routes/pubRedirects.js
```diff
@@ -109,8 +117,8 @@ app.get(
 					: res.redirect(`${prefix}/release/${pubData.releases.length}`);
 			}
 			return versionNumber
-				? res.redirect(`${prefix}/draft/${Number(versionNumber) + 1}`)
-				: res.redirect(`${prefix}/draft`);
+				? res.redirect(`${prefix}/draft/${Number(versionNumber) + 1}${accessHashString}`)
```
idreyn, line 120: Depending on the details of what goes into `prefix` you might find the `pubUrl()` util to be a good abstraction here.

isTravis: We don't have an actual community object to pass to `pubUrl()` here. We could query for that, but it felt heavier than I wanted for a redirect route.

Issue #782, opened 2020-04-30
blank spot for article description


KFGBot:
_We use GitHub Issues for tracking active code development. If you have general questions or feedback about PubPub, we invite you to use our [Discourse forum](https://discourse.knowledgefutures.org)._

### Required
---

**What went wrong, step-by-step?**
1. when there is no article description and the header text has a highlight design, it appears that theres a blank highlighted spot where the description would be, below the title 
2. if there's no description this should not exist at all
...

**What URL can we use to see the issue?**
example: https://hdsr.mitpress.mit.edu/pub/dgg03vo6/release/1

**What browser/operating system were you using?**

### Optional

---

**Have any screenshots? (Attach them by dragging the image here)**

**Who reported this? (If not you)**
CA

Issue #773, opened 2020-04-29
Block Pub header spacing a bit wonky with some longer titles


gabestein:
_We use GitHub Issues for tracking active code development. If you have general questions or feedback about PubPub, we invite you to use our [Discourse forum](https://discourse.knowledgefutures.org)._

### Required
---

**What went wrong, step-by-step?**
With certain titles set to white/black block in the new Pub header, the blocks get awkwardly very close to the new pub header menu. Seems to be just a function of wrapping of certain longer titles.

**What URL can we use to see the issue?**
https://notes.knowledgefutures.org/pub/april2020-newsletter/release/1

**What browser/operating system were you using?**
Chrome 81 / Mac 10.15.4 (Catalina)

### Optional

---

**Have any screenshots? (Attach them by dragging the image here)**
<img width="1185" alt="Screen Shot 2020-04-29 at 11 47 21" src="https://user-images.githubusercontent.com/639110/80616881-6ad97480-8a0f-11ea-8bf2-e955966013da.png">

**Who reported this? (If not you)**

PR #785, opened 2020-04-30
Small header fixes

gabestein:
Fix for #782 and #773

Screen Shot 2020-04-30 at 17 44 55

vs.

Screen Shot 2020-04-30 at 17 35 55

I wonder if there are better ways to fix the grid overflow? But after exploring a few options, this seemed like the best way to preserve the padding without changing the de facto spacing in most cases.

### DIFF
--- a/client/components/DashboardFrame/DashboardFrame.js
+++ b/client/components/DashboardFrame/DashboardFrame.js
@@ -34,8 +34,12 @@ const DashboardFrame = (props) => {
 			<div className="dashboard-content-header">
 				<div className="dashboard-header-left">
 					<div className="title">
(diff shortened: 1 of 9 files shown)


idreyn:
Thanks for taking care of this! I looked at the code and ended up making a few changes:

- There's precedent for using JS instead of CSS to avoid rendering an empty description, only there is a mistake in that logic. So I fixed that instead.
- I made the `column-gap` apply only to blocks-themes; with the non-blocks themes, the gap is unnecessary and, since the header text is so enormous, we really need every pixel of horizontal space we can get.

Hope I'm not out of line going in and mucking up your PR; it felt easier to just do it myself.

gabestein:
Nope, I appreciate it!

PR #786, opened 2020-04-30
fixes: pub discussions touch-up

idreyn:
After the Dashboard migration, there's some stuff we can clean up around discussions:

- Re-enabling discussion labels with some permissions tweaks
- Applying a modicum of styling to the discussion management buttons
- Some props cleanup — it is not necessary to be passing `firebaseRef` and `collabData` deep into the discussions render tree.

Here is a discussion with the new buttons layout — I think the text labels improve the buttons:
image

_Test plan:_
- Create a discussion from both the pub bottom and the pub document
- Edit a ThreadComment
- Archive a discussion
- As an admin, un-archive a discussion
- Re-anchor a discussion
- As an admin, create a label
- As an admin, apply a label to a discussion
- As an author, apply a label to a discussion
- Run `npm run test server/discussion/`

### DIFF
--- a/client/components/Thread/ThreadComment.js
+++ b/client/components/Thread/ThreadComment.js
@@ -13,7 +13,8 @@ const propTypes = {
 };
 
 const ThreadComment = (props) => {
-	const { author, content, createdAt } = props.commentData;
+	const { commentData } = props;
+	const { author, content, createdAt } = commentData;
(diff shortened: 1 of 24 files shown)


idreyn:
Thanks — both of those issues should be fixed now.

PR #787, opened 2020-05-01
fix: Upward permission checking

isTravis:
`scopeGet` handles trickle down permissions. So, if a user is a Community member,

```
Community [isMember]
	Collection []
		Pub []
```

their community-scope permission will be applied to all children elements.

If however, a user is a Pub member,

```
Community []
	Collection []
		Pub [isMember]
```

when viewing the Community dashboard, Collection dashboard, a Page, or a User Profile (the last two defaulting to Community-scope), Pub-level permissions are not calculated and fed upward. The user will not see the Pub in any of these views even though they have permissions to the pub.

`scopeGet` was built around 'can you see this thing' - not 'can you see this thing or any of its children'. This becomes an issue for 'overview' views - any view that's showing an overview of content items, rather than focusing on the scope-level target. So, this isn't an issue (that I can see) for API routes or Pub views, as those are always focused on the specific `activeTarget` element.

Content (Collections or Pubs) is stripped out using sanitize functions (pubSanitize.js and collectionSanitize.js). My thinking is that we have to add a check in each sanitize function that says 'if !canView from activeScope permissions, check for individual element permission'.

Querying scopeGet for each pub and then filtering based on that led to 16s sanitize times locally (2ms originally), ~2s when on a production server. So - that was a dead-end performance wise.

Adding Members to pubs and filtering on that led to no discernible change in query time. (< 1ms)
Adding Members to collections and filtering on that led to no discernible change in query time. (< 1ms)

When we fully implement PublicPermissions at Collection and Community level, the same sort of upward-permission-checks will be necessary.

_Test Plan:_
- Use the Review App deployment below and log in with a Community Admin account (user1) in one window and a non-admin account (user2) in an incognito window
- Assign user2 as a Pub-level Member and verify that they can view the Pub from the Community Dashboard Overview pane even though they only have pub-level permissions
- Verify the same for viewing a Pub on a Page, on the User's profile, and at the Collection Dashboard Overview pane 
- Verify that the above is not true on the same routes currently live on duqduq.org

### DIFF
--- a/server/utils/queryHelpers/collectionSanitize.js
+++ b/server/utils/queryHelpers/collectionSanitize.js
@@ -1,14 +1,10 @@
 export default (collection, initialData) => {
-	/* Collection access is granted when */
-	/* 1. it is public collection, or */
-	/* 2. the user is a community admin, or */
-	/* 3. the user has explicit permissions on the collection */
-
-	const { activePermissions, memberData } = initialData.scopeData;
-	const { canAdminCommunity } = activePermissions;
-	const hasCollectionAccess = memberData.find((member) => {
-		return member.collectionId === collection.id;
+	/* Collections are included in all communityData in initialData */
+	/* communitySanitize implements collection filtering, so this */
+	/* function simply needs to check if that collection exists */
+	/* within initialData.communityData */
+	const hasAccess = !!initialData.communityData.collections.find((cl) => {
+		return cl.id === collection.id;
 	});
-	const hasAccess = canAdminCommunity || !!hasCollectionAccess || collection.isPublic;
 	return hasAccess ? collection : null;
 };
--- a/server/utils/queryHelpers/collectionSanitize.js
+++ b/server/utils/queryHelpers/collectionSanitize.js
@@ -3,7 +3,7 @@ export default (collection, initialData) => {
 	/* communitySanitize implements collection filtering, so this */
 	/* function simply needs to check if that collection exists */
 	/* within initialData.communityData */
-	const hasAccess = !!initialData.communityData.collections.find((cl) => {
+	const hasAccess = initialData.communityData.collections.some((cl) => {
 		return cl.id === collection.id;
 	});
 	return hasAccess ? collection : null;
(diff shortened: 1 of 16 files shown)


server/utils/queryHelpers/collectionSanitize.js
```diff
@@ -1,14 +1,10 @@
...
-
-	const { activePermissions, memberData } = initialData.scopeData;
-	const { canAdminCommunity } = activePermissions;
-	const hasCollectionAccess = memberData.find((member) => {
-		return member.collectionId === collection.id;
+	/* Collections are included in all communityData in initialData */
+	/* communitySanitize implements collection filtering, so this */
+	/* function simply needs to check if that collection exists */
+	/* within initialData.communityData */
+	const hasAccess = !!initialData.communityData.collections.find((cl) => {
```
idreyn, line 6: Tiny nit: `Array.some()` might communicate intent more clearly here.

server/utils/queryHelpers/pubSanitize.js
```diff
@@ -15,11 +15,36 @@ const sanitizeHashes = (pubData, activePermissions) => {
...
+			return currCp.collectionId === cl.id;
+		});
+		const hasCurrCollectionMemberAccess = !!currCollection.members.find((member) => {
+			return member.userId === initialData.loginData.id;
+		});
+		return prev || hasCurrCollectionMemberAccess;
+	}, false);
+	/* If there are no releases and the user does not have view access, they don't have scope-level */
+	/* We then must check if they have pub-level access and community-level access, otherwise */
+	/* we return null. Returning null will cause a 404 error to be returned. */
```
idreyn, line 40: Nit — `Returning null will cause a 404 error to be returned` might be more misleading than helpful; this code has several callers and it's up to them to decide what happens when `null` is returned.

server/utils/queryHelpers/collectionSanitize.js
isTravis: Yes - thank you! I always forget about `.some()`. Changed that here and in `pubSanitize`.

server/utils/queryHelpers/pubSanitize.js
isTravis: Good catch. Changed.

PR #790, opened 2020-05-04
update: Citation parser input types 

isTravis:
This PR includes three updates to the editor's citation parsing capabilities:

1. Enable PubMed ID parsing
2. Enable ISBN parsing
3. Disable @else/url parser (i.e. arbitrary URL parsing)

See Freshdesk ticket #1308 for more context.

_Test Plan_
- Visit a Pub draft
- Confirm that bibtex and DOI parsing still function as expected
- Confirm that ISBN (-10 and -13) parsing produces a citation output
- Confirm that PubMed ID parsing produces a citation output.

### DIFF
--- a/client/components/FormattingBar/controlComponents/ControlsFootnoteCitation.js
+++ b/client/components/FormattingBar/controlComponents/ControlsFootnoteCitation.js
@@ -94,7 +94,7 @@ const ControlsFootnoteCitation = (props) => {
 			<div className="title">Structured Data</div>
 			<textarea
 				className="structured-data"
-				placeholder="Enter bibtex, DOI, wikidata url, PubMed ID, ISBN, or bibjson..."
+				placeholder="Enter bibtex, DOI, wikidata url, or bibjson..."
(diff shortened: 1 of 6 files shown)


gabestein:
DOI parsing looks good, and takes care of the security issue.

Was our intention to add the ability to do lookups by PubMed ID and ISBN, or is that just part of the workaround?

isTravis:
The intention was to add in PubMed and ISBN support because they were mentioned in the ticket, and it seemed like a simple and quick bonus. I must have gotten lucky when testing - as ISBN and PubMed ID both worked for me when tested locally. I can still get some ISBNs to resolve (see image below) - but it seems like a very small percentage of PubMed IDs resolve. It's likely not worth tracking down the cause of the sporadic coverage right now, so I removed the ISBN and PubMed bits.

Screen Shot 2020-05-06 at 9 16 00 AM

gabestein:
Bummer. Woulda been nice to get that. Looks good now.

PR #798, opened 2020-05-05
dev: Add custom public discussion disable

isTravis:
This PR adds a simple manual community-level override for `canCreateDiscussions` which will disable the public ability to create discussions.

This will eventually be more properly implemented through the PublicPermissions table.


gabestein:
Lgtm! Thanks for the workaround.

Issue #792, opened 2020-05-04
Created on in pub details shows wrong date


gabestein:
ie: https://notes.knowledgefutures.org/pub/b6z5z2v5/draft

Seems to be taking first release date or today's date rather than date created on.

PR #805, opened 2020-05-06
update: Add customPublishedAt field

isTravis:
This PR addresses two items related to Pub dates:

1. It adds a `customPublishedAt` field to the Pubs table. For now, this can only be manually set by someone on the PubPub team, but we can eventually add a user-facing field to set the value. There are some technical and permission-focused questions I'd like us to discuss before adding the user-facing UI though.
2. It fixes the handling of dates as outlined in #792. The createdAt date now only shows if you have at least `canView` permissions on a pub. The publishedAt date will always show, taking the `customPublishedAt` field if it exists, and showing `Unpublished` if necessary. The updatedAt field is unchanged. The image below shows the dates as listed with a `customPublishedAt` date.

Screen Shot 2020-05-06 at 10 08 14 AM


gabestein:
👍

Issue #800, opened 2020-05-05
Collection search broken


gabestein:
### Required
---
This is a two-parter.

First, it looks like pre-filled search pages are no longer working. The page loads, but spins forever and no request is sent to our API. See: https://www.pubpub.org/search?q=hello

Second, it looks like we're not searching over and/or indexing collections anymore, which has had the effect of breaking collection "pages" if you don't explicit link them. E.g.:

**What went wrong, step-by-step?**
1. Visit https://www.keepcalmlogon.com/pub/digital-security-work-from-home-remote/release/2
1. Click security-all
1. (delete an l and add an l to resolve the issue above)
1. Notice that there are no results, despite the article being in that collection (and the collection has 24 pubs, so we would expect more)
...

**Who reported this? (If not you)**
Gus / Keep Calm

isTravis:
The first part is addressed in #806. 

To my knowledge, the second part _never_ worked. Collections were never indexed by search, and clicking a Collection tab without a linked page always took you to a Search page that had shoddy results (i.e. it wasn't showing you the list of pubs within that collection). Navigating to the search page was a minimal attempt at answering: 'Where do we send folks that click on unlinked-collection buttons'. We could send them to the Collection Dashboard now if that seemed right.

gabestein:
Hrm...I don't think sending folks to the dashboard seems quite right. Let's close this with the PR and return to it in the great Collection/Pages debate of 2020.

PR #806, opened 2020-05-06
fix: Search on page load bug

isTravis:
This PR fixes the issue where loading Search with a query in the URL would not launch a search request. This addresses part of #800.

_Test Plan:_
- Navigate to the review app below
- append `/search?q=a` to the URL and verify that the search completes with results.


gabestein:
Thanks!

Issue #784, opened 2020-04-30
Support for middle names/initials


pubpubBot:
_We use GitHub Issues for tracking active code development. If you have general questions or feedback about PubPub, we invite you to use our [Discourse forum](https://discourse.knowledgefutures.org)._

### Required
---

**What went wrong, step-by-step?**
1. Currently there's no support for middle names but many authors chose to write with a middle initial or name. This affects how names appear in citations, as raised by an HDSR 2.2 author. His name is Michael B. Hawes but his citation is generated as B. Hawes, M. should be Hawes, M.B.
2.
...

**What URL can we use to see the issue?**
https://hdsr.mitpress.mit.edu/pub/dgg03vo6/draft

**What browser/operating system were you using?**

### Optional

---

**Have any screenshots? (Attach them by dragging the image here)**

**Who reported this? (If not you)**
CA via HDSR

isTravis:
I think there are two separate considerations here:

1. How citations source structured name data
2. How PubPub deals with names for Users

For citations, we likely want to allow authors/shadow-authors to have structured data depending on the types of fields accepted by Crossref and other citation/indexing formats.

Towards the second point, I don't think we want to bake citation-friendly structures into the User profile schema as trying to add structure to names for a tool that has diverse audiences is [notoriously difficult](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/).

gabestein:
Agreed. We're going to need to implement some more structured data based around citation needs to fully fix this.

But as a short-term fix, I think it would be reasonable to slightly change the way we parse the name field for shadow authors. We're currently assuming everything after the first space is the last name. Instead, I think we should look back and assume everything after the last space is the last name, and everything else first + middle. 

e.g. Gabriel H. Stein = Stein, Gabriel H.

Right now it would return H. Stein, Gabriel.

That's obviously not perfect, but since we're already making an assumptions, this feels like a better one to make in most cases for now until we can implement some more structure.

gabestein:
We're going to implement the above for now, but that will not close the issue -- we need a longer term solution.

gabestein:
Duplicate of #1063

PR #807, opened 2020-05-06
update: Change shadow author last name assumption

isTravis:
This pull request implments the temporary fix suggested in #784. For shadow authors, the assumption for last name parsing (as used to generate citation data) is that everything after the final space is the last name. 

A bit of cleanup to use the `ensureUserForAttribution` everywhere is also included.

Finally, the `id` field has been removed from the citation data in `server/utils/citations.js` as it was causing strange caching behavior. APA and Harvard citations weren't being updated to reflect new author information. Cite must keep a working cache in memory that is references with the `id` field.


gabestein:
Looks good!

Issue #794, opened 2020-05-04
Always show contents in right hand side of pub header on desktop


gabestein:
Leave in bottom on mobile

PR #809, opened 2020-05-06
update: Use consistent ToC placement

isTravis:
This PR addresses #794. It sets the contents button to consistently render in the right-hand list for all cases: Release, Draft, and mobile on both.

Screen Shot 2020-05-06 at 2 55 52 PM

### DIFF
--- a/client/containers/Pub/PubHeader/DraftReleaseButtons.js
+++ b/client/containers/Pub/PubHeader/DraftReleaseButtons.js
@@ -9,7 +9,6 @@ import { pubUrl } from 'shared/utils/canonicalUrls';
 import { formatDate } from 'shared/utils/dates';
 
 import ResponsiveHeaderButton from './ResponsiveHeaderButton';
-import PubToc from './PubToc';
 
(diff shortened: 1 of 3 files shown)


client/containers/Pub/PubHeader/DraftReleaseButtons.js
```diff
@@ -136,7 +105,7 @@ const DraftReleaseButtons = (props) => {
 			typeof latestRelease.sourceBranchKey !== 'number' ||
 			latestRelease.sourceBranchKey < latestKey;
 		return (
-			<>
+			<React.Fragment>
```
gabestein, line 108: for my own knowledge: is this a stylistic choice, or is there a substantive difference between the empty el and the fragment?

isTravis: Stylistic at the end of the day. But also the babel syntax highlighter for SublimeText doesn't recognize `<>` as a tag and causes all sorts of highlighting problems.

PR #816, opened 2020-05-08
GitHub gist embed

gabestein:
Got a number of requests for Gists and/or syntax highlighting for code blocks recently from folks coming over from Medium.

This PR allows users to embed Gists via the media bar. Unfortunately, GitHub no longer maintains oembed URLs. So, this pings the GitHub API to verify that the Gist exists, and if so, reconstructs the script tag embed tag and adds it to the iframe src as a data url, along with some styling.

This approach looks basically native for single-file gists, has few security implications, fits well into our existing embed model, and avoids having to build a new block in the editor for Gists. 

But there are two downsides: for multi-file gists, there's an additional scrollbar that's a little wonky (see screenshot). And because we use a data url for the iframe source, changing the source field isn't as straightforward as for other types of iFrames (although other "apps" have a similar problem since they use links from oembed endpoints, not the main site that a user would type in).

I think it's worth it to support some kind of syntax-highlighted code as people are looking at us for those purposes. Ultimately, I think we could improve gist support and build in native syntax highlighting for our code blocks.

_Test plan_
1. Go to a pub
1. Add a gist from a gist url (e.g. `https://gist.github.com/youruser/12345`)
1. Make sure it loads
1. Try something that's not a gist and make sure it does not load
1. Try a malicious URL with some js in it (ie `http://hacker/"%20+%20document.cookie;`) and make sure it doesn't load.

Screen Shot 2020-05-08 at 12 17 02
Screen Shot 2020-05-08 at 12 18 49

### DIFF
--- a/client/components/FormattingBar/media/Media.js
+++ b/client/components/FormattingBar/media/Media.js
@@ -13,6 +13,7 @@ import MediaYoutube from './MediaYoutube';
 import MediaCodepen from './MediaCodepen';
 import MediaVimeo from './MediaVimeo';
 import MediaSoundcloud from './MediaSoundcloud';
+import MediaGithub from './MediaGithub';
 
 require('./media.scss');
 
@@ -50,6 +51,7 @@ class Media extends Component {
 			{ text: 'Codepen', icon: 'codepen' },
 			{ text: 'Vimeo', icon: 'vimeo' },
 			{ text: 'SoundCloud', icon: 'soundcloud' },
+			{ text: 'GitHub Gist', icon: 'github' },
 		];
 		const activeItem = this.state.activeItem;
 		const componentProps = {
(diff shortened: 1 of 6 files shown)


irhum:
Would it be possible to extend/control the width of the embed window (assuming major changes aren't required) to cover at least 79 chars? (basing this off Python's PEP 8 line length). Making a rough count, it currently appears to be 60 chars wide.

The margin on the side on PubPub exists to allow comments directly on text, but it's not possible to comment on the gist (which is an embed) anyway

isTravis:
@irhum Yep! The gist uses the standard iframe embed which supports a 'full-width' option in alignment:

Screen Shot 2020-05-09 at 8 08 05 AM


This will let you set the width of the gist as a percentage of the full-page width:

Screen Shot 2020-05-09 at 8 10 24 AM

PR #821, opened 2020-05-14
Export fixes

### LINKED ISSUE #737: Add license in generated PDFs
_We use GitHub Issues for tracking active code development. If you have general questions or feedback about PubPub, we invite you to use our [Discourse forum](https://discourse.knowledgefutures.org)._

### Required

---

**What's your request?**
Include CC license type in generated pdfs.

Helps when reading a ch or article out of context

All

### Optional

---

Nick at MITP, after conversation with TOME who requested it for Documentarity
Submitted here by CA

### LINKED ISSUE #815: Font sizes for lists in PDFs too large
Likely related to the new dashboard releases. See: https://demo.duqduq.org/pub/v4htuyzb/ and PDF it generates.

Screen Shot 2020-05-06 at 15 15 25

gabestein:
This PR bundles a couple of export fixes.

- Adds license information to all exports (Resolves #737)
- Fixes font sizes for `ol` and `ul` with an override (Resolves #815)
- So that style updates to Pubs carry over automatically to PDF and HTML exports, builds the css for pagedjs directly from the editor styles
- Bumps the pandoc version to the latest (2.9.2.1) to support different jats export targets
- Includes pandoc's templates in the pubpub directory so we can use them to generate various files
- Specifies a template file for most pandoc export types we support, which allows us to create more structured exports
- Specifies pandoc yaml metadata based on the needs of common pandoc templates, with a special version for jats

_Test Plan_
1. Visit a complex pub with many authors, ie: http://hdsr.mitpress.mit.edu/pub/ct67j043/draft
1. Download the pre-generated PDF that already exists
1. Go to draft, make and then undo a change, and export each file type
1. For PDF and HTML, verify that ul and li sizes are the same as main text, that the license text appears, and that otherwise, the PDF matches the previous download.
1. For other types, verify that they include common metadata, including date published, authors, license, title, collection, etc.
1. Create a pub. Include some shadow authors with affiliations, as well as yourself with affiliations and export the PDF and JATS.
1. Verify that shadow authors and affiliations show up correctly, and that your affiliation shows up correctly.


workers/tasks/export/pandoc.js
```diff
@@ -1,35 +1,78 @@
+import path from 'path';
 import fs from 'fs';
 import nodePandoc from 'node-pandoc';
 import YAML from 'yaml';
 
+import { getLicenseBySlug } from 'shared/license';
 import { getTmpFileForExtension } from './util';
 
-const dataRoot = process.env.NODE_ENV === 'production' ? '/app/.apt/usr/share/pandoc/data ' : '';
+const dataRoot = path.dirname('templates');
```
idreyn, line 9: My guess is that this is intended to resolve the absolute path of the `templates` directory, but this function just shaves off the part of the pathstring inferred to be the directory path and will always return `"."`, no matter what — that happens to work in this case, but I would rather just write `"."` into the `--template` flag below, or use `__dirname`.

workers/tasks/export/styles/printDocument.scss
```diff
@@ -1,19 +1,37 @@
...
+    @import 'components/Editor/styles/main.scss';
+    @import 'components/Editor/styles/collab.scss';
+    @import 'components/Editor/styles/figure.scss';
+    @import 'components/Editor/styles/file.scss';
+    @import 'components/Editor/styles/table.scss';
+}
 
+/* Import some global variables and styles, including fonts */
 @import 'styles/variables.scss';
 @import 'containers/Pub/PubDocument/pubBody.scss';
```
idreyn, line 20: I have a proposed change to this strategy (that I'm happy to implement myself):

- Don't import `pubBody.scss` — instead, move the contents of its `pub-body-styles` mixin to `Editor/styles/main.scss` — where I think it ought to belong — and then remove the dummy `div.pub-body-component` wrapper from the export HTML? That would put all the CSS that's core to document presentation in one place and make the structure of the export documents easier to reason about from reading the code.

- Rather than writing all of the `.editor`-scoped imports out twice (except for the one having to do with figures), create a mixin for the entire set of editor styles (call it `editor-styles`) and use an `@if {}` block to conditionally import the content of `figure.scss`. _That_ file can then be written as bare styles rather than as a mixin.

- Finally, add a single `.editor { @include editor-styles(); }` in `Editor/styles/index.scss` and require that file from `Editor.js`.

workers/tasks/export/styles/printDocument.scss
```diff
@@ -1,19 +1,37 @@
...
-
-@import './editorBase.scss';
+/* Imports the equivalent of components/Editor/styles/base.scss,
+but without the media query mixin, which breaks PagedJS-cli:
+https://gitlab.pagedmedia.org/tools/pagedjs-cli/issues/11 */
+@import '~prosemirror-view/style/prosemirror';
+@import '~prosemirror-gapcursor/style/gapcursor';
+@import '~prosemirror-tables/style/tables';
+
+@import 'components/Editor/styles/latex.scss';
```
idreyn, line 8: Should this be scoped to `.editor` as well?

workers/tasks/export/pandoc.js
gabestein: 👍

workers/tasks/export/html.js
```diff
@@ -107,10 +108,27 @@ const blankIframes = (nodes) =>
...
+	targetPandoc,
+) => {
 	const showUpdatedDate = updatedDateString && updatedDateString !== publishedDateString;
+	const communityAndCollectionString =
+		communityTitle + (primaryCollectionTitle ? bullet + primaryCollectionTitle : '');
+	const license = getLicenseBySlug(licenseSlug);
+
 	return (
 		<React.Fragment>
+			{(targetPandoc === 'docx' || targetPandoc === 'plain' || targetPandoc === 'odt') && (
```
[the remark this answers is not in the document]
gabestein: I think this is gonna be a one-off. These are the formats that either don't have templates at all or, in the case of plain, don't support metadata for issue or journal.

workers/tasks/export/html.js
```diff
@@ -197,6 +225,14 @@ const renderFrontMatterForHtml = ({
 						<strong>DOI:</strong> {doi}
 					</div>
 				)}
+				{license && (
```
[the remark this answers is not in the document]
gabestein: Yeah...almost went that direction, but felt a little awkward to do it just for licenses given that DOI and some of the other stuff is also duplicated. If that's your feeling on the right approach, go for it.

workers/tasks/export/styles/printDocument.scss
gabestein: 👍

workers/tasks/export/styles/printDocument.scss
gabestein: I don't think so? It isn't in the editor base.

idreyn: Thinking that was probably an oversight, since this style comes from the former editor package.

workers/tasks/export/styles/printDocument.scss
idreyn: This ended up being a bit half-baked, as I forgot to account for the fact that not all Editor instances ought to take the pubBody styles, and it's apparently impossible to `@import` within a sass mixin. So I'm going to leave this alone.

PR #822, opened 2020-05-14
fix: Add appropriate empty state for Pub header with no contributors

idreyn:
Some fixes in response to issues that Deepak identified when there are no byline contributors:

- The byline editor button is still accessible from the Pub header via an empty state
- The Pub details pane no longer goes blank
- I changed the language in the empty state of the `PubAttributionsDialog` to talk about "contributors" rather than "attribution" as an abstract, mass noun.

_Test plan:_
- Ensure the byline editor button is visible whether or not there are byline contributors
- Ensure the pub details pane is visible whether or not there are contributors
- Check the byline editor button against all header styles and a range of viewport widths

_Screenshots:_

Light/dark theme:
Screen Shot 2020-05-14 at 11 51 08 AM

Blocks theme:
Screen Shot 2020-05-14 at 11 51 20 AM

New language in attributions modal:
Screen Shot 2020-05-14 at 11 52 44 AM

### DIFF
--- a/client/components/AttributionEditor/AttributionEditor.js
+++ b/client/components/AttributionEditor/AttributionEditor.js
@@ -190,8 +190,8 @@ class AttributionEditor extends Component {
 									hasEmptyState && (
 										<NonIdealState
 											icon="person"
-											title="No attribution yet!"
-											description="Start typing a person's name above to add attribution."
+											title="No contributors yet!"
+											description="Start typing a person's name above to add a contributor."
 										/>
 									)
 								}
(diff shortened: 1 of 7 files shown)


idreyn:
Not sure how I didn't see that as a bug, but fixed!

idreyn:
Regarding the empty-state style — I think visual consistency would be nice, but it's worth noting that the muted style of the "Add description" field is baked into the Blueprint input as a cue that it's a text placeholder. I'm sort of hesitant to either copy its color out into PubPub's codebase, or to override it to something else.

PR #823, opened 2020-05-14
feature: Allow custom inline content for citations

### LINKED ISSUE #813: Allow custom override of inline citation text
Plan is to add a field for inline citation override that, when a value is entered, replaces the entire inline citation text (including any parenthesis or other dividers/spacers).

This allows folks to do more natural citations, e.g.: `(see Stein 2019)` or `(Stein 2019, pg. 10)` or `As Stein says...` within our native system, without needing to add more complex global options.

idreyn:
Resolves #813 by adding a new field to the citation controls to add custom text. Right now it looks a little cramped in there, and I'm going to tweak the styling and solicit some feedback before shipping, but it's functionally all there.

_Test plan:_
- Create a new citation and verify that the usual behavior still works — automatically creating citations from structured data, adding unstructured data, etc.
- Now make sure that you can add and remove custom labels with the new "Inline Content" field.
- Make sure that this information is retained across page refreshes.
- Try this with a non-default inline citation style and ensure that the correct style is rendered into the "Default" state of the inline content controls.

_Screenshot:_
Screen Shot 2020-05-14 at 7 26 26 PM

### DIFF
--- a/client/components/Editor/schemas/citation.js
+++ b/client/components/Editor/schemas/citation.js
@@ -28,8 +28,7 @@ export default {
 		],
 		toDOM: (node) => {
 			const { href, id, count, customLabel } = node.attrs;
-			const { citationsRef } = node.type.spec.defaultOptions;
-			const { citationInlineStyle } = node.type.spec.defaultOptions;
+			const { citationsRef, citationInlineStyle } = node.type.spec.defaultOptions;
 			/*	There is a two-fold approach here. toDOM will render the
 				citations from citationsRef so that server-side and PDF rendering will function as
 				intended (that is, take the citationInlineStyle regardless of node.attrs.label).
--- a/client/components/Editor/schemas/citation.js
+++ b/client/components/Editor/schemas/citation.js
@@ -1,3 +1,5 @@
+import { getCitationInlineLabel } from '../utils/citation';
+
 export default {
 	citation: {
 		atom: true,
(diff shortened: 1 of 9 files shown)


client/components/Editor/schemas/citation.js
```diff
@@ -24,9 +27,9 @@ export default {
 			},
 		],
 		toDOM: (node) => {
-			const { href, id, count } = node.attrs;
-
-			const { citationsRef, citationInlineStyle } = node.type.spec.defaultOptions;
+			const { href, id, count, customLabel } = node.attrs;
+			const { citationsRef } = node.type.spec.defaultOptions;
+			const { citationInlineStyle } = node.type.spec.defaultOptions;
```
isTravis, line 32: Is there a reason for these two destructures from `node.type.spec.defaultOptions` being separated?

idreyn: No, that was pretty silly.

idreyn:
Good call on the label text. The issue with new citations taking on the inline value of their neighbors when they are created is present in production, and the solution to that is probably the same as the collab issues we've been discussing on Slack. I'd rather take care of those two issues in a separate PR.

idreyn:
@isTravis I don't think we have the horizontal space to pull this off on smaller screens or in the version of the formatting bar capped at 1200px width that Deepak has cooking in a parallel PR.

image

PR #825, opened 2020-05-18
Increased width for pub header and pub document

djagdish:
- Wider pub header and pub document to match width of primary header
- Wider panel area for Discussions or other content in the right-side 
- Formatting bar width reduced from full window width to max 1200px 
centered.

Resolves #825

### DIFF
--- a/client/components/FormattingBar/formattingBar.scss
+++ b/client/components/FormattingBar/formattingBar.scss
@@ -17,7 +17,7 @@
 @mixin hide-scrollbar {
     -ms-overflow-style: none;
     scrollbar-width: none;
-    &::-webkit-scrollbar { 
+    &::-webkit-scrollbar {
         display: none;
     }
 }
@@ -161,7 +161,7 @@
         z-index: 2;
     }
     &.full-screen-width:not(.floating) {
-        width: 100vw;
+        width: 100%;
     }
     textarea, .minimal-editor-component {
         border-radius: 4px;
--- a/client/components/Header/header.scss
+++ b/client/components/Header/header.scss
@@ -24,6 +24,8 @@
             display: flex;
             justify-content: space-between;
             white-space: nowrap;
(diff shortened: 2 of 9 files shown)


client/components/FormattingBar/formattingBar.scss
```diff
@@ -161,7 +161,7 @@
         z-index: 2;
     }
     &.full-screen-width:not(.floating) {
-        width: 100vw;
+        width: 100%;
```
idreyn, line 164: This is a big visual improvement, but the popover should probably be the full 1200px width of `header-parts` rather than the width of only the `formatting-bar-component` — the difference between the two is determined by the text of the connection status indicator, which can jump between `connected`, `saving`, and `disconnected`. We don't want the popover to reflow when one of those transitions occurs.

I can pop this open and figure out what the best way to implement it will be — I suspect by nesting the connected indicator inside of the formatting bar itself.

djagdish: Sure thing! I figured someone who has worked with the formatting bar before is probably better placed to take care of this dynamic resizing anomaly. We also had a chat recently about the 'Connected' part going away and probably living someplace else on the pub view. But that's later in the roadmap I suppose.

idreyn:
After playing with this a bit, I think we need to further modify the grid to give the Pub content more breathing room on iPad-sized layouts. Likely what we want to do is let the side column width shrink from its new width of 465px down to its current width of 275px before reducing the Pub content width past 675px. I played with this for a few minutes and realized that we'll probably need to add a new media query and change the way certain values are calculated, and I'd like to give it more attention when I have a bit more time. In the meantime, you're more than welcome to keep playing with this, or I can pick it up sometime soonish, probably next cycle.
image

idreyn:
Closing this as stale for now but we'll know where to find this code when the time finally comes :)

PR #838, opened 2020-05-28
feature: Automatically extract useful metadata from documents on import

### LINKED ISSUE #841: Extract metadata from Pandoc on import

idreyn:
This PR extends the functionality of the importer to grab some useful metadata from Pandoc. This can be inferred from the structure of the document, but could also come from a separate file representing a metadata block if we choose to support that. The fields we will import are:

- **`title`** becomes the Pub title
- **`subtitle`** becomes the Pub description. @gabestein is this right? Pandoc supports both `subtitle` and `description`, but in practice Pub descriptions act more like subtitles?
- **`abstract`** is inserted into the Pub body as an abstract-styled first paragraph
- **`keywords`** and **`categories`** are inserted into the Pub body as comma-separated lists
- **`authors`** becomes a list of Pub attributions.

In the case of `title`, `subtitle`, and `authors`, we provide some new UI in the importer pane that lets users reject or modify these fields before they are added. Author names are checked against existing PubPub users, and these are listed in a dropdown from which the user can choose to link them to the Pub attribution.

_Screenshot:_
image

_Test plan:_
Import this Markdown file and make sure that the appropriate metadata comes through:

```.md
---
title:  'This is the title: it contains a colon'
subtitle: 'This is a subtitle. Or a description?'
author:
- Travis Rich
- Gabriel Stein
- Ian Reynolds
- Another Author
keywords: [kw1, kw2]
categories: [ct1, ct2]
abstract: |
  This is the abstract.
...

Here is some actual content.
```

Also try a variant of this with no authors, and another file with no extractable metadata, and make sure the importer looks nice and works in these cases as well.

Still working out the kinks and expect to find edge cases to seal up as work on the importer continues, so I might merge this into a more durable feature branch rather `master` directly.

Resolves #841


idreyn:
One known issue — the Pub header title doesn't always (ever?) reflow correctly when a long title is added. Will fix this later once some of the hairier import stuff is figured out.

client/containers/Pub/PubDocument/PubFileImport/MetadataEditor.js
```diff
@@ -0,0 +1,275 @@
...
+						</span>
+					),
+				}}
+				buttonContent={`${users.length} matching ${users.length === 1 ? 'user' : 'users'}`}
+				placement="bottom-end"
+			>
+				{users.map((user) => (
+					<MenuItem
+						text={user.fullName}
+						icon={<Avatar width={20} avatar={user.avatar} />}
```
isTravis, line 86: Ideally we would also pass in user.initials to `<Avatar />` in case the user doesn't have an image uploaded.

client/containers/Pub/PubDocument/PubFileImport/MetadataEditor.js
```diff
@@ -0,0 +1,275 @@
...
+});
+
+const propTypes = {
+	onSetMetadataUpdater: PropTypes.func.isRequired,
+	proposedMetadata: PropTypes.shape({
+		attributions: PropTypes.arrayOf(attributionShape),
+	}).isRequired,
+};
+
+const guessInitials = (name) => {
```
isTravis, line 27: It'd be nice if we broke out the firstName/lastName split and initial generation into a single util file. There was a recent PR (#807) addressing the fact that we wanted to change our assumption on how last names are guessed, so it's something we ought to be consistent about.

isTravis:
Really impressive! It's such a nice loop to have the matching users pop up directly - definitely a moment of joy in the UI.

A made a couple comments on non-critical pieces of code, but otherwise good to merge from my view!

client/containers/Pub/PubDocument/PubFileImport/MetadataEditor.js
idreyn: Great idea, I created `shared/utils/names.js` to address this.

PR #839, opened 2020-05-29
feature: Extract citations from JATS files

### LINKED ISSUE #842: Extract bibliography from JATS files using XSLT on import

idreyn:
Thanks to Gabe, who did about 90% of the legwork on finding an approach for this, we will now be able to extract citations from JATS files that we import. Because Pandoc doesn't seem to understand how citations are represented in JATS, we take a rather winding path, using an XSL transformation (again, credit to Gabe) that a JATS file into something that more or less looks like a `.bib` file; this file in turn is treated as if it were uploaded manually.

_Test plan:_
Run the importer against the two files in [this Pub](https://notes.knowledgefutures.org/pub/095607lo/draft) and make sure the citations come through. We should probably find some more to test against; who knows what's out there.

Closes #842


idreyn:
(PS I hope this works on Heroku, waiting to make sure the `xsltproc` binary gets installed properly)

gabestein:
Thanks @idreyn, this is so great!

> Perhaps a question for @gabestein - is it worth creating an Issue on the pandoc repo pointing out that JATS citations aren't supported properly? Will this fix be something we eventually revert due to direct pandoc support?

I don't think so. There's already an issue for solving lack of support of mixed citations (https://jats.nlm.nih.gov/archiving/tag-library/1.1/element/mixed-citation.html), which is Pandoc's main shortcoming when it comes to JATS citation support.

This solves another issue, which is pretty outside Pandoc's remit: JATS citation types (https://jats.nlm.nih.gov/archiving/tag-library/1.1/element/mixed-citation.html) are not directly compatible with bibtex, which causes pandoc-citeproc to throw errors. This is based on https://github.com/PeerJ/jats-conversion, which converts from JATS types to bibtex.

PR #849, opened 2020-06-02
dev: Update sentry to use release info

isTravis:
This PR adds release information to our `Sentry.init` configuration, thanks to the recently discovered [Dyno Metadata lab option](https://devcenter.heroku.com/articles/dyno-metadata). I've set this flag for both `pubpub-v6-dev` and `pubpub-v6-prod`.

I also added a Sentry initialization to `workers/queue.js` which (from what I can tell) was missing altogether.


isTravis:
Great point! `appCommit` is more descriptive and non-conflicting. Updated to use that naming convention.

PR #850, opened 2020-06-02
fix: Add draft to Sharing URLs

### LINKED ISSUE #814: Hard to share view/edit links of drafts where there is a release
_We use GitHub Issues for tracking active code development. If you have general questions or feedback about PubPub, we invite you to use our [Discourse forum](https://discourse.knowledgefutures.org)._

### Required
---
Right now it's hard to send a share a view or edit link to a draft that has an existing release, because the naked pub slug forwards to the latest release rather than the draft. Then, when you click edit draft, it doesn't work because the access code isn't included in the draft url. 

I'm torn on how best to do this – but we should either explicitly include `/draft` in the sharing links when sharing from draft, or we should include the access code in link from release to draft. I think we probably should do the former, since the intended behavior of sharing is to share the draft, not the release, and landing on the release could cause confusion.

1. Visit any pub draft with a public release, ie https://demo.duqduq.org/pub/v4htuyzb/draft
1. Click the share link and copy the edit link
1. Visit the edit link in incognito
1. Get forwarded to the release, where you can't edit

As a temporary, very non-obvious workaround, you can add `/draft` explicitly to the copied share link (ie https://you.pubpub.org/pub/r32jlkr2/draft?access=rfjlkre89)

**Who submitted (if not you)**
Scott / GSU

isTravis:
This PR resolves #814. Sharing URLs now include `/draft` to avoid unwanted redirection to a Release (if it exists).

Screen Shot 2020-06-02 at 3 10 13 PM

### DIFF
--- a/client/components/PubShareDialog/PubShareDialog.js
+++ b/client/components/PubShareDialog/PubShareDialog.js
@@ -46,7 +46,7 @@ const AccessHashOptions = (props) => {
 	const { viewHash, editHash } = pubData;
 
 	const renderHashRow = (label, hash) => {
-		const url = pubUrl(communityData, pubData, { accessHash: hash });
+		const url = pubUrl(communityData, pubData, { accessHash: hash, isDraft: true });
 		return (
 			<ControlGroup className="hash-row">
 				<ClickToCopyButton minimal={false} copyString={url}>
@@ -92,7 +92,7 @@ const MembersOptions = (props) => {
 	const localMembers = membersByType[activeTargetType];
 
 	return (
-		<>
+		<React.Fragment>
 			<p>{getHelperText(activeTargetName, activeTargetType, canManage)}</p>
 			{canManage && (
 				<ControlGroup className="add-member-controls">
@@ -126,7 +126,7 @@ const MembersOptions = (props) => {
 			{!!membersByType.organization.length && activeTargetType !== 'organization' && (
 				<InheritedMembersBlock members={membersByType.organization} scope="Organization" />
 			)}
-		</>
+		</React.Fragment>
 	);
 };
 
@@ -145,7 +145,7 @@ const PubShareDialog = (props) => {
 
 	const renderInner = () => {
 		return (
-			<>
+			<React.Fragment>
 				<div className="pane">
 					<h6 className="pane-title">Members</h6>
 					<div className="pane-content">
@@ -153,15 +153,15 @@ const PubShareDialog = (props) => {
 					</div>
 				</div>
 				{hasHash && (
-					<>
+					<React.Fragment>
 						<Divider />
 						<div className="pane">
 							<h6 className="pane-title">Share a URL</h6>
 							<AccessHashOptions pubData={pubData} />
 						</div>
-					</>
+					</React.Fragment>
 				)}
-			</>
+			</React.Fragment>
 		);
 	};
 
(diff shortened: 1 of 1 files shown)


client/components/PubShareDialog/PubShareDialog.js
```diff
@@ -46,7 +46,7 @@ const AccessHashOptions = (props) => {
 	const { viewHash, editHash } = pubData;
 
 	const renderHashRow = (label, hash) => {
-		const url = pubUrl(communityData, pubData, { accessHash: hash });
+		const url = pubUrl(communityData, pubData, { accessHash: hash, isDraft: true });
```
idreyn, line 49: It would be better to say `isDraft: !pubData.isRelease` here. While the sharing dialog is effectively only shown on drafts right now, the causal relationship is a little indirect, and is down to the fact that `membersData` isn't computed when showing releases. This dialog should be robust to potential changes here.

client/components/PubShareDialog/PubShareDialog.js
```diff
@@ -92,7 +92,7 @@ const MembersOptions = (props) => {
 	const localMembers = membersByType[activeTargetType];
 
 	return (
-		<>
+		<React.Fragment>
```
idreyn, line 95: Matter of taste, but we can use the `<> </>` syntax for these now, which feels cleaner to me.

client/components/PubShareDialog/PubShareDialog.js
isTravis: Agreed. Changed to set `isDraft: !isRelease`

client/components/PubShareDialog/PubShareDialog.js
isTravis: I've had 3 (though now 2) reasons for preferring `<React.Fragment>`:

1. The sublime text babel syntax highligher would fail on `<>` syntax causing the whole file's syntax highlighting to break. Digging into it, I [found an issue](https://github.com/babel/babel-sublime/issues/368) pointing towards a pre-release that fixes this. I've installed that pre-release, so the syntax highlighting isn't an issue for me anymore.
2. I prefer the explicitness of declaring `React.Fragment` because it feels friendlier to folks unfamiliar with JSX or Fragments. If you come across that and don't know what a Fragment is, it's clear that 1) it's part of React, and 2) you could search `React.Fragment` in a search engine to learn more. If you come across `<>` without knowing what it is, it's not immediately clear whether that's part of React, HTML, or some webpack pre-processor. A understandably incorrect search "html empty tag" in google doesn't land you near React Fragments.
3. `<>` doesn't support keys whereas `<React.Fragment>` does. [React docs also hint](https://reactjs.org/docs/fragments.html#keyed-fragments) that future attributes may be supported, which would only be available on the `<React.Fragment>` flavor. My overarching preference is to consistently use one syntax or the other, and given that we already have a few cases where we use `<React.Fragment key={something}>`, I lean towards consistently using that syntax.

isTravis:
Thanks! Responded inline and pushed up a commit to set `draft: !isRelease`.

client/components/PubShareDialog/PubShareDialog.js
idreyn: Thanks, some good food for thought here!

PR #851, opened 2020-06-02
remove PubSuspendWhileTyping

idreyn:
Previously, we introduced the `PubSuspendWhileTyping` component as a performance enhancement that defers expensive React updates while the user is typing into the editor, avoiding low refresh rates or delayed response to keystrokes. On my 2013 MBP it would frequently save 25-100ms of work bound to a single `keypress` task, for the cost of a small React state update to mark when the last keyboard input occurred.

Since then we have also started using `requestIdleCallback()` in `PubSyncManager` to defer these same kinds of updates, which I have long suspected makes `PubSuspendWhileTyping` redundant. Today I discovered that it's actually slowing things down, since the state update it requires takes a few milliseconds with no payoff. So I have retired the component, hung its jersey in the rafters, etc.

_Test plan:_
Try typing into a pub and make sure that (1) it works and (2) it's speedy.

_With `PubSuspendWhileTyping` (note `markLastInput` here):_

Screen Shot 2020-06-02 at 6 02 46 PM

_Without `PubSuspendWhileTyping`:_

Screen Shot 2020-06-02 at 6 04 09 PM


idreyn:
Removed `onKeyPress` — if we ever need it again it will be much easier to add now that the Editor is part of the PubPub repo.

PR #854, opened 2020-06-03
Add support for PDF and TIFF images in imports

### LINKED ISSUE #843: Automatically convert PDF and TIFF images to web-friendly ones on import
Also handle images without explicit extensions

idreyn:
This PR adds importer support for .pdf and .tif(f) files, by way of converting them to .png files on import. I anticipate occasional confusion from folks trying to upload PDFs for their text content, so we'll show this warning when they are queued up without a document we can import:

image

I also enabled references to external image files with the file extension, which seems to be a pattern in LaTeX-world.

_Test plan:_
- Try uploading a file that references .pdf, .tiff, and currently supported (e.g. .png) images by name, with and without file extensions, and verify that they are imported correctly.
- Also verify that a .docx with embedded .tiff images works ([this one](http://assets.pubpub.org/_testing/21591292612062.docx) will do nicely)
- Check that the above error message appears when only .pdfs are queued for import.

Fixes #843.


idreyn:
@gabestein @isTravis FYI I wrote up a real PR description of this yesterday, and it looks like Github clobbered it so it says WIP again. But this is good to review.

client/containers/App/app.scss
```diff
@@ -17,6 +17,10 @@
         line-height: 30px;
         background: #c23030;
         z-index: 99999;
+        &:hover {
+            opacity: 0;
+            pointer-events: none;
+        }
```
isTravis, line 23: This is such a good idea, however it doesn't seem to work consistently for me on Firefox. Hovering the block triggers `pointer-events: none` which causes the `:hover` css to unattach. I get inconsistent behavior that I think may be dependent on how the browser decides to render toggling between `:hover` and not.

idreyn: Yeah, I noticed that too. I'll either fix or remove it.

idreyn:
> One things that confuses me is that the slug size doesn't seem to have changed as much as I would expect given the addition of `xsltproc`, `poppler-utils`, and `imagemagick`. Do buildpack files (and thus Aptfile installs) not count towards the slug size? I thought pandoc was a major chunk of our deployment slug.

My understanding is that these do count towards slug size. `imagemagick` is actually installed on the dynos by default and installing it is a no-op, but I added it here so we don't have any implicit dependencies that might get silently removed. Regardless, they're just compiled binaries, so I guess they're not that big?

PR #860, opened 2020-06-14
dev: utils and import refactors

isTravis:
This PR implements most of the refactor ideas listed in [this discourse post](https://discourse.knowledgefutures.org/t/refactor-inspiration-from-r1/243). The specifics that this refactor adheres to are documented in the updated [STYLEGUIDE.md](https://github.com/pubpub/pubpub/compare/utils-refactor?expand=1#diff-ba2dc84c4bcbe54f368b6c7d3bc1b13e).

This PR makes four kinds of code edits:
1. Moving file locations
1. Breaking `utils/index.js` files into their own specifically named files
1. Editing webpack, storybook, eslint config files to create aliases
1. Editing import paths to account for 1-3

**_Test Plan:_**
To test, run each of the applicable build and server functions:

- Run `npm run lint`
- Run `npm run test`
- Run `npm run storybook`
- Run `npm run prod`
- Run `npm run workers-dev`
- Run `npm start`


isTravis:
Bundle analysis to sanity check that this PR doesn't accidentally include/exclude a large dependency from our previous build. The two are pretty much identical (a couple kb savings in the new build).

Old:
old


New:
new

isTravis:
Those are definitely good points, and I do feel a bit conflicted. I really like the symmetry and explictness of:

```
server/utils
client/utils
shared/utils
```

The place where I circle back comes (I think) from a difference in the implication of the word 'utils'. I don't think of `utils` as low-complexity functions. I understand the perspective of utility functions being simple low-complexity string manipulators or hash generators. However, my interpretation of `utils` is roughly: "any code that is used in multiple locations across the repo that isn't explicitly captured by other code hierarchy structure (e.g. components, containers, models, routes, apis)."

The folder structure below feels off to me because I think of the code in `crossref` as "crossref utilities".
```
shared/
   crossref/
   collections/   
   utils/
```
If the above was the correct path, that would suggest to me that our client folder should instead be:
```
client/
   components/
   containers/
   collections/
   editor/
   ...
   utils/
```
That feels wrong to me because it flattens a top level folder too much in my opinion. We do essentially this in the `server` folder, but the difference for me there is that each of those folders have the same internal structure of `api.js`, `model.js`, `queries.js`, etc. But, FWIW, I do find navigating the `server/` folder more cumbersome than navigating `client/`

If there were some other set of code that was shared that would allow the following, I'd be happy with making the change:
```
shared/
   things/
   utils/
      crossref/
      strings.js
```

PR #865, opened 2020-06-23
tools: Bulk importer

idreyn:
This PR contains the new bulk importer code, along with some other changes to support it:

- Renaming `localPath` to `clientPath` in importer source files to remove ambiguity
- Cleaning up code that deals with S3 and abstracting it away (for easy replacement someday 🤞)
- Refactoring the single-Pub import code currently in production to better interface with the bulk import CLI
- A few other things have been refactored or cleaned up.

This PR is far too large, which I can only partially justify by saying that most of this code isn't "in production" yet — the bulk importer is being run from my laptop for the time being, although we should eventually be able to spin it up on a Heroku dyno. I won't go into any detail about how it works here because it's described in excruciating detail in the included `README.md`.

_Test plan:_
I am mostly interested in testing that the parts of the existing codebase that I refactored still work, specifically:

- Pub step creation
- Single-pub import


client/components/Editor/plugins/collaborative/document.js
```diff
@@ -65,13 +64,11 @@ export default (schema, props, collabDocPluginKey, localClientId) => {
...
 						return undefined;
 					}
-					return {
-						id: uuidv4(), // Keyable Id
-						cId: localClientId, // Client Id
-						bId: ref.key.replace('branch-', ''), // Origin Branch Id
-						s: sendable.steps.map((step) => compressStepJSON(step.toJSON())),
-						t: firebaseTimestamp,
-					};
+					return createFirebaseChange(
```
isTravis, line 67: I don't love the word `keyable` - but it feels slightly more descriptive (at least internally) than `change`. In my mind a `firebaseChange` could be a cursor or metadata update. I think the underlying problem is that we likely just need a satisfying name for the thing we've been calling a `keyable`. That is, a name for the "steps + metadata" object.

idreyn:
> The error looks like something that could be caused by variable name refactors, so it seemed worth flagging.

oh, pff, I had the fix for this sitting here uncommitted on my local branch. thank you for testing!

client/components/Editor/plugins/collaborative/document.js
idreyn: We invented the word `keyable` as shorthand for "`change` or `merge`", though — since this is specifically a `change`, it seems to under-specify what this object is. I am down to find a different name for changes and to propagate that through the codebase, though — I agree the name is vague enough to potentially refer to anything.

isTravis: Ah right. I forgot about the `merge` part of that. It seems like we may be able to deprecate the idea of storing `merges` as firebase objects that get concated into a list of changes now that Releases can work as pointers to a branch-key pair. I'm happy leaving this as is for now, and addressing it again once we figure out what the eventual firebase refactor looks like.

PR #866, opened 2020-06-24
fix: always update search results after user stops typing

### LINKED ISSUE #819: Search hangs until additional text is entered
Mentioned in #pubpub-team:

> _Weird search behavior. I've noticed that when I'm searching for a pub on the notes site, I type in some keywords and the search just spins, but when I hit backspace to try to think of another set of keywords, then the results appear below. Anyone else experiencing this?_

**Who reported this? (If not you)** Heather

3mcd:
This PR resolves #819. A search will now execute at the tail end of user input into the search text box.

### DIFF
--- a/utils/hooks.js
+++ b/utils/hooks.js
@@ -1,4 +1,5 @@
-import React, { useContext } from 'react';
+import React, { useContext, useCallback, useEffect } from 'react';
+import throttle from "lodash.throttle";
 
 export const PageContext = React.createContext({});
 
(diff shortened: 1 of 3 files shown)


utils/hooks.js
```diff
@@ -14,3 +15,11 @@ export const usePageContext = (previewContextObject) => {
 export const usePendingChanges = () => {
 	return useContext(PendingChanges);
 };
+
+export const useCallbackThrottled = (fn, timeout, throttleOptions, deps = []) => 
+	useCallback(throttle(fn, timeout, throttleOptions), deps);
+
+export const useEffectThrottled = (fn, timeout, throttleOptions, args = []) => {
+	const execute = useCallbackThrottled(fn, 1000, throttleOptions);
```
idreyn, line 23: My guess is that you meant to eventually forward `timeout` instead of `1000` into `useCallbackThrottled`?

3mcd: Yep good catch!

PR #869, opened 2020-06-25
update: Improve metadata imports

idreyn:
This PR makes two changes to support upcoming bulk imports:

- Adds a `metadata` field to the `Pub` model that can support a JSON blob of arbitrary metadata, to hold onto later. I'm open to calling this something else.
- Provides the ability to extract metadata from YAML document headers in non-Markdown documents, which Pandoc doesn't seem to support without some kludges.

_Test plan:_

Colocate these two files in the same directory:

test.html:
```html
---
title: This Pub Has a Title
subtitle: It really does, wow
authors:
  - Travis Rich
  - Mystery Author
  - Gabriel Stein
# slug: this-pub-has-a-slug
date: 2019-05-25
pubMetadata:
  hi: hello
  yes: it works
---
<h1>Hello!</h1>
<p>This is a test.</p>
```

config.pubpub.yaml

```yaml
type: community
subdomain: test

children:
  test.html:
    type: pub
    extractMetadataFromDocument: true
    matchSlugsToAttributions: ['gabriel-stein', 'travis-rich']
```

Then run:

```
npm run tools bulkimport -- --actor=pubpub-team --directory path/to/your/files --receipt anywhere/receipt.json
```

Verify that a Pub on test.duqduq.org is created and all the fields in the HTML file header block are correctly inferred and applied to the Pub, as shown below. Check that `pubMetadata` appears in the Pub's database entry as `metadata`.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2551,9 +2551,9 @@
 			"integrity": "sha1-p3c2C1s5oaLlEG+OhY8v0tBgxXA="
 		},
 		"@pubpub/prosemirror-pandoc": {
-			"version": "0.4.0",
-			"resolved": "https://registry.npmjs.org/@pubpub/prosemirror-pandoc/-/prosemirror-pandoc-0.4.0.tgz",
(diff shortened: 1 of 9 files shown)


idreyn:
++ to the automated testing. I have a directory called "imports" on my desktop that is swelling with manual test cases that I would love to set up as unit tests! We need to find works with appropriate licenses/copyright to check into our source, though 🤔

PR #871, opened 2020-06-25
update: Refactor server-side rendering of Prosemirror content

idreyn:
This code was written as part of an exploration of converting image captions from HTML embedded in a `data-` attribute to first-class Prosemirror body text. That effort will have to wait for another day, but in order to close up an edge case in server-side rendering that this caught, I refactored the SSR code into something a little cleaner.

_Test plan:_
I turned off JavaScript, then loaded up HDSR with:
```
PUBPUB_PRODUCTION=true PUBPUB_LOCAL_COMMUNITY=hdsr npm start
```
and saw that every Pub that I could click through to the homepage rendered successfully, including images, equations, and tables.


client/components/Editor/utils/renderStatic.js
```diff
@@ -0,0 +1,128 @@
...
+		}
+		const [tagName, maybeAttrs, ...restItems] = spec;
+		const { attrs, hasAttrs } = getAttrsFromOutputSpec(maybeAttrs);
+		const children = hasAttrs ? restItems : [maybeAttrs, ...restItems];
+		return React.createElement(
+			tagName,
+			{ ...attrs, key: key },
+			...children.map((child, index) => createReactInner(child, `${key}-${index}`)),
+		);
+	};
```
3mcd, line 78: Just out of curiosity, why do we have `createReactInner ` inside of an outer function, when the inner function has no dependencies on the parent scope and recursion is handled internally?

idreyn: hahaha...absolutely no reason. this outer function used to close over a third variable which I since removed, but I forgot to simplify things here. Nice catch!

idreyn:
Dragging this across the finish line in order to integrate it with the other Prosemirror work that's going on right now.

The render tests are really helpful, thanks! After fixing the problems I found, the remaining differences I see between the client and server-side React-rendered HTML are:

- React rendering escapes special characters into HTML entities in the `data-caption` and `alt` fields
- React does not render `undefined` Prosemirror node attributes as empty-string DOM node attributes, which I do not believe is significant
- React renders `<video controls={true}/>` as `<video controls=""/>`, probably as an optimization
- React explicitly closes self-closing tags
- Some client-side nodes have `contenteditable="false"` set by Prosemirror
- There are some attribute casing differences which are significant to React, but not the browser
- The client-rendered table has much more markup, thanks to the table plugin
- The client-rendered inline equation has an extra `<br>` (which I don't think we want)

None of which seem to have a visible effect on the result.

idreyn:
> On, `context` and "we can use this space to explicitly tell our nodes that we want raw HTML to be rendered by React rather than the DOM": do we have situations where we're doing this using the old, implicit pathway? I don't know if I'm understanding the use case right.

I think I sort of explained this on the call this morning, but yes — the old static renderer would attach a `key` attribute to Prosemirror nodes when requesting their `DOMOutputSpec`, as reproduced here:

```js
renderReactFromSpec(
    NodeComponent.toDOM({
        ...node,
        attrs: { ...node.attrs, key: index },
        type: schema.nodes[node.type],
    }),
    index,
    children,
);
```

This was the signal for `renderHtmlChildren` to provide a React element rather than a DOM node:

```js
const renderHtmlChildren = (node, html, wrapperElement) => {
	const hasKey = node.attrs.key !== undefined;
	if (hasKey) {
		return React.createElement(...);
	}
	return document.createElement(...);
};
```

But now, we ignore the `attrs` and just directly tell `renderHtmlChildren` what kind of representation we want.

PR #872, opened 2020-06-25
new: Impact dashboard

gabestein:
Implements embedded impact dashboards from our Metabase install for communities and pubs, with a warning that User data collected before 4/30/2020 used a different system and isn't directly comparable and a coming soon notice for collections (I need to build the collection dash from the ground up because of the way collection data is captured in heap). Dashboards are only visible to users who have `canView` permissions. Users without those permissions, including unauthenticated, are prompted to login or ask their admins for access.

Metabase provides secure embeds via jwt request to its api. To implement this, I added a PubPub API route for requesting the dashboards from Metabase on the server side.

_Test plan_
To test locally, you'll need to update your config.js to include our Metabase api key, now included in 1Pass.

- Visit a community, ie HDSR, while logged into an account with access to it.
- Verify that you can see impact tab at the Community and Pub level.
- Logout and visit an impact tab (or visit from an incognito window). Make sure you see the message prompting to ask the community admin for impact data, and that no requests to the metabase API endpoint are made.
- Make sure you cannot change the Community ID or Pub ID for the dashboards.


gabestein:
Sorry...I actually had an idea, which is to build a "benchmark" dashboard for communities older than 4/30 that shows Users during the overlap period. Will add the tomorrow. But would take a review of the basic mechanism for now.

isTravis:
Nice!

Impact tab shows up for me at each scope, but the dashboard seems to be stuck in a loading loop for Pub and Community scopes while testing on the `test` community.

The loop goes from briefly showing the loading spinner, to showing an empty state image, and then back to the loading spinner.

gabestein:
Heh. Maybe it would help if I added the metabase key to the test app! Try now.

isTravis:
Nice. The test deployment is working for me now.

However - my initial comment came after testing locally with the proper metabase key in my `config.js` too. So, there may be something about the `test` community that's worth poking into still.

gabestein:
Huh. It's loading fine for me locally using `test`. I did notice that the dash displayed warnings even if logged out, though, so I fixed that. Give it another look?

gabestein:
Okay, just pushed a new historical user benchmark for Pubs and Communities, and updated all our historical data to include pubids, which Keen stopped capturing in many cases after the release updates, by joining it with the production db. Totally crazy, but worked out well!

So, it's actually ready for review now.

gabestein:
Nice, thank you. I didn't even think of doing this server-side. Sometimes I forget this is a universal app :(. Much better.

Tooltip and layout changes look good. And the iframe resizer is a freaking miracle!

As discussed, Metabase automatically saves revisions to dashboards that can be reverted. Click edit button -> then the history button. Your version is, I agree, much nicer, so let's launch with that. I updated the Pub dash to match. Let's ship!

PR #873, opened 2020-06-26
update: Pricing page

gabestein:
Updates the pricing page for 2020.

screencapture-localhost-9876-pricing-2020-06-26-14_53_40

Tooltip for pro:
Screen Shot 2020-06-26 at 15 57 09

Tooltip for org:

Screen Shot 2020-06-26 at 15 57 38

Biggest changes:
- Adds pro tier
- Adds member definition
- Lowers free DOIs for all tiers given much lower cost
- Removes custom PDF export and adds it (and some other customizations) to the additional services section.

_Test plan_
- Read and give feedback?
- Click on all the mailto links and make sure they work as expected

### DIFF
--- a/client/containers/Pricing/Pricing.js
+++ b/client/containers/Pricing/Pricing.js
@@ -12,8 +12,8 @@ const Pricing = () => {
 				<h1>Pricing</h1>
 				<p className="description">
 					PubPub's goal is to provide an affordable, high-quality, open-source,
-					institution-owned, hosted alternative to commercial publishing software. We
-					offer a Free tier with no user or publishing limits, as well as a Pro tier for
+					institution-led, hosted alternative to commercial publishing software. We offer
+					a Free tier with no user or publishing limits, as well as a Pro tier for
 					individual communities who need more advanced features (or just want to support
 					our mission), and an Organization tier for those who need to manage multiple
 					communities.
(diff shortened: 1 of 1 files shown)


heatherKFG:
Should say Seat, instead of Member, no? Add definitions somewhere for Seat, Member, Community, Organization?

gabestein:
Ah, forgot to say: member is defined in the tooltip. Added a screenshot. I think it's the right word (vs. seats), since it's what's in the UI.

For simplicity, I'm leaning against defining all the terms here. We introduced those in the latest contract because we were sort of hacking it together, but it seems to me like most of those concepts are contractual and largely not needed for people to at least click get in touch – and also probably likely to change once we rewrite the contract. Once we do have a contract re-written, we should add those terms for sure.

isTravis:
Awesome - it'll be great to have something closer to our current thinking go live. 

I agree with Heather and think 'per Member' will add confusion here. We don't charge for Members, so we shouldn't say we do. If we conflate Member on the pricing page to mean special-paid-member, we'll have to constantly clarify whether we're talking about paid members or all members. I think it'll also cause confusion if people don't hover the tooltip and then go and see the Members tab on the Pub scope. I'd vote we agree on the word that means 'edit-or-above Member at Collection-or-above scope'. I'm comfortable having that word be 'Seat', but it could also be something like 'Active Member' or 'Team Member'.

I agree that we don't need to define the terms on the pricing page.

I also have a couple other consistency/wording updates to suggest. I've put all those in a PR (#875) so you can decide if you want to merge or cherry-pick specific ideas.

gabestein:
Changed lgtm. I think we're ready (pending that flaky test).

Issue #853, opened 2020-06-03
Searching sometimes has strange behavior (Pubs vs Pages) with broken result links   


metasj:
When searching some communities (tested just now on _credcon_) , search indexing seems to have the following quirk:

Whether logged in or logged out, 
* Sometimes searching for a term in "Pubs" returns few or no results, while clicking on the "Pages" tab under the search bar produces results.  
* Sometimes an initial search (under "Pubs") produces results; then clicking on the "Pages" tab produces similar results, and then clicking on the "Pubs" tab no longer shows results (but results continue to appear under "Pages"). 
* In either case above, in the "Pages" results-view, search results include pubs, but the URL of each result is missing "/pub/",  and so does not resolve (generating a "**Page Not Found**" error).  Adding "/pub/" back in makes them resolve as expected.  [Example search](https://credcon.pubpub.org/search?q=citation)

gabestein:
Resolved in #874.

PR #874, opened 2020-06-26
refactor: refactor search and fix issues with pagination and mode transitions

3mcd:
This PR fixes a couple of issues on the Search page with pagination and mode transitions that I suspect are the same issues outlined in #853, specifically the second and third bullet points:

> Sometimes an initial search (under "Pubs") produces results; then clicking on the "Pages" tab produces similar results, and then clicking on the "Pubs" tab no longer shows results (but results continue to appear under "Pages").

and

> In either case above, in the "Pages" results-view, search results include pubs, but the URL of each result is missing "/pub/", and so does not resolve (generating a "Page Not Found" error). Adding "/pub/" back in makes them resolve as expected.

I believe these two points are caused by the same issue - that pubs would still appear in the results when the user attempted to filter by page. Then, re-selecting the "Pubs" filter would yield and empty set of results.

I'm not 100% sure about the first bullet point in #853 related to Pubs not loading. I haven't witnessed this behavior first-hand, but I do think that this refactor will improve the reliability of the search page in general.

**Test Plan:**

1) Visit the search page
2) Type in a search query, e.g. "Untitled". Take a mental note of the results.
3) Toggle the filter to "Pages". You should observe the results changing, and the pubs from the prior filter should no longer be visible
4) Toggle the filter back to "Pubs". You should see the original results in step 1
5) Type in a query that results in more than one page of results
6) Go to page two of the results. Take a mental note of the results.
7) Reload the page. You should see the original results in step 6

### DIFF
--- a/client/containers/Search/Search.js
+++ b/client/containers/Search/Search.js
@@ -16,6 +16,18 @@ const propTypes = {
 	searchData: PropTypes.object.isRequired,
 };
 
+const getSearchPath = (query, page, mode) => {
+	const queryString = query ? `?q=${query}` : '';
+	const pageString = page ? `&page=${page + 1}` : '';
+	const modeString = mode !== 'pubs' ? `${queryString ? '&' : '?'}mode=${mode}` : '';
+
+	return `/search${queryString}${pageString}${modeString}`;
+};
+
+const updateHistory = (query, page, mode) => {
+	window.history.replaceState({}, '', getSearchPath(query, page, mode));
+};
+
 const Search = (props) => {
 	const { searchData } = props;
 	const { locationData, communityData } = usePageContext();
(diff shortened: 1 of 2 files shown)


client/containers/Search/Search.js
```diff
@@ -16,6 +16,18 @@ const propTypes = {
 	searchData: PropTypes.object.isRequired,
 };
 
+const getSearchPath = (query, page, mode) => {
+	const queryString = query ? `?q=${query}` : '';
```
idreyn, line 20: One alternative to this that I'd like to float is using `new URLSearchParams({...}).toString()`. We'd have to include the polyfill [here](https://github.com/pubpub/pubpub/blob/ddcfb2c2ba09e1dc700fdfb9a1ef44d0d043ab86/server/Html.js#L17) in order to support IE11, but I'd consider it a worthy addition.

client/containers/Search/Search.js
```diff
@@ -16,6 +16,30 @@ const propTypes = {
...
+		params.append('q', query);
+	}
+
+	if (page) {
+		params.append('page', page + 1);
+	}
+
+	if (mode !== 'pubs') {
+		params.append('mode', mode);
+	}
```
3mcd, line 32: Not the prettiest code, but this seemed like the best way to avoid empty query parameters in the case of no query, e.g. `/search?q=`, or worse, stringified `undefined`s leaking into the URL when using the constructor dictionary initializer.

PR #890, opened 2020-07-02
"Related Objects" models and API

### LINKED ISSUE #881: [RO] Define Related Objects data model
- `RelatedObject` model
- any required database migrations
- ideally an adjacent `queries.js` file to CRUD Related Objects

### LINKED ISSUE #882: [RO] Related Objects API
- Implement permissions for RO and CRUD API at `/api/relatedObjects`
- API tests

idreyn:
This PR implements the models and API for "Related Objects" (FKA LP). Recall that we wish to specify a link between a given Pub and either another Pub or a publication that exists outside of PubPub. To do this, we introduce two new models:

### `ForeignPublication`

A `ForeignPublication` is a container for basic information about a publication (e.g. an article) living elsewhere on the internet. You might imagine some potential uses for this model later on — I can, too!

### `PubEdge`

A `PubEdge` specifies the relationship from a Pub (given by its `pubId`) to either another "target Pub" (`targetPubId`) or a ForeignPublication (`foreignPublicationId`). Its `relationType` is one of a fixed set including usual suspects like `comment`, `reply`, `translation`, etc. By setting `pubIsParent`, the Pub creating the edge can choose to act as the "parent" or "child" in the relationship — this is the difference between "is replied to by" and "is a reply to", and the difference between the target Pub appearing in the header or footer of the Pub body. Thus, the nature of the relationship between the two Pubs is specified fully by `(relationType, pubIsParent)`.

Edges created by a Pub are bound to its `outboundEdges` property by Sequelize. Edges targeting a Pub are bound to `inboundEdges`. Pubs won't necessarily want to display their inbound edges, which could come from elsewhere on PubPub — so `PubEdge` has an `approvedByTarget` property that can be configured by the API. This is set to `true` when the edge is created if both the creating and target Pubs are managed by the creating user.

------

I am actively seeking feedback on the names chosen here. Do `PubEdge`, `targetPub`, `ForeignPublication`, etc. make sense? We are going to be stuck with them for a long time.

Eventually we also need some mechanism by which an edge can remember a specific Release that's connected to, probably once we start wiring this into Crossref. One related consequence of this design worth calling out is that we will ultimately have to store `abstract` information on `Release` models for access that's more performant than cracking open Firebase documents. More to come

------

### Test plan:

We have unit tests!

```
npm run test server/pubEdge
``` 

Resolves #881 
Resolves #882


server/utils/queryHelpers/scopeGet.js
```diff
@@ -68,6 +68,17 @@ const getActiveIds = ({ activePub, activeCollection, activeCommunity }) => {
 	};
 };
 
+const getActivePubQuery = ({ id, slug, communityId }) => {
```
idreyn, line 71: @isTravis can you take a peek at this specifically :)

server/pub/model.js
```diff
@@ -123,21 +110,21 @@ export default (sequelize, dataTypes) => {
...
 					Pub.hasMany(PubVersion, {
 						onDelete: 'CASCADE',
 						as: 'pubVersions',
 						foreignKey: 'pubId',
 					});
-					// Pub.hasMany(Review, {
-					// 	onDelete: 'CASCADE',
-					// 	as: 'reviews',
-					// 	foreignKey: 'pubId',
-					// });
```
3mcd, line 140: Nice! Down with commented code!

server/pubEdge/__tests__/api.test.js
```diff
@@ -0,0 +1,210 @@
...
+		.send({
+			pubId: sourcePub.id,
+			relationType: 'review',
+			pubIsParent: true,
+			targetPubId: targetPub.id,
+		})
+		.expect(403);
+});
+
+it('lets a manager can create a PubEdge to another Pub', async () => {
```
3mcd, line 72: Very minor, looks like you can remove the word "can".

server/pubEdge/model.js
```diff
@@ -0,0 +1,35 @@
+export default (sequelize, dataTypes) => {
+	return sequelize.define(
+		'PubEdge',
+		{
+			pubId: { type: dataTypes.UUID, allowNull: false },
+			targetForeignPublication: { type: dataTypes.UUID, allowNull: true },
+			targetPubId: { type: dataTypes.UUID, allowNull: true },
+			relationType: { type: dataTypes.STRING, allowNull: false },
+			rank: { type: dataTypes.TEXT, allowNull: false },
```
3mcd, line 9: It was obviously intentional based on your unit tests, but just out of curiosity, why is `rank` represented w/ a string rather than a numeric type like `INTEGER` ?

server/utils/queryHelpers/scopeGet.js
```diff
@@ -68,6 +68,17 @@ const getActiveIds = ({ activePub, activeCollection, activeCommunity }) => {
...
+	if (!communityId && slug) {
+		throw new Error('Pub slug must be passed to scopeGet with a communityId');
+	}
+	return {
+		...(id && { id: id }),
+		...(slug && { slug: slug }),
+		...(communityId && { communityId: communityId }),
+	};
```
3mcd, line 79: Not strongly opinionated here but this could be cleaned up with a helper fn, e.g.

```js
return removeEmpty({
  id: id,
  slug: slug,
  communityId: communityId,
});
```

You don't win any points in terms of line count but is a bit easier to read. Feel free to ignore.

server/foreignPublication/model.js
```diff
@@ -0,0 +1,11 @@
+export default (sequelize, dataTypes) => {
+	return sequelize.define('ForeignPublication', {
+		title: { type: dataTypes.TEXT, allowNull: false },
+		url: { type: dataTypes.TEXT, allowNull: false },
+		byline: { type: dataTypes.TEXT },
+		doi: { type: dataTypes.TEXT },
+		description: { type: dataTypes.TEXT },
+		avatar: { type: dataTypes.TEXT },
+		publicationDate: { type: dataTypes.DATE },
+	});
```
isTravis, line 10: One thought that could be out of scope for this PR is to consider what [open graph tags](https://ogp.me/) we expect most publications to support that would be useful. Would we want to grab `of:type` to note when something is a video, for example? Perhaps `image:alt`?

server/pubEdge/__tests__/api.test.js
```diff
@@ -0,0 +1,210 @@
...
+		.send({
+			pubId: sourcePub.id,
+			relationType: 'review',
+			pubIsParent: true,
+			targetPubId: targetPub.id,
+		})
+		.expect(403);
+});
+
+it('does not let a manager create a Pub targeting their Pub', async () => {
```
isTravis, line 58: Typo nit: 'does not let a manager create a _PubEdge_ targeting their Pub'

server/utils/queryHelpers/scopeGet.js
```diff
@@ -68,6 +68,17 @@ const getActiveIds = ({ activePub, activeCollection, activeCommunity }) => {
 	};
 };
 
+const getActivePubQuery = ({ id, slug, communityId }) => {
+	if (!communityId && slug) {
+		throw new Error('Pub slug must be passed to scopeGet with a communityId');
+	}
```
isTravis, line 74: I'm not sure this is even needed (nor the older code on the `activePub` query below). `slug` is constrained to be unique across the Pubs table, so there's no need to specify the `communityId` regardless of whether you have `id` or `slug`.

server/utils/queryHelpers/scopeGet.js
```diff
@@ -89,18 +100,19 @@ getScopeElements = async (scopeInputs) => {
 		activeTargetType = 'collection';
 	}
 
-	if (!activeCommunity) {
+	if (!activeCommunity && communityId) {
 		activeCommunity = await Community.findOne({
 			where: { id: communityId },
 		});
 	}
```
isTravis, line 107: Given the note above, and that we don't need `communityId` to find an `activePub`, perhaps we simply remove this section and rely on the `Community.findOne` below with a slightly more general query:

```
if (!activeCommunity) {
	activeCommunity = await Community.findOne({
		where: { id: communityId || activePub.communityId },
	});
}
```

We would also have to move that function out of the `if (activeTargetType === 'pub') {` block to handle the case where `activeTargetType === 'collection'`

server/utils/queryHelpers/scopeGet.js
idreyn: Good idea! Calling it `stripUndefinedValues` because that's the behavior we actually need.

server/foreignPublication/model.js
idreyn: Maybe! But definitely out of scope here.

server/utils/queryHelpers/scopeGet.js
idreyn: Cool — I will remove this check.

server/utils/queryHelpers/scopeGet.js
idreyn: pfft, wait, no it's not.

server/utils/queryHelpers/scopeGet.js
idreyn: Going to run with something similar to this, making it possible to scope the query to a community by leaving this block near the top of the function:

```js
if (!activeCommunity && communityId) {
    activeCommunity = await Community.findOne({
        where: { id: communityId },
    });
}
```
and then adding this one below (which is a more general version of what was living in the `activeTargetType === 'pub'` block:

```
if (!activeCommunity && activeTarget) {
    activeCommunity = await Community.findOne({
        where: { id: activeTarget.communityId },
    });
}
```

So the new behavior is, in psuedocode:

1. Query for `activeCommunity` if there is a `communityId` passed
2. Query for `activePub` (optionally scoped to `activeCommunity`)
3. Query for `activeCollection` (optionally scoped to `activeCommunity`)
4. If there is no `activeCommunity`, try to infer it from `activeTarget.communityId`

This makes me most confident that the various `getScope()` callsites that pass in a community ID where one is not strictly needed will continue to work as expected, but we can gradually move away from that as a best practice.

server/pubEdge/model.js
idreyn: The short answer (which will become longer as I keep typing) is because we need to be able to reorder things with `rank`, and moving the item with `rank = 100` to the beginning of the list means incrementing the `rank` of 99 other items (the first item is now the second, etc). No doubt there is some Postgres-fu to make this happen with a single query (and I would like to learn it!) but the limiting case still means updating hundreds or thousands of rows.

So historically we've used a scheme where `rank` (FKA `order`) is a `DOUBLE`, and when item C is inserted between items B and A, we set `C.order = (A.order + B.order) / 2`. But we run out of precision pretty fast. So these days, we use strings to create ordered collections in Postgres because they have infinite precision — given two strings `a < c` and the lexical ordering relation `<`, you can always slot a `b` in between them such that `a < b < c` just by choosing `b = a + (some characters)`. We have a library that uses a more intelligent algorithm to pick `b` for us, and it tends to work great!

idreyn:
@isTravis `ExternalPublication` is definitely better, we should rename it.

PR #893, opened 2020-07-07
fix: Create checkpoints on branch merge

idreyn:
A failure mode I encountered recently that was causing an uptick in timeouts:

- Someone creates a very long Pub (30k firebase changes)
- They can load the draft just fine because of incremental checkpoints
- They create a Release, which bunches all of the changes into a merge, but doesn't create a checkpoint
- Loading that Release is slow because it has to reassemble the doc from changes 0 to 30k, causing timeouts/OOM

To solve this I have added functionality to create a checkpoint when two branches are merged, e.g. when a Release is created. It will only be a bit of extra work to arrive at one of these more stable solutions:

- drop Release branches entirely and read Release content from `(draftBranchId, key)` pairs
- just store Release docs in Postgres and load them from there

...but this is a good interim solution.

_Test plan:_

I observed the three ways a checkpoint can be created:

- Created a new Pub and started typing
- Refreshed the page at `key=57` and noted that a checkpoint was created there.
- Typed past `key=100`, `key=200`, etc. and noticed that checkpoints were created there.
- Created a release of the Pub and observed that on the release/public branch, a checkpoint was created at `key=0` (representing the merge).

### DIFF
--- a/client/components/Editor/utils/branches.js
+++ b/client/components/Editor/utils/branches.js
@@ -26,50 +26,43 @@ export const createBranch = (baseFirebaseRef, newFirebaseRef, versionNumber) => 
 	});
 };
 
-export const mergeBranch = (sourceFirebaseRef, destinationFirebaseRef, prosemirrorSchema) => {
-	/* TODO-BRANCH At the moment, this merge simply appends new changes in a merge */
-	/* It does not properly handle 'commonAncestor' or any similar */
-	/* concept which would be needed for multi-direction merging */
-	/* or multi-branch merge trees */
-	return destinationFirebaseRef
+export const mergeBranch = async (sourceFirebaseRef, destinationFirebaseRef, prosemirrorSchema) => {
+	const mergesSnapshot = await destinationFirebaseRef
 		.child('merges')
 		.orderByKey()
 		.startAt(String(0))
-		.once('value')
-		.then((mergesSnapshot) => {
-			const mergesSnapshotVal = mergesSnapshot.val() || {};
-			const numKeyables = Object.values(mergesSnapshotVal).reduce((prev, curr) => {
-				return prev + curr.length;
-			}, 0);
-			const nextMergeKey = Object.values(mergesSnapshotVal).length;
-			const getSourceChanges = sourceFirebaseRef
-				.child('changes')
-				.orderByKey()
-				.startAt(String(numKeyables))
-				.once('value');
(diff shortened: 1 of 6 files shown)


server/utils/firebaseAdmin.js
```diff
@@ -58,19 +59,23 @@ const maybeAddKeyTimestampPair = (key, timestamp) => {
...
 	const branchRef = getBranchRef(pubId, branchId);
 
 	const [
-		{ doc, key: currentKey, timestamp: currentTimestamp, checkpointMap },
+		{ doc, docIsFromCheckpoint, key: currentKey, timestamp: currentTimestamp, checkpointMap },
 		{ timestamp: firstTimestamp, key: firstKey },
 		{ timestamp: latestTimestamp, key: latestKey },
 	] = await Promise.all([
-		getFirebaseDoc(branchRef, editorSchema, historyKey, updateOutdatedCheckpoint),
+		getFirebaseDoc(branchRef, editorSchema, historyKey, createMissingCheckpoints),
```
3mcd, line 70: Are we using this fourth argument any more? Looks like it was removed.

client/components/Editor/utils/branches.js
```diff
@@ -59,8 +60,16 @@ export const mergeBranch = (sourceFirebaseRef, destinationFirebaseRef) => {
...
-			});
+
+			await Promise.all([setLastMergeKey, appendMerge]);
+
+			const { doc } = await getFirebaseDoc(
+				destinationFirebaseRef,
+				prosemirrorSchema,
+				nextMergeKey,
+			);
+			await storeCheckpoint(destinationFirebaseRef, doc, nextMergeKey);
```
isTravis, line 71: I might be missing a step in the logic, but won't building this checkpoint still fail and hit OOM/hanging issues for pubs that have 30k+ steps (just now at Merge creation rather than pub load)? That's still an improvement, as it doesn't let you publish something only to have it inaccessible. 

We could `getFirebaseDoc` based on the sourceFirebaseRef which would grab any relevant incremental checkpoint to handle the large pub case. Apologies if this is already doing that and I missed it.

idreyn: 🤦 you're right, I meant to pass in `sourceFirebaseRef` here.

server/utils/firebaseAdmin.js
idreyn: Right you are, good catch!

PR #894, opened 2020-07-07
update: Add one-line explanations to the top of most dashboard pages

idreyn:
While working on the dashboard tab for Pub edges, I realized that the `details` prop of the `DashboardFrame` is criminally underused, and could be employed to give brief explanations about what each Dashboard tab does. I've taken a stab at the copy here, feel free to poke holes in it or just tell me that this is a silly idea.

Here's what it looks like when text is rendered into `details`:
image

### DIFF
--- a/client/containers/DashboardReviews/DashboardReviews.js
+++ b/client/containers/DashboardReviews/DashboardReviews.js
@@ -28,7 +28,11 @@ const DashboardReviews = (props) => {
 		return pub.reviews.length;
 	});
 	return (
-		<DashboardFrame className="dashboard-reviews-container" title="Reviews">
+		<DashboardFrame
+			className="dashboard-reviews-container"
+			title="Reviews"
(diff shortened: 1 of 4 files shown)


client/containers/DashboardReviews/DashboardReviews.js
```diff
@@ -28,7 +28,11 @@ const DashboardReviews = (props) => {
 		return pub.reviews.length;
 	});
 	return (
-		<DashboardFrame className="dashboard-reviews-container" title="Reviews">
+		<DashboardFrame
+			className="dashboard-reviews-container"
+			title="Reviews"
+			details="Reviews allow members of your Community to request feedback on their Pubs and release them to the world."
+		>
```
3mcd, line 35: Each of the other descriptions is written in active tense while this one is more of a description. Here's a really bad stab at it: "Manage requests for feedback created by members of your Community"?

PR #900, opened 2020-07-10
Integrate the Pub Connections edge proposal API and edge editor

idreyn:
This PR lets you add Pub connections by PubPub URL or DOI, and edit edges that resolve with an `ExternalPublication`:
Screen Recording 2020-07-10 at 9 54 51 AM 2020-07-10 10_08_24

_Test plan:_
- Enter a PubPub url like https://cursor.pubpub.org/pub/uv2c6nd4/merge/2/1 and ensure you can create an edge
- Enter a DOI like `10.1016/j.trd.2010.12.004` and ensure you can create an edge, as well as edit its metadata.
- Enter any old url and ensure that you can create an edge with it, even if we don't resolve a useful proposal.


client/components/PreviewImage/PreviewImage.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
idreyn, line 1: This was mostly moved out of `PubPreviewImage` and given a slightly more generalized API.

client/components/PubEdge/PubEdgeEditor.js
```diff
@@ -23,19 +24,56 @@ const PubEdgeEditor = (props) => {
...
 	const handlePublicationDateChange = (evt) => {
 		const nextDateString = evt.target.value;
-		setPublicationDateString(nextDateString);
-		onUpdateExternalPublication({ publicationDate: new Date(nextDateString).toString() });
+		onUpdateExternalPublication({
+			publicationDate: nextDateString ? new Date(nextDateString).toString() : null,
+		});
+	};
+
+	const renderPublicationDate = () => {
```
idreyn, line 40: Sigh...I can't believe `<input type="date">` is so hard to work with.

server/pubEdgeProposal/api.js
```diff
@@ -28,6 +27,6 @@ app.get(
 			edge = await createPubEdgeProposalFromCrossrefDoi(object);
 		}
 
-		return edge ? res.status(200).json(edge) : res.status(404);
+		return res.status(200).json(edge);
```
idreyn, line 30: I forgot that when `apiFetch` gets a non-2XX status code, it throws an error — so we should really just use 200 here and think of `null` as the successfully resolved edge proposal :p

server/pubEdgeProposal/queries.js
```diff
@@ -27,7 +27,7 @@ export const createPubEdgeProposalFromCrossrefDoi = async (doi) => {
 			description: abstract,
 			doi: DOI,
 			publicationDate: publicationDate,
-			title: title,
+			title: Array.isArray(title) ? title.join(': ') : title,
```
idreyn, line 30: Not so sure what to do with the multiple title components here

server/utils/queryHelpers/pubOptions.js
```diff
@@ -99,11 +99,13 @@ export default ({ isAuth, isPreview, getCollections, getMembers, getCommunity, g
 			{
 				model: PubEdge,
 				as: 'outboundEdges',
+				separate: true,
```
idreyn, line 102: Huge performance speedup by not computing this as a single query

client/components/Byline/Byline.js
```diff
@@ -48,7 +50,7 @@ const Byline = (props) => {
 							);
 						}
 
-						const user = author.user;
+						const { user } = ensureUserForAttribution(author);
```
3mcd, line 53: I'll remember this for next time!

client/components/PubEdge/PubEdgeEditor.js
3mcd: It's so bad. If you put in an invalid date, its value becomes `''` instead of the incorrectly formatted date, which makes it impossible to provide useful feedback to the user. Same with input type="number".

server/pubEdgeProposal/api.js
3mcd: Sounds good.

PR #901, opened 2020-07-10
feature: add scraping functionality to pubEdgeProposal api

### LINKED ISSUE #883: [RO] Metadata extractor endpoint
Given a URL or a doi.org URL, or Pub ID, returns (best guess) at:

- Title
- Authors
- Abstract (as Prosemirror JSON)
- Publication date
- Image?

Also resolve PubPub URLs into Pub objects

3mcd:
This PR adds support to the pubEdgeProposal endpoint for arbitrary urls.

This was implemented with a utility called `runQueries` (needs a better name) in a new file `utils/scrape.js`. `runQueries` takes a `Request` (fetch spec) object and a query configuration, and returns a result set. A query configuration looks like:

```ts
const queries = {
  title: ["article h1"]
};
```

Each query is an array of DOM selectors. You can also specify more advanced selectors as objects containing a selector and process function used to transform a jQuery collection into a serializable value, e.g.

```ts
const queries = {
  // ...
  thumbnail: [
    'meta[property="og:image:secure_url"]',
    'meta[property="og:image"]',
    { selector: 'article img, .abstract img', process: ($el) => $el.attr('src') },
  ]
};
```

Selectors are run in the same order as their position in the array. The process function will execute on the first element encountered in the document, unless the `multiple` flag is provided.

```ts
const queries = {
  // ...
  authors: [
    { selector: ".author", multiple: true, process: ($el) => $el.toArray().map((el) => $(el).text()) }
  ]
};
```

```ts
const response = await request("https://www.nature.com/articles/d41586-020-02058-1");
const { authors, title, thumbnail } = await runQueries(queries, response);

authors: // -> string[]
title; // -> null | string
thumbnail; // -> null | string
``` 

**Screenshots:**

Screen Shot 2020-07-10 at 2 34 08 PM

captured (1)

**Test Plan:**

```
# cambridge.org article
curl http://localhost:9876/api/pubEdgeProposal?object=https%3A%2F%2Fwww.cambridge.org%2Fcore%2Fjournals%2Fmrs-bulletin%2Farticle%2Fmeasurement-of-local-strain%2FBC00EA61D22C88F3DF6B0EDC15DD55ED

# nature article
http://localhost:9876/api/pubEdgeProposal?object=https%3A%2F%2Fwww.nature.com%2Farticles%2Fd41586-020-02058-1

# magazine article
curl http://localhost:9876/api/pubEdgeProposal?object=https%3A%2F%2Fcrackmagazine.net%2Farticle%2Flists%2Fdeforrest-brown-reading-list%2F
```

Closes #883


server/pubEdgeProposal/queries.js
```diff
@@ -33,6 +47,31 @@ export const createPubEdgeProposalFromCrossrefDoi = async (doi) => {
...
+const mergeProposalWithCrossrefProposal = async (edge, doi) => {
+	const proposalFromDoi = await createPubEdgeProposalFromCrossrefDoi(doi);
+
+	return {
+		...edge,
+		externalPublication: mergewith(
+			{},
+			edge.externalPublication,
+			proposalFromDoi.externalPublication,
+			(a, b) => b ?? a,
```
idreyn, line 59: I...did not know we had this available to us!

utils/scrape.js
```diff
@@ -0,0 +1,102 @@
+import cheerio from 'cheerio';
```
idreyn, line 1: This file should live in `server/utils` since `cheerio` is not a dependency we want to have on the client.

PR #902, opened 2020-07-10
Add "Pub Connections" views to Pub body

idreyn:
This PR (finally) adds Pub connections to the Pub body, including a new query for sibling Pub edges. There is also a new `canViewPubEdge` function that gets called to avoid leaking information about unreleased Pubs. It doesn't quite fully determine whether a user can see a given Pub if it's outside of the current Community scope, but it's good enough to handle intra-community linking without being too heavyweight.

This needs extensive testing and I'm pretty certain I missed something somewhere, so feel free to try to break it!

_Some screenshots:_

Pub connections at the top of a Pub:
Screen Shot 2020-07-10 at 4 50 38 PM

Pub connections at the bottom of a Pub (you'll have to take my word for it):
Screen Shot 2020-07-10 at 4 50 51 PM

Same as above, but showing sibling objects.
Screen Shot 2020-07-10 at 4 51 17 PM

_Test plan:_

- Make `B` a reply to `A` and verify that `A` shows up at the top of `B` and `B` shows up at the bottom of `A`
- Make `C` a reply to `A` and verify that `B` and `C` reference each other as siblings
- Create an unreleased Pub `D` linked to `A` and verify that it appears on `A` iff you are logged in.


client/components/PubEdgeListing/PubEdgeListingCard.js
```diff
@@ -33,22 +34,33 @@ const defaultProps = {
...
+		// If the `pub` in the edge relationship is the parent, then the `targetPub` is the child.
+		// We want this edge to display the child -- in other words, we want to view from the
+		// `targetPub` towards the `pub` only if the `pub` is the child.
+		return !pubIsParent;
+	}
+	// If this edge is inbound, that means another Pub (relative to our vantage point) created it,
+	// which is to say that we're viewing it from the target.
+	return isInboundEdge;
+};
+
```
3mcd, line 49: Love the comments. Makes this easier to wrap my head around.

3mcd:
Looks like my merge left you with a conflict here. Otherwise LGTM!

PR #905, opened 2020-07-14
fix: improve pub edge styling; use anchor instead of clickable div

3mcd:
This PR addresses the following feedback:

1. `.pub-edge-component .title-container` : change font-size to 18px and line-height to 24px (using px instead of em for specificity)
2. `.pub-edge-listing-card-component>.relation .drop-return`: change bottom to 1px (for better alignment with adjacent text layer's baseline)
3. Use multidisplay font for metadata.
4. Use 12px font size for metadata.
5. Use lighter border when PubEdge card is inactive/not hovered. This was accomplished using `mouseenter`/`mouseleave` events.
6. Drop return icon now has accent color before and after hovering.

### DIFF
--- a/client/components/PubEdge/PubEdge.js
+++ b/client/components/PubEdge/PubEdge.js
@@ -109,41 +109,52 @@ const PubEdge = (props) => {
 		onKeyDown: handleClick,
 		role: 'link',
 		tabIndex: '0',
+		alt: title,
 	};
 
 	const maybeLink = (element, restProps = {}) => {
 		if (actsLikeLink) {
-			return element;
+			return (
+				<span {...restProps} className={classNames(restProps.className, 'link')}>
+					{element}
+				</span>
+			);
 		}
-		return (
-			<a href={url} {...restProps}>
-				{element}
-			</a>
-		);
+
+		return <a href={url}>{element}</a>;
 	};
 
-	return (
+	const maybeWrapWithLink = (element, restProps = {}) => {
+		if (actsLikeLink) {
+			return (
+				<a href={url} {...restProps}>
+					{element}
+				</a>
+			);
+		}
+
+		return <div {...restProps}>{element}</div>;
+	};
+
+	return maybeWrapWithLink(
 		<PubEdgeLayout
-			outerElementProps={linkLikeProps || {}}
-			className={classNames('pub-edge-component', actsLikeLink && 'acts-like-link')}
 			topLeftElement={avatar && maybeLink(<img src={avatar} alt="" />, { tabIndex: '-1' })}
-			titleElement={maybeLink(title)}
+			titleElement={maybeLink(title, linkLikeProps)}
 			bylineElement={<Byline contributors={contributors} />}
 			metadataElements={[
 				description && (
 					<Button
-						as="a"
+						as={actsLikeLink ? 'span' : 'a'}
 						onClick={handleToggleDescriptionClick}
 						onKeyDown={handleToggleDescriptionClick}
 						tabIndex="0"
+						className="link"
 					>
 						{open ? 'Hide Description' : 'Show Description'}
 					</Button>
 				),
 				<>Published on {publishedAt}</>,
-				<a href={url} alt={title} tabIndex="0">
-					{url}
-				</a>,
+				maybeLink(url, linkLikeProps),
 			]}
 			detailsElement={
 				<details open={open}>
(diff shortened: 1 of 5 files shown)


client/components/PubEdge/PubEdge.js
```diff
@@ -109,41 +109,52 @@ const PubEdge = (props) => {
...
 	};
 
 	const maybeLink = (element, restProps = {}) => {
 		if (actsLikeLink) {
-			return element;
+			return (
+				<span {...restProps} className={classNames(restProps.className, 'link')}>
+					{element}
+				</span>
+			);
```
3mcd, line 121: `maybeLink` now returns a "false" link, which is just a span with the class `link` and whatever props the function was passed.

client/components/PubEdge/PubEdge.js
```diff
@@ -109,41 +109,52 @@ const PubEdge = (props) => {
...
+	const maybeWrapWithLink = (element, restProps = {}) => {
+		if (actsLikeLink) {
+			return (
+				<a href={url} {...restProps}>
+					{element}
+				</a>
+			);
+		}
+
+		return <div {...restProps}>{element}</div>;
```
3mcd, line 136: This function needs a better name. It just wraps the element with an anchor tag if `actsLinkLike` is true.

client/components/PubEdge/PubEdge.js
```diff
@@ -109,41 +109,52 @@ const PubEdge = (props) => {
...
 			bylineElement={<Byline contributors={contributors} />}
 			metadataElements={[
 				description && (
 					<Button
-						as="a"
+						as={actsLikeLink ? 'span' : 'a'}
 						onClick={handleToggleDescriptionClick}
 						onKeyDown={handleToggleDescriptionClick}
 						tabIndex="0"
+						className="link"
```
3mcd, line 151: Use a false link if PubEdge parent acts as a link.

client/components/PubEdgeListing/PubEdgeListingCard.js
```diff
@@ -61,6 +61,10 @@ const PubEdgeListingCard = (props) => {
 	} = props;
 	const { communityData } = usePageContext();
 	const viewingFromTarget = getIsViewingFromTarget(pubEdge, viewingFromSibling, isInboundEdge);
+	const [hover, setHover] = useState(false);
+	const handleMouseEnter = useCallback(() => setHover(true), []);
+	const handleMouseLeave = useCallback(() => setHover(false), []);
+	const style = hover ? { borderColor: accentColor || communityData.accentColorDark } : {};
```
3mcd, line 67: I couldn't get my silly idea of `!important` working since there's no way to switch off a rule in a different scope in CSS, just override it with a new rule.

client/components/PubEdgeListing/pubEdgeListingCard.scss
```diff
@@ -5,18 +5,25 @@
 	border-width: 1px;
 	margin-bottom: 24px;
 	position: relative;
-	transition: filter 100ms linear;
+	transition: all 100ms linear;
```
3mcd, line 8: Changed to `all` to transition border color from gray to community accent color when it is set.

client/components/PubEdgeListing/pubEdgeListingCard.scss
```diff
@@ -5,18 +5,25 @@
...
 
 	&.in-pub-body {
-		filter: grayscale(100%);
+		> * {
+			filter: grayscale(100%);
+		}
```
3mcd, line 13: Instead of making the whole card grayscale, make all children _except_ the `.relation` element (with the return arrow icon) grayscale.

client/components/PubEdge/PubEdge.js
```diff
@@ -109,41 +109,52 @@ const PubEdge = (props) => {
 		onKeyDown: handleClick,
 		role: 'link',
 		tabIndex: '0',
+		alt: title,
```
idreyn, line 112: Most non-`img` tags do not accept an `alt`, and I don't think one is necessarily needed here either, since I believe VoiceOver does a pretty good job of extracting spoken text from these cards.

client/components/PubEdge/PubEdge.js
idreyn: Seems fine to me!

client/components/PubEdge/PubEdge.js
```diff
@@ -109,41 +109,52 @@ const PubEdge = (props) => {
...
 	};
 
 	const maybeLink = (element, restProps = {}) => {
 		if (actsLikeLink) {
-			return element;
+			return (
+				<span {...restProps} className={classNames(restProps.className, 'link')}>
```
idreyn, line 118: IMO it's unnecessary to pass `restProps.className` here, since I don't believe this functionality is ever used.

PR #924, opened 2020-07-17
fix: pub connection styling fixes and placeholder thumbnails

3mcd:
Based on feedback from @djagdish, this PR:
* Changes the un-hovered border color to `#dddddd`
* `.pub-edge-listing-card-component` -- increases margin-bottom to 30px to add a bit more breathing space between cards when viewed in list format
* `.pub-edge-listing-component` -- adds padding-top and padding-bottom of 20px to add a bit more space between Pub body and the related pubs.
* Adds placeholder thumbnails when a connected pub or external publication doesn't resolve with an avatar.

**Without hover state:**

Screen Shot 2020-07-17 at 12 44 45 PM

**With hover state:**

Screen Shot 2020-07-17 at 12 44 27 PM

### DIFF
--- a/client/components/PubEdge/PubEdge.js
+++ b/client/components/PubEdge/PubEdge.js
@@ -23,7 +23,7 @@ export const propTypes = {
 };
 
 const defaultProps = {
-	accentColor: '#dddddd',
+	accentColor: '#ddd',
 	actsLikeLink: false,
 	viewingFromTarget: false,
 };
--- a/client/components/PubEdge/PubEdge.js
+++ b/client/components/PubEdge/PubEdge.js
@@ -142,7 +142,7 @@ const PubEdge = (props) => {
 		<PubEdgeLayout
 			topLeftElement={maybeLink(
 				avatar ? (
-					<img src={avatar} alt="" />
+					<img src={avatar} alt={title} />
 				) : (
 					<PubEdgePlaceholderThumbnail
 						color={accentColor}
--- a/client/components/PubEdge/PubEdge.js
+++ b/client/components/PubEdge/PubEdge.js
@@ -48,7 +48,6 @@ const getValuesFromPubEdge = (pubEdge, communityData, viewingFromTarget) => {
 			avatar: avatar,
 			contributors: attributions,
 			description: description,
-			isExternal: false,
 			publicationDate: getPubPublishedDate(displayedPub),
 			title: title,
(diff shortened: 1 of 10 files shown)


client/components/PubEdge/PubEdge.js
```diff
@@ -11,16 +11,19 @@ import { getPubPublishedDate } from 'utils/pub/pubDates';
...
 
 export const propTypes = {
+	accentColor: PropTypes.string,
 	actsLikeLink: PropTypes.bool,
 	pubEdge: pubEdgeType.isRequired,
 	viewingFromTarget: PropTypes.bool,
 };
 
 const defaultProps = {
+	accentColor: '#dddddd',
```
idreyn, line 26: Or just `#ddd`

client/components/PubEdgeListing/PubEdgeListingCard.js
```diff
@@ -52,7 +52,7 @@ const getIsViewingFromTarget = (pubEdge, viewingFromSibling, isInboundEdge) => {
 
 const PubEdgeListingCard = (props) => {
 	const {
-		accentColor,
+		accentColor = communityData.accentColorDark,
```
idreyn, line 55: ⚠️ `communityData` used before it's defined.

client/components/PubEdgeListing/pubEdgeListing.scss
```diff
@@ -1,6 +1,7 @@
 .pub-edge-listing-component {
 	display: flex;
 	flex-direction: column;
+	padding: 20px 0;
```
idreyn, line 4: If this is in place to provide the correct amount of distance from other Pub body elements, I think I'd prefer to address them individually by their class names given in `PubDocument.js` and referenced in `pubDocument.scss`. I don't think this component ought to BYOP, so to speak. At any rate I'm concerned tha the combined effect of this padding plus the margin there is maybe too much.

client/containers/Pub/PubDocument/pubDocument.scss
```diff
@@ -10,6 +10,11 @@
 		margin-bottom: 30px;
 	}
 
+	.bottom-pub-edges {
+		padding-top: 30px;
+		padding-bottom: 30px;
+	}
```
3mcd, line 16: I changed the padding to affect the footer pub edge listing only.

PR #925, opened 2020-07-17
Export bullet font size fix

gabestein:
Very quick fix to make bullets in exports the same size as everything else. Previously, they were much larger. I also commented out some CSS that I had added but realized didn't actually take effect yet because I didn't capitalized Mirror in ProseMirror. I intend to go back and make these changes eventually

Screen Shot 2020-07-17 at 16 03 08

Screen Shot 2020-07-17 at 16 05 00

_Test Plan_
1. Export a PDF with a paragraph, an unordered list, and an ordered list (ie: https://demo.pubpub.org/pub/md4wb7lk/draft). Make sure it's all the same size.

### DIFF
--- a/workers/tasks/export/styles/printDocument.scss
+++ b/workers/tasks/export/styles/printDocument.scss
@@ -29,7 +29,7 @@ https://gitlab.pagedmedia.org/tools/pagedjs-cli/issues/11 */
 
 .pub-body-component {
     // override bullet sizing for PDFs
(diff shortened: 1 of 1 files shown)


gabestein:
Thanks! Yeah, we probably need to go through and re-build the CSS now that we're automatically generating it from other pub styles. We can probably make it way simpler. I'll add a card for it.

PR #932, opened 2020-07-20
Some Release fixes

idreyn:
This PR includes a bundle of fixes related to Releases that have been piling up:

- There is a UI bug that appears to let you create a Release from an empty Pub, which is something we need to continue to disallow for the time being due to technical debt in our Firebase infra (cc @metasj)
- Last week we had a conversation the possibility of Releases created from bulk import clogging up the worker queue. I dug into it and it turns out this is behavior we _should_ be seeing, but aren't — the dashboard-adjacent migration from branch-merges to Releases didn't carry over the desired behavior of kicking off exports at publish time. So not only do we need to restore that, we also need a flag to selectively disable the newly-restored behavior during bulk import!
- Sort of central to all of this, the Release API doesn't have any tests. Releasing is a complicated enough process to justify a much more comprehensive test suite than I have here, but this is a start — and I sort of hope the process will be simpler by the time we get to writing any more.

Anyway, this PR addresses all of that. I'll try to add some elucidation inline, because this is a lot.

_Test plan:_
- Make sure you can actually create a release through the UI on your devserver
- Run `npm run test server/release`

### DIFF
--- a/client/components/PubReleaseDialog/PubReleaseDialog.js
+++ b/client/components/PubReleaseDialog/PubReleaseDialog.js
@@ -68,7 +68,7 @@ const PubReleaseDialog = (props) => {
 			draftKey: historyData.latestKey,
 			makeDraftDiscussionsPublic: makeDraftDiscussionsPublic,
 		})
-			.then(({ release }) => {
+			.then((release) => {
 				setReleleaseError(null);
 				setCreatedRelease(release);
 				setIsCreatingRelease(false);
--- a/client/containers/DashboardEdges/NewEdgeInput.js
+++ b/client/containers/DashboardEdges/NewEdgeInput.js
@@ -42,6 +42,7 @@ const indeterminateMenuItem = (
 		isSkeleton={true}
 		showImage={true}
 		disabled={true}
+		onClick={() => {}}
 	/>
 );
 
--- a/server/release/api.js
+++ b/server/release/api.js
@@ -30,7 +30,6 @@ app.post(
 	wrap(async (req, res) => {
 		const {
 			communityId,
-			createExports,
 			draftKey,
 			makeDraftDiscussionsPublic,
 			noteContent,
@@ -55,7 +54,6 @@ app.post(
 			noteText: noteText,
 			noteContent: noteContent,
 			makeDraftDiscussionsPublic: makeDraftDiscussionsPublic,
-			createExports: createExports,
 		});
 
 		return res.status(201).json(release);
--- a/server/release/api.js
+++ b/server/release/api.js
@@ -8,20 +8,20 @@ const getRequestValues = (req) => {
 	const user = req.user || {};
 	const {
 		communityId,
-		pubId,
 		draftKey,
-		noteText,
-		noteContent,
 		makeDraftDiscussionsPublic,
+		noteContent,
+		noteText,
+		pubId,
 	} = req.body;
 	return {
-		userId: user.id,
 		communityId: communityId,
-		pubId: pubId,
 		draftKey: draftKey,
-		noteText: noteText,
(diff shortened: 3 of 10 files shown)


client/components/PubReleaseDialog/PubReleaseDialog.js
```diff
@@ -68,7 +68,7 @@ const PubReleaseDialog = (props) => {
 			draftKey: historyData.latestKey,
 			makeDraftDiscussionsPublic: makeDraftDiscussionsPublic,
 		})
-			.then(({ release }) => {
+			.then((release) => {
```
idreyn, line 71: There is a corresponding change in the Releases `api.js` file to return the bare Release object.

client/containers/DashboardEdges/NewEdgeInput.js
```diff
@@ -42,6 +42,7 @@ const indeterminateMenuItem = (
 		isSkeleton={true}
 		showImage={true}
 		disabled={true}
+		onClick={() => {}}
```
idreyn, line 45: This is just to fix a PropType error that was preventing any other logging from appearing in the Jest runner /shrug

server/release/api.js
```diff
@@ -8,34 +8,34 @@ const getRequestValues = (req) => {
...
 	} = req.body;
 	return {
-		userId: user.id,
 		communityId: communityId,
-		pubId: pubId,
 		draftKey: draftKey,
-		noteText: noteText,
-		noteContent: noteContent,
 		makeDraftDiscussionsPublic: makeDraftDiscussionsPublic,
+		noteContent: noteContent,
```
idreyn, line 21: Alphabetization

server/release/api.js
```diff
@@ -8,34 +8,34 @@ const getRequestValues = (req) => {
...
 	wrap(async (req, res) => {
 		const {
-			userId,
 			communityId,
-			pubId,
 			draftKey,
-			noteText,
-			noteContent,
 			makeDraftDiscussionsPublic,
+			noteContent,
```
idreyn, line 35: Alphabetization

server/release/api.js
```diff
@@ -56,6 +56,6 @@ app.post(
 			makeDraftDiscussionsPublic: makeDraftDiscussionsPublic,
 		});
 
-		return res.status(201).json({ release: release });
+		return res.status(201).json(release);
```
idreyn, line 59: The only meaningful change in this file

stubstub/stub.js
```diff
@@ -4,28 +4,45 @@ import sinon from 'sinon';
 import { getEmptyDoc } from 'components/Editor';
 import * as firebaseAdmin from '../server/utils/firebaseAdmin';
 
-export const stubModule = (module, functionNames) => {
-	if (!Array.isArray(functionNames)) {
-		// eslint-disable-next-line no-param-reassign
-		functionNames = [functionNames];
+const stubManyFunctions = (module, stubEntries) =>
```
idreyn, line 7: None of this is central to the purpose of the PR — just improving the ergonomics of `stubOut` so you can pass it a `name: function` mapping as well as a single function name or list of function names, and a module path rather than a `required` module.

workers/tasks/import/bulk/README.md
```diff
@@ -63,6 +63,8 @@ Any Communities, Collections, and Pubs created will be immediately visible onlin
 npm run tools bulkimport -- --actor=pubpub-user-slug --receipt /path/to/receipt --publish
 ```
 
+You can choose whether to create Pub exports (which you may not want during testing) using the `--create-exports` flag, which defaults to `true`.
```
idreyn, line 66: From here on down this is bulk import code

PR #934, opened 2020-07-21
feat: deposit preview for journals, conferences, books

### LINKED ISSUE #908: [PC DOI Deposits] Deposit Preview
- Create new endpoint that returns a preview of the deposit XML
- Add intermediate step when Assign DOI button is pressed.
- Show a human-readable preview of the proposed deposit.
- Show a disclosure/details element that, when clicked, displays XML in a monospaced pre element or readonly text field.

3mcd:
This PR adds a new component `AssignDoi` which handles previewing and submission of DOI deposits to Crossref. Existing deposit submission logic was removed from the `<Doi>` component used in Pub settings and replaced with `<AssignDoi>`. The deposit preview has both a human-readable view of the deposit and the raw XML that will be sent to Crossref. Previews are supported for journal articles, conferences, and books.

### Journal Article
Screen Shot 2020-07-21 at 11 57 43 AM

### Book Chapter
Screen Shot 2020-07-21 at 11 56 50 AM

### Conference Paper
Screen Shot 2020-07-21 at 11 55 51 AM

Resolves #908

### DIFF
--- a/client/components/AssignDoi/assignDoiPreview.scss
+++ b/client/components/AssignDoi/assignDoiPreview.scss
@@ -1,0 +1,44 @@
+.assign-doi-preview-component {
+	margin: 12px 0;
+	padding: 12px;
+	border: 1px dashed #ccc;
+	border-radius: 3px;
+	font-size: 0.9em;
+
+	dl {
+		display: inline-grid;
+		grid-template-columns: auto auto;
+		margin: 0;
+
+		&:not(:last-child) {
+			margin: 0 0 18px 0;
+		}
+
+		> dt {
+			padding: 4px;
+			word-break: break-word;
+			color: #666;
+		}
+
+		> dd {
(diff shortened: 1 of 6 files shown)


client/components/AssignDoi/assignDoiPreview.scss
```diff
@@ -0,0 +1,44 @@
...
+	dl {
+		display: inline-grid;
+		grid-template-columns: auto auto;
+		margin: 0;
+
+		&:not(:last-child) {
+			margin: 0 0 18px 0;
+		}
+
+		> dt {
```
idreyn, line 17: What do you think about a `text-align: right` on the `dt` elements? Gives the whole thing a bit more visual structure IMO.

3mcd: I like it!

PR #940, opened 2020-07-23
dev: Implement a devshell

idreyn:
I wanted a Node REPL that executes in a PubPub environment, so I made one:

devshell

You can access it with `npm run tools devshell`. Give it a shot!


3mcd:
This is awesome, and insanely useful!

PR #941, opened 2020-07-23
fix: Fix the updateCommunityData worker task

idreyn:
The `updateCommunityData` worker task references an obsolete Algolia API function — I've updated it to use the new equivalent here.

_Test plan:_
I ran the `updateCommunityData` task locally on a community and verified that it completed without error.

### DIFF
--- a/workers/tasks/search.js
+++ b/workers/tasks/search.js
@@ -53,31 +53,18 @@ export const setPubSearchData = (pubId) => {
 };
 
 export const updateCommunityData = (communityId) => {
-	const getObjectIds = (index) => {
-		return new Promise((resolve, reject) => {
-			const browser = index.browseAll('', {
-				filters: `communityId:${communityId}`,
-				attributesToRetrieve: ['objectId'],
-			});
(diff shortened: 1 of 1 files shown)


workers/tasks/search.js
```diff
@@ -53,31 +53,18 @@ export const setPubSearchData = (pubId) => {
...
-				reject(err);
-			});
+	const getObjectIds = async (index) => {
+		let hits = [];
+		await index.browseObjects({
+			filters: `communityId:${communityId}`,
+			attributesToRetrieve: ['objectId'],
+			batch: (batch) => {
+				hits = hits.concat(batch);
+			},
```
3mcd, line 63: Algolia uses an interesting mix of streaming and Promise APIs here.

idreyn: If I had to guess it's that they don't want to assume you want the entire query result in memory so they make you store entries as a side effect. But I agree, very odd.

isTravis:
🎉 Thank you for this!

Issue #910, opened 2020-07-15
[PC DOI Deposits] Editable DOI Field


3mcd:
- DOI should be editable in both Pubs and Collections
- DOI field is immutable once a CrossrefDeposit exists for Pub or Collection
- Display warning message on Deposit Preview to indicate to user that DOI will no longer be editable after deposit is created.

PR #942, opened 2020-07-24
feature: editable doi field

3mcd:
This PR adds an editable DOI field to the Pub settings page. The DOI is editable when users have the correct permissions and a Crossref deposit has not been submit. Once a Crossref deposit record is created, the DOI field will become uneditable.

#### Permissions

* Create a new Pub.
* Log in with a user without manage permissions.
* Go to the settings page.
* The DOI field should not be editable.

#### Update DOI

* Log in with a user with manage permissions.
* Go to the setting spage.
* The DOI field should be editable.
* Enter an invalid DOI, e.g. `10.116299608f92.834c6596`
* An error message should appear in red and the DOI should not be updateable.
* Enter a valid DOI, e.g. `10.1162/99608f92.834c6596`
* Click "Update DOI"
* The error message should disappear and be replaced by the "DOI updated successfully" message in green.

#### Submit Crossref Deposit

* Click the "Assign DOI" button.
* Click the "Submit Deposit" button.
* The DOI field should no longer be editable, and the DOI should be replaced with a PubPub generated DOI, prefixed with `10.21428/`

Partial #910

### DIFF
--- a/client/containers/DashboardSettings/PubSettings/Doi.js
+++ b/client/containers/DashboardSettings/PubSettings/Doi.js
@@ -1,8 +1,10 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { FormGroup } from '@blueprintjs/core';
+import { FormGroup, ControlGroup, Button, InputGroup } from '@blueprintjs/core';
 
+import { apiFetch } from 'client/utils/apiFetch';
 import { getSchemaForKind } from 'utils/collections/schemas';
+import { isDoi } from 'utils/crossref/parseDoi';
 
 import { AssignDoi } from 'components';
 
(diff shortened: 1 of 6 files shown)


server/pub/permissions.js
```diff
@@ -18,6 +18,7 @@ export const getPermissions = async ({ userId, communityId, pubId, licenseSlug }
 		licenseSlug === 'cc-0';
 	const canManage = validLicenseSlug && scopeData.activePermissions.canManage;
 	const editProps = [
+		'doi',
```
idreyn, line 21: Only admins can update the DOI, so we need to add this to the list if `scopeData.activePermissions.canAdmin`.

client/containers/DashboardSettings/PubSettings/Doi.js
```diff
@@ -88,21 +180,28 @@ class Doi extends Component {
...
 		const { justSetDoi } = this.state;
+		const hasExistingDeposit = justSetDoi || Boolean(pubData.crossrefDepositRecordId);
 
 		return (
 			<div className="pub-settings-container_doi-component">
 				{this.renderStatusMessage()}
 				{this.renderCollectionContextMessage()}
 				{this.renderDoi()}
 
+				{!hasExistingDeposit && (
```
idreyn, line 193: I think there needs to be a separate bit of copy for non-admins, who get the rundown on what the button does but are unable to click it, as seen here:

image

For those users, who do not have any write access to the DOI, it may be simplest to have a different render branch that shows them the DOI if it exists, and either the "admins may assign a DOI" copy or nothing at all otherwise.

client/containers/DashboardSettings/PubSettings/Doi.js
```diff
@@ -69,15 +108,68 @@ class Doi extends Component {
...
+				<ControlGroup>
+					<InputGroup
+						label="DOI"
+						placeholder="Enter a DOI..."
+						value={doi}
+						onChange={(e) => this.setState({ doi: e.target.value })}
+					/>
+					<Button
+						disabled={!doi || invalidDoi}
+						text="Update DOI"
```
idreyn, line 166: Or just "Update"

PR #954, opened 2020-07-28
fix: use date helper to get published date for meta tags

### LINKED ISSUE #948: [BUG] Publication date is not added to HTML <meta> tags
I would like to have the publication date of the article added to the HTML <meta> tags in the header of Pubs. 
It could easily be done by simply including the tag:
`<meta name="citation_publication_date" content="2020/07/22">`  for instance.

More control over the meta tags would of course be amazing, but this simple addition would already help lots.

Google Scholar requires this in order to parse the rest of the meta tags in the Pub-HTML, this would further increase our chances of having our Pubs indexed, as currently only the PDF is indexed by Scholar.
image
Source: https://scholar.google.com/intl/en/scholar/inclusion.html#indexing

---

Absence of publication meta-tag.
image

https://jtrialerror.pubpub.org

gabestein:
Resolves #948 by updating the pub route to use the `getPubPublishedDate` helper instead of relying on now nonexistent `publishedAt` fields in `pubData`.

Also adds dc.date metadata if there is a published on date.

_Test plan_
1. Visit a Pub that is not released (ie http://demo.duqduq.org/pub/b7ag4toy/draft), verify that no `dc.date` nor `citation_publication_date` meta tag is present.
1. Visit a pub that is released (ie http://demo.duqduq.org/pub/xp0zizup/release/1), verify that `dc.date` meta tag is present with a result in the `YYYY-MM-DD` format, and verify that `citation_publication_date` meta tag is present with a result in the `YYYY/MM/DD` format.

### DIFF
--- a/server/utils/ssr.js
+++ b/server/utils/ssr.js
@@ -197,10 +197,12 @@ export const generateMetaComponents = ({
 	if (publishedAt) {
 		const googleScholarPublishedAt = `${publishedAt.getFullYear()}/${publishedAt.getMonth() +
 			1}/${publishedAt.getDate()}`;
+		const dcPublishedAt = `${publishedAt.getFullYear()}-${publishedAt.getMonth()}-${publishedAt.getDate()}`;
 		outputComponents = [
 			...outputComponents,
 			<meta key="pa1" property="article:published_time" content={publishedAt} />,
-			<meta key="pa2" name="citation_publication_date" content={googleScholarPublishedAt} />,
+			<meta ket="pa2" property="dc.date" content={dcPublishedAt} />,
+			<meta key="pa3" name="citation_publication_date" content={googleScholarPublishedAt} />,
 			<meta key="pub1" name="citation_publisher" content="PubPub" />,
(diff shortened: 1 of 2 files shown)


server/utils/ssr.js
```diff
@@ -197,10 +197,12 @@ export const generateMetaComponents = ({
 	if (publishedAt) {
 		const googleScholarPublishedAt = `${publishedAt.getFullYear()}/${publishedAt.getMonth() +
 			1}/${publishedAt.getDate()}`;
+		const dcPublishedAt = `${publishedAt.getFullYear()}-${publishedAt.getMonth()}-${publishedAt.getDate()}`;
 		outputComponents = [
 			...outputComponents,
 			<meta key="pa1" property="article:published_time" content={publishedAt} />,
-			<meta key="pa2" name="citation_publication_date" content={googleScholarPublishedAt} />,
+			<meta ket="pa2" property="dc.date" content={dcPublishedAt} />,
```
3mcd, line 204: Minor typo here: `ket` -> `key`

gabestein: Whoops, good catch, thanks!

PR #955, opened 2020-07-28
feature: generate crossref relations from pub connections

3mcd:
This PR improves the Crossref deposit submission process by automatically generating relations from Pub Connections. The relations that would be generated for a deposit can be reviewed in the `AssignDoiPreview` component in both human readable and XML formats. 

**Test Plan**
* Create a Pub with one or more inbound _and_ outbound connections.
* Click the "Assign DOI" (or "Resubmit DOI deposit") button.
* Preview the connections that would be deposited in the `AssignDoiPreview` component:

    Screen Shot 2020-07-28 at 2 53 01 PM
* Repeat for a Pub whose primary collection is a Conference and a Pub whose primary collection is a Book.


utils/crossref/transform/pub.js
```diff
@@ -1,18 +1,71 @@
...
+		}
+	} else {
+		if (targetPub.doi) {
+			identifier = targetPub.doi;
+			identifierType = 'doi';
+		} else {
+			identifier = pubUrl(targetPub.community, targetPub);
+			identifierType = 'uri';
+		}
+	}
```
3mcd, line 30: Kind of an ugly block of code - open to suggestions

package.json
```diff
@@ -10,7 +10,7 @@
 	},
 	"main": "init.js",
 	"scripts": {
-		"api-dev": "NODE_PATH=./client:./ node init.js --watch ./server --watch ./shared --watch ./client",
+		"api-dev": "NODE_PATH=./client:./ node init.js --watch ./server --watch ./utils --watch ./client",
```
idreyn, line 13: Aha! I knew something was up here!

server/doi/queries.js
```diff
@@ -42,14 +41,12 @@ const findCollection = (collectionId) =>
...
 	Pub.findOne({
 		where: { id: pubId },
-		include: [
-			{ model: Release, as: 'releases' },
-			{
-				model: PubAttribution,
-				as: 'attributions',
-				include: [includeUserModel({ as: 'user' })],
+		...buildPubOptions({
+			getEdges: 'approved-only',
```
idreyn, line 45: This is the default and I think not strictly necessary.

gabestein:
I won't be able to get to it today, but will take a look at the deposit xml on the deployed qubqub first thing tomorrow.

package.json
3mcd: Yeah it was driving me crazy!

PR #960, opened 2020-07-29
feature: Editable DOI suffix for communities with custom DOI prefix

3mcd:
This PR makes the DOI suffix editable for communities with custom DOI prefixes. The field is no longer editable after the first deposit is made. If no custom DOI prefix is specified, the DOI will be automatically generated in the same manner it usually is.

A delete button was added to remove a DOI from a Pub. This button is also only accessible while the DOI is editable.

**Test Plan**
1. Create a new Pub within a community with a custom DOI prefix.
2. Navigate to the Pub settings page in the dashboard.
2. Assign a custom DOI prefix using the text field and clicking "Update".
3. Open the deposit preview by clicking the "Preview Deposit" button. The custom suffix should be applied to the DOI within the preview.
4. Assign a new custom DOI prefix by repeating step 2.
5. The custom suffix should update within the preview.
6. Click the "Delete" button. The text field should be cleared (except for the DOI prefix) and the DOI within the deposit preview should change the a PubPub-generated DOI.
7. Click the "Submit Deposit" button. The DOI should convert to read-only.
8. Reload the page. The DOI should remain read-only.

### DIFF
--- a/client/containers/DashboardSettings/PubSettings/Doi.js
+++ b/client/containers/DashboardSettings/PubSettings/Doi.js
@@ -222,29 +222,32 @@ class Doi extends Component {
 				{this.renderCollectionContextMessage()}
 				{this.renderDoi()}
 
-				{!hasExistingDeposit && (
-					<p>Use the button below to have PubPub deposit this work to Crossref.</p>
-				)}
-
-				{this.isDoiEditable() && (
-					<p>
-						PubPub will automatically assign a DOI if the suffix is left blank. Please
-						note that <strong>once submit, the DOI will no longer be editable.</strong>
-					</p>
-				)}
-
 				<FormGroup
 					helperText={
 						pubData.doi &&
 						!justSetDoi && (
-							<React.Fragment>
+							<>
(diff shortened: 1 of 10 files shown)


client/containers/DashboardSettings/PubSettings/Doi.js
```diff
@@ -238,6 +257,56 @@ class Doi extends Component {
...
+		const { doiSuffix, updating, deleting } = this.state;
+		const fullDoi = this.getFullDoi();
+		const invalidDoi = doiSuffix && !isDoi(fullDoi);
+		const intent = this.getIntent(invalidDoi);
+		const helperText = this.getHelperText(invalidDoi);
+
+		if (this.isDoiEditable()) {
+			return (
+				<FormGroup helperText={helperText} intent={intent}>
+					<ControlGroup>
```
idreyn, line 271: This is a matter of taste but I really don't care for the way Buttons look all smushed together in a ControlGroup. I would weakly prefer some space between them.

PR #961, opened 2020-07-29
fix: Fixes for date picking and display for ExternalPublication objects

### LINKED ISSUE #945: Date picker in Connections creation interface returns the previous day’s date
1. When manually adding or editing publication date in the Connections pane in the Pub dashboard, the date picker returns the day before the selected date. E.g., July 24, 2020 returns July 23, 2020.
2. Tested in Safari for iPad and Chrome for Mac

idreyn:
Resolves #945 

This is a quick PR to fix my rather goofy attempt at using the native `input type="date"` in the `ExternalPublication` connection creator interface. Things that change:

- We use the `input.valueAsDate` property where available, and fall back to parsing `input.value` in browsers where this isn't supported — as far as I can tell, this is just Safari, which makes you input a date by hand as well.
- We render the stored `publicationDate` in UTC time, so users in western time zones don't see it rendered as the day before.

image

_Test plan:_
- Make sure you can still create Connections to other Pubs as always
- In (Chrome, Firefox, Edge) make sure that you can select a publication date, and that the date, rather than an adjacent date, is actually chosen and displayed.
- In Safari make sure that you can type in a value in `YYYY-MM-DD` format, and that intermediate, invalid date values don't crash the browser. Trying to create the edge with an unparseable date string should cause a server-side error.
- Make sure that the dates displayed in the created Connections cards are the same as the ones chosen, irrespective of your own time zone.
- Make sure that `ExternalPublication` connections without a date don't show any date at all.

### DIFF
--- a/client/components/PubEdge/PubEdge.js
+++ b/client/components/PubEdge/PubEdge.js
@@ -9,6 +9,7 @@ import { pubUrl, pubShortUrl } from 'utils/canonicalUrls';
 import { getPubPublishedDate } from 'utils/pub/pubDates';
 
 import { pubEdgeType } from './constants';
+import { getHostnameForUrl } from './util';
 import PubEdgeLayout from './PubEdgeLayout';
 import PubEdgePlaceholderThumbnail from './PubEdgePlaceholderThumbnail';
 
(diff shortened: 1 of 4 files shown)


3mcd:
Approved - but I did notice one thing during some subsequent testing:

1) Create a connection to https://futurespodcast.net/episodes/19-edwardalee
2) Leave the date field empty.
3) Notice the date appears as July 22nd, 2001:

Screen Shot 2020-07-29 at 1 54 01 PM

idreyn:
Nice catch with that test case — the suggested date is actually 7/22/2001, but the problem is that we should have been showing it!

PR #967, opened 2020-07-31
feature: add content version field and preprint (posted_content) deposits

### LINKED ISSUE #912: [PC DOI Deposits] Deposit Status
Deposit status — adds a field to support different crossmark statuses (can be deposited with main pub deposit, see schema). For now, we should support:

Preprint

once deposited with this status, it cannot be changed, and should also update the type of submission to be posted_content with a type of preprint

Pending publication

Advanced online publication

Author accepted manuscript

Version of record

3mcd:
Closes #912

This PR adds a new field to the deposit form that allows a Pub admin to update the resource's `content_version`.

Screen Shot 2020-07-31 at 4 32 19 PM

The field's options are detailed below:

* **(None)** un-sets the `<resource>` element's `content_version` attribute of the work
* **Advance Manuscript** sets the `<resource>` element's `content_version` attribute to `"am"`(advance manuscript)
* **Version of Record** sets the `<resource>` element's `content_version` attribute to `"vor"`
* **Preprint** converts the record to a `posted_content` item

Since changing the content version of the work results in changes to the deposit, the entire deposit preview is re-fetched each time the dropdown's selected item changes, as demonstrated below:

(sorry for the blurry gif)

**This PR does not include:**

- Automatically generating more specific deposits for peer reviews and supplementary material.
- Defaulting to preprint content version for Pubs with an `isPreprintOf` edge.

**Test Plan:**

1. Create a new Pub and publish it.
2. Click the "Preview Deposit" button in the Pub settings.
3. Toggle between the various options in the content version dropdown and verify they work as described above.

### DIFF
--- a/client/components/AssignDoi/AssignDoi.js
+++ b/client/components/AssignDoi/AssignDoi.js
@@ -34,7 +34,7 @@ const defaultProps = {
 	onPreview: noop,
 };
 
-const AssignDoiState = {
+const AssignDoiStatus = {
 	Initial: 'initial',
 	Previewing: 'previewing',
 	Previewed: 'previewed',
@@ -51,20 +51,20 @@ const AssignDoiActionType = {
 	Error: 'error',
 };
 
-const buttonTextByState = {
-	[AssignDoiState.Initial]: 'Preview Deposit',
-	[AssignDoiState.Previewing]: 'Getting Preview',
-	[AssignDoiState.Previewed]: 'Submit Deposit',
-	[AssignDoiState.Depositing]: 'Depositing',
-	[AssignDoiState.Deposited]: 'DOI Deposited',
+const buttonTextByStatus = {
+	[AssignDoiStatus.Initial]: 'Preview Deposit',
+	[AssignDoiStatus.Previewing]: 'Getting Preview',
+	[AssignDoiStatus.Previewed]: 'Submit Deposit',
+	[AssignDoiStatus.Depositing]: 'Depositing',
+	[AssignDoiStatus.Deposited]: 'DOI Deposited',
 };
 
-const getButtonText = (state, deposit) => {
-	if (state === AssignDoiState.Initial && deposit) {
+const getButtonText = (status, deposit) => {
+	if (status === AssignDoiStatus.Initial && deposit) {
 		return 'Resubmit DOI deposit';
 	}
 
-	return buttonTextByState[state];
+	return buttonTextByStatus[status];
 };
 
 export function reducer(state, action) {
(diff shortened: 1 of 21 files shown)


gabestein:
Looks awesome! One very small request: could we order the dropdown (none), preprint, advance manuscript, version of record, since that's sort of the logical order?

3mcd:
Yep you got it!

client/components/AssignDoi/AssignDoiPreview.js
```diff
@@ -162,35 +172,28 @@ const renderBookPreview = (doi_batch) => {
...
-					contributors,
-					titles: paperTitles,
-					'rel:program': { related_item: relatedItems },
-				},
-				event_metadata: {
-					conference_name,
-					conference_date: { '#text': conferenceDate },
-				},
-				proceedings_metadata: { proceedings_title, publication_date, publisher },
+		conference: {
```
idreyn, line 182: This has got to be one of the deepest destructurings I've ever seen. Respect.

client/components/AssignDoi/AssignDoi.js
```diff
@@ -64,7 +70,14 @@ const getButtonText = (state, hasExistingDeposit) => {
 export function reducer(state, action) {
 	switch (action.type) {
 		case AssignDoiActionType.FetchPreview:
-			if (!(state.state === AssignDoiState.Initial || state.state === AssignDoiState.Error)) {
+			if (
+				!(
+					state.state === AssignDoiState.Initial ||
```
idreyn, line 75: Definitely no need to change this here, but I find that calling this kind of value a `status` helps distinguish it from the containing `state` object.

server/doi/api.js
```diff
@@ -8,7 +8,9 @@ import { getPermissions } from './permissions';
 const previewOrDepositDoi = async (req, options = { deposit: false }) => {
 	const { deposit } = options;
 	const user = req.user || {};
-	const { target, communityId, collectionId, pubId } = deposit ? req.body : req.query;
+	const { target, communityId, collectionId, pubId, contentVersion } = deposit
```
idreyn, line 11: Looking at this with fresh eyes, I think it's better to have the callers pass in `req.body` or `req.query` as appropriate, rather than using the implicit assumption that `deposit: false` means that `req.method = 'GET'`.

utils/crossref/createDeposit.js
```diff
@@ -6,11 +6,16 @@ import doiBatch from './schema/doiBatch';
...
 import renderJournal from './render/journal';
+import renderPreprint from './render/preprint';
 import createDoi from './createDoi';
 import getCollectionDoi from '../collections/getCollectionDoi';
 
 const renderBody = (context) => {
-	const { collection } = context;
+	const { collection, globals } = context;
+
+	if (globals.contentVersion === 'preprint') {
```
idreyn, line 16: Very cool how straightforward this bit turned out to be!

utils/crossref/parseDeposit.js
```diff
@@ -0,0 +1,79 @@
...
+			j++;
+		}
+
+		i++;
+	}
+
+	return null;
+};
+
+export const setDepositRecordContentVersion = (depositRecord, contentVersion) => {
```
idreyn, line 37: Okay, I think I understand your earlier comment in the reducer a little better now. It is odd that this function mutates its argument (right?) but then also returns it — it makes me think this should have a return type of `void` to make that explicit to the caller. But that makes it harder to use in the reducer...so, I don't have a concrete idea of what the best thing to do here is.

utils/crossref/transform/collection.js
```diff
@@ -19,13 +20,17 @@ const transformMetadata = (metadata, kind, timestamp) =>
...
 export default ({ globals, community }) => (collection) => {
-	const { title, metadata, attributions } = collection;
+	const { timestamp, dois, contentVersion } = globals;
+	const { crossrefDepositRecord, title, metadata, attributions } = collection;
 	return {
 		url: collectionUrl(community, collection),
 		...transformMetadata(metadata, collection.kind, globals.timestamp),
 		title: title,
-		timestamp: globals.timestamp,
+		timestamp: timestamp,
```
idreyn, line 29: In retrospect this design where you have to pass `timestamp` around like a hot potato is a little silly, sorry

utils/crossref/parseDeposit.js
3mcd: I'll make it a void function that way it's clearer! Good idea.

PR #968, opened 2020-07-31
feature: Custom publication date picker for Pubs

### LINKED ISSUE #903: Add customDate setting in Pub Settings

idreyn:
This PR allows users to set the `customPublishedAt` field that is displayed in place of the first Release date for a Pub as its publication date, as shown here:

Screen Recording 2020-07-31 at 4 54 30 PM 2020-07-31 16_56_26

There is a new `DatePicker` component in this PR that will make this sort of thing easier in the future. It uses the browser's built-in datepicker, which is good and getting better in most browsers, but still requires desktop Safari users to manually type dates in the indicated YYYY-MM-DD format, which is not the worst fate. The innards of this component do not look like idiomatic React, but its interface is much nicer than `<input type="date">` as it will only return a valid `Date` object or `null`.

_Test plan:_
- In the Storybook fixture for `DatePicker`, make sure the component works (including picking, typing, and clearing dates) in Firefox, Chrome, Safari, and Edger.
- Make sure that the `customPublishedAt` field works as expected as shown in the GIF above. Try picking dates from both sides of US daylight savings to make sure offsets are calculated correctly throughout the year when reverse-engineering a local date from a UTC date.

Resolves #903

### DIFF
--- a/client/components/DatePicker/DatePicker.js
+++ b/client/components/DatePicker/DatePicker.js
@@ -1,3 +1,10 @@
+/**
+ * This component picks dates in UTC time, e.g.: March 10, 2018 is `Fri Mar 10 2018 00:00:00 GMT+0`.
+ * This means that without special handling, the dates it provides may render unexpectedly in
+ * certain timezones, very possibly the one in which you are reading this code! It is the caller's
+ * responsibility to handle these dates correctly, perhaps by using the inUtcTime option of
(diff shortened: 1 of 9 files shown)


gabestein:
Looks good from my testing! Also looked at the PDF, html metadata, RSS, and it's all good, even if I enter boundary dates like 2018-12-31!

Issue #930, opened 2020-07-19
Bug: Citation Creation Broken on Pages


writingpoli:
**What went wrong, step-by-step?**
1. Inserted citation on sareview.org/about (formatting and link options still didn't show up -- even using Chrome)
2. No footnote or citation created on published page. 
...

**What did you expect to happen?**

Pop-up or footnote at bottom of page, like in Pub.

**What URL can we use to see the issue?**

www.sareview.org/about

**What browser/operating system were you using?**

Google Chrome, Windows 10

---

**Have any screenshots? (Attach them by dragging the image here)**
![image](https://user-images.githubusercontent.com/3883763/87881905-1f0cea80-c9c2-11ea-9bff-fead6dc5c9ab.png)

idreyn:
Thanks for writing in about this. We restructured our citations code quite some time ago in a way that didn't account for their use on Pages. This is a somewhat rare usage, but one we definitely want to restore support for in the near future. In the interim, can a hyperlink serve this purpose?

PR #972, opened 2020-08-04
feature: Reactive Prosemirror node attributes

idreyn:
[This post from Travis](https://discourse.knowledgefutures.org/t/editor-numbering-and-ids/244) gives most of the context for both why this is needed and the approach taken in this PR, but in brief, we need a way to reliably render content to Prosemirror Nodes _other_ than what's in the Node's attributes. This is tricky, since `toDOM` is supposed to be a pure function of a `Node` (rather than any editor state), so we do the following:

- Use Travis' approach to generating stable IDs for elements in the document
- Create a plugin that holds a state store mapping stable node IDs to volatile (non-document) attributes
- Provide an API for custom functions to derive attribute values from editor state, promise results, etc. Right now I am calling those functions "cells", but I may choose a more straightforward name like "reactive values", or pick from [this list](https://twitter.com/sophiebits/status/1055711617604759552).
- On every transaction, let every cell update. If it changes value since the last transaction, invalidate its Node and force it to re-render by applying a dummy Prosemirror decoration to it.
- Finally, and this is the tricky part: for reactive node types, provide a custom `NodeView` function that manually renders the node based on its attributes _and its reactive attributes_ first into a `DOMOutputSpec` and then into actual DOM nodes.

So there's a lot left to do here:

-  Figure out an SSR approach (we'll have to do something in `renderStatic`) that can hopefully be reused when generating exports server-side.
- Remove all editor-external sources of truth (like `citationsRef`) for citation content and move that gathering into a cell.
- Replace `PubMouseEvents` (which relies on external sources of truth) with editor-internal popovers for footnotes and citations. This is a boon for accessibility as well as reuse of these nodes across the editor (see #930)
- And more...

But this is an exciting start! (I won't merge this into `master` — I'll create some other feature branch for it)

_Test plan:_ Open a new document, create several citations, change their contents and move them around, etc. — and verify that the numbering stays up to date. For now, SSR is broken, the citation popups are broken, footnotes are broken — that's okay.

test


client/components/Editor/plugins/ids.js
```diff
@@ -0,0 +1,93 @@
+import { Plugin } from 'prosemirror-state';
```
idreyn, line 1: This is a lightly retouched version of Travis' plugin for this here: https://github.com/pubpub/pubpub/blob/21391827664bb8ab15ea4aebedb3ddd37d8d1d62/client/components/Editor/plugins/hashes.js

client/components/Editor/plugins/reactive/cells/promise.js
```diff
@@ -0,0 +1,25 @@
+const compareArrayFingerprints = (from, to) => {
```
idreyn, line 1: We don't use this anywhere yet, and it will probably change or disappear, but it provides some more flavor for what a reactive cell looks like.

idreyn:
> SSR should roughly come for free as long as we're still using toDOM at the end of the day.

This approach is a little different from yours in that the NodeViews provide a "stateful" wrapper around relevant `toDOM` definitions that allow us to pass individualized reactive attributes into our Nodes. Since SSR happens with a single call to `renderStatic()`, we don't have a chance to inject these anywhere in that flow. We'll have to:

1. Initialize a reactive store
2. Touch every node in the same way the reactive plugin does to compute reactive attrs
3. Walk the tree and augment every node with its newly computed reactive attrs
4. And _then_ `toDOM` will know what to do.

isTravis:
Right - that's helpful.

Digging in a bit, the way we achieve this currently is by using the `buildSchema` function to edit the json that defines the schema (by overwriting `defaultOptions` when possible). That all assumes we'll be passing in the state through `<Editor/>` props - so I see now why a complete reactive solution needs to re-implement that initialization step.

Thanks!

PR #984, opened 2020-08-13
dev: Add snapshot testing for searchUtils functions

idreyn:
Unexpected changes to the behavior of the functions in `searchUtils.js` force us to retroactively update the Algolia index, which is time consuming and expensive, so I've added a bit of snapshot testing for them here.

_Test plan:_

Run `npm run test workers/utils`


tools/searchSync.js
```diff
@@ -6,18 +6,17 @@ import { Pub, Page } from '../server/models';
 import { deletePubSearchData } from '../workers/tasks/search';
```
idreyn, line 6: Oops — I am messing around with this file locally (and actually running it right now, so I am a little afraid to revert these changes) but this really isn't part of the proposed changes.

PR #986, opened 2020-08-14
feature: remove abbrev_title from journal deposits

### LINKED ISSUE #985: Remove abbrev_title from doi deposits
It's not required, we don't have a field for it so we're using the full community title, and there's an undocumented 150-word character limit that some communities are hitting.

3mcd:
Closes #985

Fixes an issue where deposits for communities with long titles would fail due to a 150 character restriction on the `abbrev_title` field. This PR simply removes the optional `abbrev_title` element from `journal_metadata` deposits.


gabestein:
Thanks. LGTM!

PR #995, opened 2020-08-18
feature: sitemap index and dynamic robots.txt

### LINKED ISSUE #976: Add sitemaps for communities
https://github.com/ekalinin/sitemap.js

3mcd:
Closes #976

This PR adds sitemaps to communities.

The static robots.txt file is replaced with a dynamic file which builds the robots file on a per-community basis. The robots.txt file for communities will now point to the sitemap via the `Sitemap` rule, e.g. `Sitemap: http://community.pubpub.org/sitemap-index.xml`.

If a sitemap has not been generated for a community, or the sitemap was updated over 24 hours ago, a sitemap index and child sitemaps (e.g. sitemap-0.xml, sitemap-1.xml, ...) will be generated and uploaded to a new S3 bucket, `sitemaps.pubpub.org`. The new bucket structure looks like:

```
sitemaps.pubpub.org
  dev/
    1acefbd-99ecba-301dbe-ffc9ae/
      sitemap-index.xml
      sitemap-0.xml
      ...
  prod/
    ...
```

Where dev and prod directories correspond to the application environment, and the hash corresponds to a community ID.

If a sitemap artifact is requested and the sitemap index for that community has been updated within the last 24 hours, the file will be streamed directly from S3.

The sitemap XML artifacts are generated using the `sitemap` npm library. `sitemap` will automatically chunk sitemaps for every 45,000 entries. To my knowledge, no PubPub communities have >=45,000 combined Pubs and Pages, but this should put us in a good position to scale sitemaps for communities of that size in the future.

Records for Pages are weighted slightly below Pubs. This means (hypothetically) that search engines will crawl Pubs more frequently than pages. Pubs are also labeled as being updated more frequently ("weekly" as opposed to pages' "monthly"), which also signals to search engines that Pubs have a higher rate of change than Pages.

This feature uses Node streams in a few spots, which admittedly aren't my forte. So if you have any feedback or tips, I'm all ears.

**Test Plan:**
- Visit any community's robots.txt file and ensure the `Sitemap: ` rule is present and looks like `http://{community}.pubpub.org/sitemap-index.xml`.
- Explore the sitemaps.pubpub.org S3 bucket.
- Visit any community's sitemap-index.xml file and verify that the sitemap index loads.
- Visit the sitemap-0.xml file and verify that all pages and Pubs of that community have a <url> record.
- Verify the structure of the S3 bucket matches the example above.

### DIFF
--- a/server/routes/robots.js
+++ b/server/routes/robots.js
@@ -4,21 +4,27 @@ import { getInitialData } from 'server/utils/initData';
 import { hostIsValid } from 'server/utils/routes';
 import { communityUrl } from 'utils/canonicalUrls';
 
-const getCommunityRobots = (community) => `User-agent: *
-Disallow:
-Sitemap: ${communityUrl(community)}/sitemap-index.xml
-`;
+const BASE_ROBOTS = `User-agent: *
+Disallow:`;
+
+const buildRobotsFile = (community) => {
+	if (community) {
+		return `${BASE_ROBOTS}
+Sitemap: ${communityUrl(community)}/sitemap-index.xml`;
+	}
+	return BASE_ROBOTS;
+};
 
 app.get('/robots.txt', async (req, res, next) => {
 	if (!hostIsValid(req, 'community')) {
-		return next();
+		return buildRobotsFile();
 	}
 
 	try {
 		const { communityData } = await getInitialData(req, true);
 
 		res.header('Content-Type', 'text/plain');
-		return res.send(getCommunityRobots(communityData));
+		return res.send(buildRobotsFile(communityData));
 	} catch (err) {
 		return handleErrors(req, res, next)(err);
 	}
--- a/server/routes/robots.js
+++ b/server/routes/robots.js
@@ -1,5 +1,4 @@
-import app from 'server/server';
-import { handleErrors } from 'server/utils/errors';
+import app, { wrap } from 'server/server';
 import { getInitialData } from 'server/utils/initData';
 import { hostIsValid } from 'server/utils/routes';
 import { communityUrl } from 'utils/canonicalUrls';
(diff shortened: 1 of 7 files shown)


server/routes/sitemap.js
```diff
@@ -0,0 +1,141 @@
...
+		}))
+		.concat(
+			pages.map((page) => ({
+				url: pageUrl(community, page),
+				changefreq: 'monthly',
+				priority: 0.8,
+			})),
+		);
+
+	await pipeline(
```
idreyn, line 89: Very cool, I have never really figured out pipelines but this looks super elegant!

server/routes/sitemap.js
```diff
@@ -0,0 +1,141 @@
...
+import { hostIsValid } from 'server/utils/routes';
+import { communityUrl, pubUrl, pageUrl } from 'utils/canonicalUrls';
+
+const s3 = new AWS.S3({ params: { Bucket: 'sitemaps.pubpub.org' } });
+const pipeline = promisify(stream.pipeline);
+const keyPrefix = isProd() ? 'prod' : 'dev';
+
+const uploadFromStream = (key) => {
+	const pass = new stream.PassThrough();
+	const params = { Key: key, Body: pass, ACL: 'public-read' };
```
idreyn, line 21: _(Hacker News guy with a flashlight up to his face)_ but when they went back to check the bucket a few days later...its ACL had set to `public-read-write`! 👻

server/routes/robots.js
```diff
@@ -0,0 +1,31 @@
...
+	}
+	return BASE_ROBOTS;
+};
+
+app.get('/robots.txt', async (req, res, next) => {
+	if (!hostIsValid(req, 'community')) {
+		return buildRobotsFile();
+	}
+
+	try {
```
idreyn, line 23: We have a mysterious function called `wrap` in server.js that lets you write async Express routes without having to do this error handling by hand. Hopefully the library should be able to handle this by itself soon...

3mcd: Yeah I saw it mostly used in `api` files so I didn't know if it was only idiomatic to use it there. I'll use it here!

PR #998, opened 2020-08-19
update: Reactive citations and footnotes — tooltips (2/3)

idreyn:
This is a follow-up to #997 that removes the responsibility of rendering hover tooltips for citations and footnotes from `PubMouseEvents`, and adds it to a new Prosemirror `NodeView`. Here is why:

- `PubMouseEvents` only works in the Pub body, which means that footnotes and citations are unable to render elsewhere.
- It only works with mouse events, which makes tooltips inaccessible.
- It has occasional problems with tooltips falling out of sync when citations change in the document.
- It won't work in an environment where we show several documents at once, as in some of our hypothetical Review UI.

So, moving the responsibility for rendering tooltips back into Editor-land feels like a step in right direction for me. @isTravis, one question I have about the current implementation is whether we ought to keep the 250ms timeout associated with mouse events? I rather like how snappy this is now, but I am not sure what the original intent was there so I wanted to call it out.

_Test plan:_
Open a Pub with both footnotes and citations and verify that, by hovering and tabbing, you are able to access tooltips for footnotes and citations. Check that they render structured and unstructured content as expected.


idreyn:
> Approving the PR, but I think we need to add some delay or other functionality back so your mouse can reach the popover box

Good catch, that totally makes sense. The approach I landed on removes the tooltip only when it leaves the rectangle enclosing both the citation/footnote node and the tooltip, giving you some wiggle room in how you move your mouse to get there. There is an edge case where you can open multiple tooltips for adjacent citations — which I view as acceptable, since the most recently hovered one will never be occluded:

Screen Shot 2020-08-24 at 6 04 04 PM

If folks end up really disliking this, we can hack something in to enforce that there's only ever one popover open in a given editor.

> My understanding was that [`mouseenter`] would produce fewer unnecessary events

We don't attach any mouse events to the tooltip itself, but this is a great thing to know generally!

> I think at one point we had issues with the popover showing above the formatting bar.

This seems to have gone away, but good to have tested.

PR #999, opened 2020-08-19
update: Reactive citations and footnotes — exports (3/3) 

### LINKED ISSUE #876: Citation loop causes large Pub size
Citation plugin can cause loops which creates giant pubs that crash PubPub via Firebase timeout. A couple of levels – need to fix at some level.

idreyn:
Following on from #997 and #998, this re-enables exports in a post-reactive world. By line changes, the bulk of this PR is just moving around and renaming files, and I'll call those out in the comments. I'm sort of unhappy with the number of steps required to generate an export these days, and I'm adding a bit more fuel to that fire — you can't really tell from here but during the process we rehydrate and walk the Prosemirror document several times — but this will do for now.

Test plan:
- In a Pub with lots of footnotes and citations, verify that PDF and HTML exports work, and that citations and footnotes link correctly to their full text at the bottom of the document and back, especially for citations that are repeated multiple times throughout the document.
- Make sure exports that run through Pandoc (like TeX) work as well.

Resolves #876


workers/tasks/export/SimpleNotesList.js
```diff
@@ -0,0 +1,49 @@
+/**
```
idreyn, line 1: Moved from `components` — it's too tightly coupled to the export code to be work shooting for reusability, IMO.

workers/tasks/export/html.js
```diff
@@ -75,27 +76,49 @@ const addAttrsToNodes = (newAttrs, matchNodeTypes, nodes) =>
...
+};
+
+const getFootnoteLinkage = (index) => {
+	return {
+		inlineElementId: `fn-${index}-inline`,
+		bottomElementId: `fn-${index}-bottom`,
+	};
+};
+
+const addHrefsToNotes = (nodes) => {
```
idreyn, line 94: This change is necessary because we don't have access to `count` values at this point in the process. If it turns out that needing to read reactive attributes becomes a more common part of the export process, I will want to rethink this, but for now it seems simpler to manually assign IDs to citations by their value — thus allowing us to link them to the document-bottom notes sections — than to move the "reacting" of the document further up in the pipeline.

workers/tasks/export/notes.js
```diff
@@ -0,0 +1,43 @@
+import { jsonToNode, getNotes } from 'components/Editor';
```
idreyn, line 1: This is the spiritual successor to the deleted `prosemirror.js`, but it only does notes-related things now.

PR #1019, opened 2020-09-02
update: include pages where user is collection manager

3mcd:
This PR updates page sanitation logic to also include collection pages (even when they are private) in page results when the requesting user has collection-level manage permission.

_Test Plan_
* Link a private page to a collection.
* Log in to a user with community-level view or edit permissions (without any collection permissions).
* The page should not be visible in the nav bar.
* Assign the user collection-level view or edit permissions.
* The page should still not be visible in the nav bar.
* Assign the user manage permissions.
* The page should now be visible in the nav bar, and clicking on the link should take you to the private page.

### DIFF
--- a/server/utils/queryHelpers/communitySanitize.js
+++ b/server/utils/queryHelpers/communitySanitize.js
@@ -29,14 +29,13 @@ export default (communityData, locationData, loginData, scopeData) => {
 			({ pageId }) => typeof pageId === 'string' && pageId === item.id,
 		);
 
-		// If the page has a collection, check if the user has manage permission
+		// If the page has a collection, check if the user has >= manage permission
 		// to that collection. If so, include the page.
 		const hasPageCollectionManageAccess = pageCollection
-			? Boolean(
-					pageCollection.members.find(
-						(member) =>
-							member.userId === loginData.id && member.permissions === 'manage',
-					),
+			? pageCollection.members.some(
+					(member) =>
+						member.userId === loginData.id &&
+						(member.permissions === 'manage' || member.permissions === 'admin'),
 			  )
 			: false;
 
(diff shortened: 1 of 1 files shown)


server/utils/queryHelpers/communitySanitize.js
```diff
@@ -35,5 +23,39 @@ export default (communityData, locationData, loginData, scopeData) => {
 				page: availablePages[collection.pageId],
 			};
 		});
+
+	cleanedData.pages = cleanedData.pages.filter((item) => {
```
3mcd, line 27: I moved the page filter below the collection sanitation logic to only consider collections where the user has explicit permission when checking if a page should be visible due to collection-level permissions.

server/utils/queryHelpers/communitySanitize.js
```diff
@@ -35,5 +23,39 @@ export default (communityData, locationData, loginData, scopeData) => {
...
+		const pageCollection = cleanedData.collections.find(
+			({ pageId }) => typeof pageId === 'string' && pageId === item.id,
+		);
+
+		// If the page has a collection, check if the user has manage permission
+		// to that collection. If so, include the page.
+		const hasPageCollectionManageAccess = pageCollection
+			? Boolean(
```
idreyn, line 35: We should be able to use `pageCollection.members.some()` here

server/utils/queryHelpers/communitySanitize.js
```diff
@@ -35,5 +23,39 @@ export default (communityData, locationData, loginData, scopeData) => {
...
+					pageCollection.members.find(
+						(member) =>
+							member.userId === loginData.id && member.permissions === 'manage',
```
idreyn, line 38: We also want to check for the `admin` permission here

...really what we want is to break a bunch of utility functions out of `scopeGet.js`, which handles all of these edge cases correctly, but in mega-functions which would be convenient to access as a collection of smaller functions.

idreyn (approved):
Looks great! Just make sure to account for the `admin` permission as noted inline and it's good to go.

PR #1021, opened 2020-09-03
fix: Shore up guest Pub creation mechanisms

idreyn:
This PR re-enables a couple of ways that non-Members can create Pubs that I inadvertently disabled in #1001  

- The global header's Create Pub button when shown to non-Members
- A Page banner block with a Pub creation button enabled (which I only learned about this week!)

Previously these were permitted because there was no permission gating on Pub creation. Re-enabling the global header button is as easy as checking `hideCreatePubButton` on a Community (though this field should be renamed). Enabling this for Page banner blocks proved a little trickier since the permission exists by virtue of the visitor being able to see the Page — ultimately, I went with an approach where we issue expiring JSON web tokens to users visiting Pages with these banner blocks, allowing them to create Pubs.

_Test plan:_
I checked all the ways I know how to create Pubs, including:
- The global nav, as an admin
- The Community Overview dashboard tab
- A Collection's Overview dashboard tab (the button should add the new Pub to the Collection)
- The global nav, as a non-Member with `hideCreatePubButton = false`
- A Page banner block with a `create-pub` button

Note that you will need a truthy `process.env.JWT_SIGNING_SECRET` for this to work for you. It doesn't matter at all what it is, but you'll find one if you go grab the latest config.js

### DIFF
--- a/server/pub/queries.js
+++ b/server/pub/queries.js
@@ -1,3 +1,5 @@
+import Bluebird from 'bluebird';
+
 import { Collection, Community, Pub, PubAttribution, Branch, Member } from 'server/models';
 import { setPubSearchData, deletePubSearchData } from 'server/utils/search';
 import { createCollectionPub } from 'server/collectionPub/queries';
@@ -6,7 +8,7 @@ import { generateHash } from 'utils/hashes';
 
 const months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
 
-export const createPub = async ({ communityId, collectionId, ...restArgs }, userId) => {
+export const createPub = async ({ communityId, collectionIds, ...restArgs }, userId) => {
 	const newPubSlug = generateHash(8);
 	const date = new Date();
 	const dateString = `${months[date.getMonth()]} ${date.getDate()}`;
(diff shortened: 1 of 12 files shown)


server/pub/queries.js
```diff
@@ -53,21 +55,24 @@ export const createPub = async ({ communityId, collectionId, ...restArgs }, user
...
-					return createCollectionPub({
-						collectionId: collectionIdToAdd,
-						pubId: newPub.id,
-					});
-				}
-				return null;
-			}),
+	const allCollectionIds = [...(defaultPubCollections || []), ...(collectionIds || [])];
+
+	const createCollectionPubs = Bluebird.each(
```
idreyn, line 60: We have to do these in serial because the check that makes CollectionPubs primaries relies on other CollectionPubs that have been created, and a race condition can create two primaries :(

server/utils/queryHelpers/pageGet.js
```diff
@@ -80,13 +81,41 @@ const getPubsForPageLayout = async (layout, initialData) => {
...
+				if (buttonType === 'create-pub') {
+					return {
+						...block,
+						content: {
+							...content,
+							createPubToken: issueCreatePubToken({
+								userId: userId,
+								communityId: communityData.id,
+								createInCollectionIds: defaultCollectionIds,
+							}),
```
3mcd, line 108: Am I right to think that if a user sits on a page for 1hr without clicking the button, then tries to click it, they will get a 403 error? If so, how is that error presented to the user?

3mcd (approved):
LGTM! Just left a small question.

server/utils/queryHelpers/pageGet.js
idreyn: Well, I was going to let them flounder, but I woke up more motivated this morning and added this:

image

PR #1031, opened 2020-09-08
feat: initial reference work; link to internal pub nodes w/ ids

3mcd:
This PR accomplishes two main things: 
- duplicates the baseline functionality of references (still WIP) from the `reactive-demo` branch
- adds the ability to link to any node with an id in the editor, including heading, paragraph, audio, video, image, file, citation, footnote, equation, and table nodes.

captured

Any node with an id at the top level of the document is eligible for linking. Linking works identically to how it did prior to this PR. The PubMouseEvents component now uses event delegation to avoid potential performance issues in documents with a large number of nodes with ids (which now include paragraphs, too!). To get event delegation working, I switched from using "mouseenter" and "mouseleave" events to the bubbling "mouseover" and "mouseout" events.

This PR also adds typing of props of the Menu/MenuItem components, and removes a bunch of `// @ts-expect-error` comments.

Test Plan:
- Create a document with one of each of the above node types (heading, paragraph, audio, video, image, file, citation, footnote, equation, and table).
- Ensure when the elements are hovered, a link button appears outside of the pub body column to the left of the element.
- Ensure that content-less paragraphs do not display the link button when hovered.
- Clicking on the link button should copy a "permanent" link to the element in the document.
- Move an element (i.e. copy + paste it) and ensure the link maintains the reference to the correct element.
- Export the Pub and ensure figures are rendered correctly.


client/components/Editor/plugins/table.ts
```diff
@@ -1,5 +1,22 @@
...
+import { columnResizing, tableEditing, goToNextCell, TableView } from 'prosemirror-tables';
+
+class PubTableView extends TableView {
+	constructor(node, whatever) {
+		super(node, whatever);
+		this.syncId(node);
+	}
+
+	update(node) {
+		super.update(node);
```
idreyn, line 12: ⚠️ This should return the value of `super.update(node)` to indicate to Prosemirror that the node should (not?) be redrawn

client/components/Editor/schemas/base.ts
```diff
@@ -7,34 +7,53 @@ export const baseNodes = {
 	},
 	paragraph: {
 		selectable: false,
+		// reactive: true,
 		content: 'inline*',
 		group: 'block',
 		attrs: {
+			id: { default: null },
```
idreyn, line 14: I should have thought of this earlier, but we should do some research into how adding IDs to paragraphs and other common block text nodes will affect the collaborative editing experience — for instance whether two clients attempting to open a paragraph in the same place will now open two paragraphs rather than having their changes merged into one. That could actually be a good thing! But I want to see what's up, so this is a note to myself to play around with it a little bit in a REPL.

client/components/Editor/schemas/reference.ts
```diff
@@ -0,0 +1,65 @@
...
+		selectable: true,
+		group: 'inline',
+		attrs: {
+			id: { default: null },
+			targetId: { default: null },
+		},
+		reactiveAttrs: {
+			label: function(this: Hooks, node) {
+				const { targetId } = node.attrs;
+				const { blockNames } = node.type.spec.defaultOptions;
```
idreyn, line 18: How would you feel about storing this in the `documentState` argument of the (pubpub-side) `reactive.ts` plugin wrapper, and then retrieving it with the `useDocumentState` hook? The name is a bit of a misnomer because hooks won't actually re-run based on mutations to (yet), but it's the replacement I had in mind for the use of `node.type.spec.defaultOptions`.

client/components/Editor/styles/figure.scss
```diff
@@ -1,29 +1,43 @@
 figure {
+	display: block;
```
idreyn, line 2: IIRC these changes don't result in any visible differences, it just makes the figure boundaries conveniently wider, right? If so — please check on what happens in the full-bleed (breakout) mode if you haven't already 🙂

client/components/FormattingBar/buttons.ts
```diff
@@ -126,6 +129,16 @@ export const citation = {
 	controls: nodeControls(ControlsFootnoteCitation, 'citation'),
 };
 
+export const reference = {
+	key: 'reference',
+	title: 'Reference',
+	icon: 'chevron-right',
```
idreyn, line 135: Fine to ignore this for now but we should either find a real icon for this button, or merge its behavior with the link button somehow.

client/components/FormattingBar/controlComponents/ControlsReference.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { useEffect, useState, useMemo, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '@blueprintjs/core';
+
+import { MenuButton, MenuItem } from 'components/Menu';
+import { getReferenceableNodes } from 'components/Editor/utils/references';
+
+const propTypes = {
```
idreyn, line 8: No time like the present to start making full use of typescript, if you've got the time to convert this (now or in another PR)

client/components/Menu/Menu.tsx
```diff
@@ -1,20 +1,20 @@
...
-const propTypes = {
-	'aria-label': PropTypes.string.isRequired,
-	children: PropTypes.node.isRequired,
-	className: PropTypes.string,
-	disclosure: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
-	gutter: PropTypes.number,
-	menuStyle: PropTypes.shape({}),
-	onDismiss: PropTypes.func,
-	placement: PropTypes.string,
+export type MenuProps = {
```
idreyn, line 9: Awesome, thanks!

client/containers/DashboardOverview/CollectionOverview/CollectionControls.tsx
```diff
@@ -30,22 +30,19 @@ const CollectionControls = (props: Props) => {
 			<ButtonGroup>
 				<MenuButton
 					aria-label="Set collection public or private"
-					// @ts-expect-error ts-migrate(2322) FIXME: Type 'string' is not assignable to type 'null | un... Remove this comment to see the full error message
 					buttonContent={isPublic ? 'Public' : 'Private'}
 					buttonProps={{
+						// @ts-expect-error: "lock2" not valid in TS definitions
```
idreyn, line 35: We will eventually have to change this so it accepts all the icon values provided in our own custom `components/Icon`.

client/containers/Pub/PubDocument/PubMouseEvents/PubMouseEvents.tsx
```diff
@@ -1,95 +1,116 @@
...
+	}
+
+	if (element.tagName === 'P' && element.textContent === '') {
+		return false;
+	}
+
+	return true;
+}
+
+const PubMouseEvents = (props: PubMouseEventProps) => {
```
idreyn, line 39: Thoughts on renaming this component, now that it only handles the link copying functionality?

idreyn (approved):
Looks great! Please fix the issue marked with ⚠️ but not gonna block you on that.

client/components/Editor/schemas/reference.ts
3mcd: I think it's a great idea, I'll work on that now.

client/components/FormattingBar/buttons.ts
3mcd: Yeah - I plan on wrapping that into my next PR!

client/containers/Pub/PubDocument/PubMouseEvents/PubMouseEvents.tsx
3mcd: `PubLinkController` was the best I could do. Any suggestions?

PR #1032, opened 2020-09-09
RSS improvements

### LINKED ISSUE #1028: RSS Updates

idreyn:
This PR implements the [RSS Filtering spec](https://notes.knowledgefutures.org/pub/qtf79s0u/draft), adding:

- `dc:creator` elements for every author associated with a Pub
- `enclosure` elements for downloads
- `category` elements for Collections

It also adds some useful query parameters:

- `collections` accepts a list of Collection slugs like `s1,s2,-s3` where the `-` indicates exclusion.
- `connections=false` will exclude Pubs that are children via Connections
- `published(Before|On|After)` will filter Pubs by publication date (based on Releases or `customPublishedAt`)

_Test plan:_

```
npm run test-dev server/rss
```

Also, visit a community and check out its RSS feed at `rss.xml`. Try out the query parameters listed above.

Resolves #1028

### DIFF
--- a/server/rss/queries.js
+++ b/server/rss/queries.js
@@ -71,13 +71,14 @@ const pubsIdsQuery = `
 	LIMIT :limit
 `;
 
-const normalizeDate = (datelike) => dateFormat(new Date(datelike), 'yyyy-dd-mm');
+const normalizeDate = (datelike) => dateFormat(new Date(datelike), 'yyyy-mm-dd');
 
 const parseDateFilters = ({ publishedOn, publishedAfter, publishedBefore }) => {
 	if (publishedOn) {
+		const startTime = new Date(publishedOn).valueOf();
 		return {
-			publishedAfter: normalizeDate(publishedOn),
-			publishedBefore: normalizeDate(new Date(publishedOn).valueOf() + 86400 * 1000),
+			publishedAfter: normalizeDate(startTime),
+			publishedBefore: normalizeDate(startTime + 86400 * 1000),
 		};
 	}
 	return {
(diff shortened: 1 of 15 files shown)


server/rss/queries.js
```diff
@@ -1,122 +1,257 @@
...
+				url: url,
+			},
+		},
+	};
+};
 
-		return feed.xml();
+export const getFeedItemForPub = (pubData, communityData) => {
+	const { title, description, collectionPubs, id } = pubData;
+	const formattedDownload = getFormattedDownloadUrl(pubData);
```
3mcd, line 208: Nit but I think `formattedDownloadUrl`  might be a more descriptive variable name.

server/rss/queries.js
```diff
@@ -1,122 +1,257 @@
...
+			.map(({ collection }) => collection.title),
+		custom_elements: [
+			...getAllPubContributors(pubData, false, true).map((attribution) => {
+				return {
+					'dc:creator': getContributorName(attribution),
+				};
+			}),
+			pubData.avatar && createEnclosure(pubData.avatar),
+			bestPdf && createEnclosure(bestPdf),
+			xmlExport && createEnclosure(xmlExport),
```
3mcd, line 229: Without reviewing the XML yet I assume the XML parser ignores `undefined` array elements instead of converting them to some weird or unexpected XML node?

3mcd (approved):
Looks good to me!

server/rss/queries.js
idreyn: Yes it does, but I should probably filter here anyone so this is less confusing in the future!

server/rss/queries.js
idreyn: oh yeah totally

server/rss/queries.js
3mcd: I just realized I used XML three times in the same sentence 😵

idreyn: PlushUnrulyKoalabear-size_restricted

PR #1034, opened 2020-09-12
feature: Enable removal of formatted download

### LINKED ISSUE #797: Ability to remove featured download
### Required

---

**What's your request?**
Pub managers should have the ability to remove featured downloads.

Makes it possible to correct a mistake or fall back to the auto-generated versions.

### Optional

---

idreyn:
Resolves #797 

The "formatted download" is the (typically PDF) file that we allow users to associate with their Pubs to be downloaded for offline reading. Currently there is a way to add a file to act as this download, but no way to remove it. Until now:

image

Of note in this change is that `Pub.downloads` is actually an array of files, and this change causes _all_ the contents to be overwritten when the download is removed. I think this is a vestige of a more feature-rich use of `downloads,` and can be safely ignored, since only one download is currently made available to users at this time anyway. But @isTravis please validate or correct this assumption!

 _Test plan:_
Visit the Settings for a Pub and verify that you can add, remove, and download files as formatted downloads.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -70,6 +70,7 @@
 		"react/jsx-tag-spacing": 0,
 		"react/jsx-wrap-multilines": 0,
 		"react/no-children-prop": 0,
+		"react/static-property-placement": 0,
 		"react/no-unescaped-entities": ["error", { "forbid": [">", "}"] }],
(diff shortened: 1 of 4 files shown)


3mcd (approved):
It worked!

PR #1041, opened 2020-09-18
feature: add automatic labeling of media, equations, and tables

3mcd:
This PR adds automatic labeling of audio, video, images, equations, and tables. The labels aren't customizable yet, but this code adds a hook into the labeling logic via the `blockNames` object injected via document state.

### DIFF
--- a/client/components/Editor/Editor.tsx
+++ b/client/components/Editor/Editor.tsx
@@ -13,8 +13,8 @@ import nodeViews from './views';
 
 require('./styles/base.scss');
 
-type OwnProps = {
-	blockNames: { [key: string]: string };
+export type EditorProps = {
+	blockNames?: { [key: string]: string };
 	citationManager?: any;
 	customNodes?: any;
 	customMarks?: any;
@@ -33,6 +33,7 @@ type OwnProps = {
 };
 
 const defaultProps = {
+	blockNames: {},
 	citationManager: null,
 	collaborativeOptions: {},
 	customMarks: {}, // defaults: 'em', 'strong', 'link', 'sub', 'sup', 'strike', 'code'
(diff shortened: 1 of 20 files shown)


client/components/Editor/styles/table.scss
```diff
@@ -14,20 +24,17 @@ td {
 		margin-bottom: 0;
 	}
 }
+
 /* This clunky table, tableWrapper thing is because */
```
idreyn, line 28: This comment (or at least its specific wording) appears to be out of date following these changes.

idreyn (approved):
No major concerns here, and I like the `buildLabel` function!

As I mentioned on Slack I think the approach of using a `NodeView` to render reference nodes may be more straightforward, since you can read from `editorView.editable` to determine whether links should be followable or not. This lets us avoid using a reactive attribute on every reference node to hold `isReadOnly`, and has the benefit of actually responding appropriately when the editor's, uh, editability, changes. I'll let you make the call but I do think it's a more idiomatic design.

client/components/Editor/Editor.tsx
```diff
@@ -122,6 +144,7 @@ const Editor = (props: Props) => {
 				handleClickOn: props.handleSingleClick,
 				handleDoubleClickOn: props.handleDoubleClick,
 				handleScrollToSelection: props.onScrollToSelection,
+				clipboardSerializer: wrapDomSerializer(DOMSerializer.fromSchema(schema)),
```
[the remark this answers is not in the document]
3mcd: Yep, sounds good!

client/components/Editor/styles/table.scss
3mcd: Got it, I'll update it or remove it.

PR #1046, opened 2020-09-23
update: support position: absolute within HTML page blocks

3mcd:
This small change makes the wrapper for HTML blocks `position: relative` so that we can position elements absolutely within them.


idreyn (approved):
Neat!

PR #1065, opened 2020-09-30
Reference shortcut and suggestions dropdown

3mcd:
This PR adds support for inline referencing of nodes in the editor based on a trigger character `@`. When the cursor is present within a block of text preceded by `@`, a dropdown will appear showing reference-able nodes whose labels contain the text following the trigger character, e.g:

Screen Shot 2020-09-30 at 1 39 25 PM

The state of the query within ProseMirror and React is stored in a new class called `SuggestionManager` which tracks the current highlighted index, filtered nodes, and plugin state used to replace the query with a true reference (called `params`).

Here's a gif of the shortcut and suggestions dropdown in action:

captured (1)

_Test Plan_

1. Create a Pub with multiple, reference-able nodes (e.g. media).
2. Type within an existing text paragraph, or new paragraph, starting with`@`. You should see a list of all reference-able nodes in the document.
3. Type a partial label for a node. The list of references should narrow based on the text after the `@` symbol.
4. Use the up and down arrows to navigate the menu.
5. Use either the enter key or the mouse button to select a node.
6. The shortcut and query (e.g `@fig`) should be replaced with a reference.


client/components/Editor/views/ReferenceView.ts
```diff
@@ -4,19 +4,21 @@ import { EditorView } from 'prosemirror-view';
...
 
 		this.dom.textContent = label || `(ref?)`;
 		this.dom.classList.add('reference');
 		this.dom.setAttribute('data-node-type', 'reference');
 		this.dom.setAttribute('data-target-id', targetId);
-		this.dom.setAttribute('href', `#${targetId}`);
+
+		if (!this.view.editable) {
+			this.dom.setAttribute('href', `#${targetId}`);
+		}
```
3mcd, line 21: @idreyn After removing the contentDOM property from the node view, the click-dropdown behavior of references in the (editable) editor stopped working. This seemed to work fine, but now we have anchor tags without `href`s, which somewhat defeats the purpose of moving this logic into a node view. Do you have any thoughts?

client/components/FormattingBar/controlComponents/ControlsReference.tsx
```diff
@@ -26,58 +22,33 @@ const ControlsReference = (props: ControlsReferenceProps) => {
...
-					);
-				})}
-			</MenuButton>
-		</div>
+		<ReferencesDropdown
+			references={possibleTargets}
+			selectedReference={target}
+			blockNames={blockNames}
+			onSelect={onSelect}
+		/>
```
3mcd, line 51: I broke this component out with the intention of using it for the dropdown, but then realized the suggestion dropdown didn't need `MenuButton` behavior. I left it broken out into its own component, anyways. Thoughts on merging it back in?

client/components/Icon/customIcons.tsx
```diff
@@ -417,4 +417,20 @@ export default {
...
+					fontFamily:
+						"'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Cantarell', 'Open Sans', 'Helvetica Neue', sans-serif",
+				}}
+				transform="translate(0, 16)"
+			>
+				@
+			</text>
+		),
+		viewboxDefault: 20,
+	},
```
3mcd, line 435: I couldn't figure out a way to just get an `@` rendered in the editor toolbar without using this SVG "hack".

client/components/Menu/MenuButton.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
-import { Button, IButtonProps } from '@blueprintjs/core';
+import { Button } from '@blueprintjs/core';
 
-import { Icon } from 'components';
+import Icon from 'components/Icon/Icon';
```
3mcd, line 4: This fixed our circular dependency...

client/components/Icon/customIcons.tsx
idreyn: image

We may want to straighten it out in its viewbox a little so it looks centered when the toolbar element is active. I would prefer to just create a prebaked SVG of this (and I'm happy to do so).

Thinking aloud some more, I also wonder if `@` is the best piece of iconography for this, and even whether that's the best character to use to trigger this feature, since we will eventually implement at-mentions in discussions (and maybe other Prosemirror content) and I don't know whether that will be an overloading of this feature or not. I would suggest `#` is at least as good since it leaves `@` available for social functionality, and since `#` evokes the hash at the end of a link.

idreyn (approved):
Looks and feels awesome! This is substantially good to go but I do want to tweak that icon a bit, and I have some comments inline about tying the suggestion manager more closely to React, which is a thing we could consider doing, but is not obviously something we should invest in doing right now.

client/components/Icon/customIcons.tsx
3mcd: I totally agree, @ seems kind of odd since it's conventionally used for social mentions. I'll leave the SVG viewbox alignment out for now until we figure out what icon/trigger symbol we want to use. I'm leaning against `#` for the same reasons you mentioned. How about `&`? I'd argue it could be a good contender since it's not too frequently used and gets a couple bonus points for its similar semantics in some programming languages.

Issue #1058, opened 2020-09-28
[CL] Update Community navigation editor to support both Pages and Collections in nav


idreyn:
- [ ] Update "Add Page" dropdown UI to support adding Pages and Collections
- [ ] Change underlying nav shape from `uuid[]` to `{type: 'collection' | 'page', id: uuid}[]` (or something)

PR #1066, opened 2020-09-30
dev: Prepare to refactor navbar links to support collections

idreyn:
This is the first bit of #1058, where we update Community navigation to accept both Collections and Pages. As a first step, I've added type definitions and tests to the code that generates a navbar from `community.navigation`, to get a better handle on what the possibilities are. There should be no behavioral changes at all (save for some hover CSS fixes that I snuck in).

_Test plan:_
Run the `demo` community locally and make sure the Community's header (Home, ...) and footer (RSS, Legal) navigation behave identically to that on [demo.duqduq.org](https://demo.duqduq.org). I'll also make a note to make sure this works on "base PubPub" (outside of communities) when I land on duqduq.

Run the unit tests:

```
npm run test-dev client/utils/
```


client/components/NavBar/NavBar.tsx
```diff
@@ -19,81 +23,81 @@ const defaultProps = {
...
+																		icon="share"
+																		iconSize={14}
+																	/>
+																)
+															}
+															text={subitem.title}
+														/>
+													);
+												}
+												return null;
```
3mcd, line 84: This is pretty deep indentation, at least in the GitHub diff. I wonder if we could wrap any of this in a helper function at the top-level of the component scope, or do an early return or two.

client/components/NavBar/navBar.scss
```diff
@@ -12,6 +12,17 @@
...
+			.title {
+				text-decoration: underline;
+			}
+		}
+		&:focus {
+			.title {
+				text-decoration: underline;
+				text-decoration-style: dotted;
+			}
+		}
```
3mcd, line 25: What are you thoughts on something like this? To me it reads a bit simpler:

```css
&:hover, &:focus {
  .title {
    text-decoration: underline;
  }
}
&:focus .title {
  text-decoration-style: dotted;
}
```

client/utils/navigation.ts
```diff
@@ -0,0 +1,157 @@
...
+
+type NavbarChild = {
+	title: string;
+	id: string;
+	href: string;
+	isPrivate?: true;
+	isExternal?: true;
+};
+
+export type NavbarItem = NavbarChild | { title: string; id: string; children: NavbarChild[] };
```
3mcd, line 44: All those types... it's so beautiful...

3mcd (approved):
LGTM, just a couple of comments.

client/components/NavBar/NavBar.tsx
idreyn: Sure thing — I actually did this and then reverted it because it made the diff a big mess, but the point is well-taken.

client/components/NavBar/navBar.scss
idreyn: ++

PR #1068, opened 2020-09-30
feature: Allow adding Collections to Community nav

idreyn:
More progress on #1058 

Since Collections will no longer by represented by linked Pages, we need a way to add them to the Community navigation. This PR does that, with a new "Add Collection" dropdown joining the "Add Page" one.

There are quite a few weird TypeScript casts in here to deal with the fact that a `CommunityNavigationEntry` can be a `string` right now (representing a Page ID). Once I have migrated all of those to objects, we can remove these casts, and I've left some TODOs reminding me where to do that.

Screen Shot 2020-09-30 at 4 12 20 PM

_Test plan:_
In the demo community, verify that you can add and remove Pages, Collections, links, and submenus without crashing the editor, and that any changes are persisted. Create a new Community and verify that it has a valid `navigation` field.

### DIFF
--- a/client/containers/DashboardSettings/NavBuilder.tsx
+++ b/client/containers/DashboardSettings/NavBuilder.tsx
@@ -3,7 +3,7 @@ import { DragDropContext } from 'react-beautiful-dnd';
 import { Button } from '@blueprintjs/core';
 
 import { generateHash } from 'utils/hashes';
-import { CommunityNavigationEntry } from 'client/utils/navigation';
+import { CommunityNavigationEntry, isCommunityNavigationMenu } from 'client/utils/navigation';
 
 import PageCollectionAutocomplete from './PageCollectionAutocomplete';
 import NavBuilderList from './NavBuilderList';
@@ -49,8 +49,7 @@ const NavBuilder = (props: Props) => {
 					? reorder(userSetElements, result.source.index, result.destination.index)
 					: userSetElements.map((item) => {
 							if (
-								typeof item === 'object' &&
-								'children' in item &&
+								isCommunityNavigationMenu(item) &&
 								item.id === result.destination.droppableId
 							) {
 								return {
--- a/client/containers/DashboardSettings/NavBuilder.tsx
+++ b/client/containers/DashboardSettings/NavBuilder.tsx
@@ -3,49 +3,56 @@ import { DragDropContext } from 'react-beautiful-dnd';
 import { Button } from '@blueprintjs/core';
 
 import { generateHash } from 'utils/hashes';
+import { CommunityNavigationEntry } from 'client/utils/navigation';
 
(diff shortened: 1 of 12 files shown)


client/containers/DashboardSettings/NavBuilder.tsx
```diff
@@ -3,57 +3,64 @@ import { DragDropContext } from 'react-beautiful-dnd';
...
 	const onDragEnd = (result) => {
 		if (result.destination) {
 			const nextUserElements =
 				result.destination.droppableId === 'main-list'
 					? reorder(userSetElements, result.source.index, result.destination.index)
 					: userSetElements.map((item) => {
-							if (item.id === result.destination.droppableId) {
+							if (
+								typeof item === 'object' &&
+								'children' in item &&
```
3mcd, line 53: What are your thoughts about converting some of these type checks into our own type guard functions? It might be a little more readable since we can name the function something descriptive, and they'll be reusable.

client/containers/DashboardSettings/NavBuilderList.tsx
```diff
@@ -1,26 +1,23 @@
...
-const defaultProps = {
-	items: [],
-};
-
-type Props = OwnProps & typeof defaultProps;
-
 const NavBuilderList = (props: Props) => {
-	const { id, items, updateItem, pages, newLink, removeItem } = props;
+	const { id, items = [], newLink } = props;
+	const { removeItem } = useContext(NavBuilderContext);
```
3mcd, line 20: Love the use of context!

client/containers/DashboardSettings/NavBuilderRow.tsx
```diff
@@ -1,123 +1,164 @@
...
-						value={data.title}
+						value={item.href}
 					/>
-				)}
-				{type === 'link' && (
-					<InputGroup
+				</>
+			);
+		}
+		if ('children' in item) {
```
3mcd, line 91: User defined type guards could make this a bit more readable too. Perhaps we even break out the various types of `CommunityNavigationChild` into other types, and have guards for those too. I'm a bit on the fence, though. Just thinking out loud.

3mcd (approved):
LGTM, just some small questions about type guards.

client/containers/DashboardSettings/NavBuilder.tsx
idreyn: Some of these will get simpler (`typeof item === 'object'` will eventually be implied by the type) but I think type guards are a good idea here!

client/containers/DashboardSettings/NavBuilderRow.tsx
idreyn: ++

PR #1076, opened 2020-10-07
feature: add node label editor; exclude disabled node types from labeling

3mcd:
This PR adds a new section to Pub settings that allows managers/admins to enable/disable labeling of nodes by node type, and to customize the labels of enabled node groups. Labeling is disabled by default, and custom label configuration is stored in a new column `Pubs.nodeLabels` in the database.

Disabled node groups are excluded from both the reference shortcut (currently `@`) suggestion dropdown and the reference controls menu.

_Test Plan_

1. Open the settings dashboard page for a Pub.
2. Scroll to the Block Labels section of the settings page. Labeling of all node types should be disabled/greyed out by default.
3. Enable a node group by clicking on a checkbox. Change the default label value to something unique.
4. Open the pub. Create a node of the same type of the node group you enabled, and a node of a type that remained disabled.
5. Create a reference, either by using the `@` shortcut, or the editor control bar.
6. In the dropdown, you should see **only** the node of the type you enabled in step #3. The other node should not be visible.

### DIFF
--- a/client/components/Editor/Editor.tsx
+++ b/client/components/Editor/Editor.tsx
@@ -14,11 +14,12 @@ import { collabDocPluginKey } from './plugins/collaborative';
 import { getChangeObject } from './plugins/onChange';
 import { renderStatic, buildSchema, NodeReference } from './utils';
 import nodeViews from './views';
+import { NodeLabelMap } from './types';
 
 require('./styles/base.scss');
 
 export type EditorProps = {
-	blockNames?: { [key: string]: string };
+	nodeLabels: NodeLabelMap;
 	citationManager?: any;
 	customNodes?: any;
 	customMarks?: any;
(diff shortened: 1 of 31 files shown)


gabestein:
This is awesome. Two small comments for now:

1. Ref links should just look like normal links, not be bold.
1. We're probably going to want ref links to be actual links with hrefs in them, for exporting and accessibility purposes.

gabestein:
Idea for the settings:

``You can enable automatic numbering for different types of blocks, and choose the text that will be used to reference them within the Pub. By giving two types of blocks the same name, you can group them into one ordering list (i.e., if you want both images and videos to be "figures" and share the same numbering, give them both the label "Figure".```

idreyn (approved):
Code looks great — and I pulled it and played around with the label editor and it feels very smooth, nice work! I have a couple of suggestions:

- We might consider adding `spellcheck="false"` to non-editable label elements in the ProseMirror contenteditable. Users with non-English browser dictionaries may be frustrated by the fact that the labels are marked as misspelled but cannot be directly edited, and may not have the permissions or inclination to change the labels to their own language, etc.

- An "enable/disable all" button for the node label types might be nice

This is along with the stuff via Deepak this morning about the formatting bar button for references, and the opportunity to do some user education about node references in this new Pub settings area. All of this could be handled in a separate diff, and this one lgtm.

PR #1077, opened 2020-10-07
update: Use Pub citation style preference for Pub cite

### LINKED ISSUE #1075: Collection-level DOIs + automated citations
_Before filing your enhancement, please search existing issues to make sure a similar one doesn't already exist. If one does, please leave a comment or reaction in support of the issue rather than creating a new one._

_To fill out the request, please leave the below headers (**) in tact, and fill out each section. If a section isn't applicable, leave it blank or write n/a under it. Please do not delete any template sections._

Would be great if we could include collection-level DOIs in citations. For instance, some books will not have chapter-by-chapter DOIs, but instead will have a book-level DOI assigned. Including this in the citation (and on pdf exports) would be nice. Also wondering if there's a way to automatically use the most specific DOI available, so if a chapter has a DOI, use that one, but if the DOI is at the book level, use that one. 

particularly helpful with exports and citations

---

me, and admins at HKS, Goldsmiths, and MITP

gabestein:
Resolves #1075 

- Adds "default" citation HTML to the enriched pub object and displays that in the cite menu rather than always defaulting to APA.
- Refactors how we generate both pub citations (in the cite menu) and inline citations slightly by moving the list of supported citations into a shared util consumed by both
- Updates the citation 'more options' modal to show the default citation along with the most common ones (labeling the common one as default if it exists)
- Modifies generateCiteHtml to fall back to a collection-level DOI if there is no DOI for the pub itself but there is for the collection.

_Test plan_

General cites:
1. Visit a pub, add some inline cites in various formats (doi, bibtex, etc.)
1. Make sure they display properly
1. Make sure the pub cite displays properly
1. Switch the cite type and go back to the pub
1. Make sure the cites still display, in the format selected
1. Make sure the pub cite displays

Default cite:
1. Visit any pub and, in settings, change the citation style.
1. Visit the front-end of the pub, click cite, and verify that it's using the selected citation style
1. Click 'more options' in the cite menu, make sure default shows correctly style

Doi defaults:
1. Visit a pub in a book or issue collection with a doi, but which does not itself have a doi
1. Verify that the citation lists the collection-level doi
1. Visit a pub that has a doi in a collection that also has a doi
1. Verify that the pub's citation uses the pub-level doi

### DIFF
--- a/client/containers/DashboardSettings/PubSettings/CitationChooser.tsx
+++ b/client/containers/DashboardSettings/PubSettings/CitationChooser.tsx
@@ -4,6 +4,7 @@ import { Select } from '@blueprintjs/select';
 
 import { InputField } from 'components';
 import { apiFetch } from 'client/utils/apiFetch';
+import { citationStyles as citationStyleItems } from 'utils/citations';
 
 require('./citationChooser.scss');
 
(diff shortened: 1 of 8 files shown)


idreyn (approved):
Thanks for that change, looks good to me!

Issue #1052, opened 2020-09-28
[CL] Mount Page layout editor in Collections dashboard tab


idreyn:
- [ ] Create new `layout` field on `Collection` model with JSON type and shape `{isNarrow: boolean, blocks: PageBlock[], ...otherOptions?}`
- [ ] Mount Page editor into new tab of Collection dashboard called "Layout" with same icon as Page layout
- [ ] Modify `/api/collections` to accept layout information and allow new Layout editor to persist changes to db
- [ ] Test cases
- [ ] New UI: restore default Collection layout, sort by Collection rank in Pubs block

PR #1078, opened 2020-10-07
update: Refactor client-side Layout code and add some common type definitions

idreyn:
This is probably work on #1052 but I decided to cut off a PR here before it gets too out of hand. Basically, I've moved the meat of the `Page` component into a `Layout` component that can also be consumed by a soon-to-emerge Collection container component, and generalized the `PubsBlock` component into a `BasePubsBlock` which will also be consumed by the new block type, `CollectionBlock`. In service of all of this, I created a new `utils/types.ts` file where we will host common types related to our data model, applied it in a few places, and cleaned up obsolete `@ts-expect-error` comments.

_Test plan:_
Not much here since this is pretty intermediate — if we can get a Page and a Page editor to render, that's good enough for now.


client/components/Layout/BaseLayoutPubs.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
```
idreyn, line 1: Most of this component is taken from `LayoutPubs.tsx`

client/components/Layout/Layout.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
```
idreyn, line 1: Most of this is taken from `Page.tsx`

utils/layout/util.js
```diff
@@ -0,0 +1,120 @@
+export const validBlockTypes = [
+	'pubs',
+	'text',
+	'html',
+	'banner',
+	'pages', // TODO(ian): Remove this after migration
+	'collections-pages',
+];
+
+export const getPubsByBlockIndex = (blocks, pubs) => {
```
idreyn, line 10: This is `generateRenderLists` by a more descriptive name and moved from `utils/pages.js`

client/components/PubPreview/PubPreview.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useState, useRef, useEffect } from 'react';
 import classNames from 'classnames';
 import dateFormat from 'dateformat';
 
+import { Pub } from 'utils/types';
```
3mcd, line 5: 👌

utils/types.ts
```diff
@@ -0,0 +1,175 @@
...
+	collectionPubs?: CollectionPub[];
+	exports?: Export[];
+	members?: Member[];
+	branches?: Branch[];
+	releases?: Release[];
+	pubVersions?: PubVersion[];
+	crossrefDepositRecord?: CrossrefDepositRecord;
+	inboundEdges?: InboundEdge[];
+	outboundEdges?: OutboundEdge[];
+};
```
3mcd, line 175: Love it!

3mcd (approved):
LGTM! Awesome work on all of the types!

PR #1082, opened 2020-10-13
update: toggle referencing of media, equations, and tables

3mcd:
This PR adds the ability to toggle labeling of specific nodes via the editor controls of block equations, media, and tables. Hiding the label of a node should decrement the counter on labels of subsequent, labeled nodes. Nodes with hidden labels will no longer be referenceable, and any existing references will fall back to the **(ref?)** placeholder.

I also snuck in a change to group nodes with shared custom labels into the same reactive counter state. e.g. if both images and audio are labeled "Figure", they will share the same list of potential counter values.

_Screenshots_

Standard block editor controls
Screen Shot 2020-10-14 at 10 20 13 AM

Table controls
Screen Shot 2020-10-14 at 10 24 00 AM

_Test Plan_

1. Create a Pub and enable labeling of images and tables.
2. Update the custom of images and tables to "Figure".
3. Add an image and a table to the Pub.
4. Note that it should have a label and counter by default. The image should be labeled **Figure 1** and the table should be labeled **Figure 2**.
5. Create a reference to **Figure 1** either with the trigger character `@` or editor toolbar.
6. Select the image (Figure 1) and check the Hide Label option in the editor controls.
7. Ensure the following things happen:
    - The image's label is hidden.
    - The table's label is updated to read **Figure 1**.
    - The reference you created in step 5 is reverted to the **(ref?)** placeholder.
    - The image is no longer referenceable via the trigger character or editor toolbar.

### DIFF
--- a/client/components/Editor/plugins/onChange.ts
+++ b/client/components/Editor/plugins/onChange.ts
@@ -20,6 +20,9 @@ import { getReactedCopyOfNode } from '@pubpub/prosemirror-reactive';
 
 import { collabDocPluginKey } from './collaborative';
 import { domEventsPluginKey } from './domEvents';
+import { findParentNodeClosestToPos } from '../utils';
+import { EditorView } from 'prosemirror-view';
+import { ReferenceableNodeType } from '../types';
 // import { collaborativePluginKey } from './plugins/collaborative';
 
 const getInsertFunctions = (editorView) => {
@@ -41,7 +44,33 @@ const getInsertFunctions = (editorView) => {
 		}, {});
 };
 
-const getMenuItems = (editorView) => {
+const toggleTableLabel = (editorView: EditorView, editorProps: any, dryRun = false) => {
+	if (!editorProps.nodeLabels[ReferenceableNodeType.Table].enabled) {
+		return false;
+	}
+
+	const table = findParentNodeClosestToPos(
+		editorView.state.selection.$from,
+		(node) => node.type.name === 'table',
+	);
+
+	if (table) {
+		const transaction = editorView.state.tr.setNodeMarkup(table.pos, table.node.type, {
+			...table.node.attrs,
+			hideLabel: !table.node.attrs.hideLabel,
+		});
+
+		if (!dryRun) {
+			editorView.dispatch(transaction);
+		}
+
+		return true;
+	}
+
+	return false;
+};
+
+const getMenuItems = (editorView: EditorView, editorProps) => {
 	const schema = editorView.state.schema;
 
 	/* Marks */
@@ -71,7 +100,7 @@ const getMenuItems = (editorView) => {
 			return false;
 		}
 		const $from = editorView.state.selection.$from;
-		const selectedNode = editorView.state.selection.node;
+		const selectedNode = (editorView.state.selection as any).node;
 		let isActive = false;
 		const checkForNode = (node) => {
 			const isType = type.name === node.type.name;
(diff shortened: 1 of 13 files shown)


client/components/Editor/plugins/onChange.ts
```diff
@@ -71,7 +101,7 @@ const getMenuItems = (editorView) => {
 			return false;
 		}
 		const $from = editorView.state.selection.$from;
-		const selectedNode = editorView.state.selection.node;
+		const selectedNode = (editorView.state.selection as any).node;
```
3mcd, line 104: I need to figure out the best strategy here since the prosemirror type definitions do not include a `node` property on `Selection<any>`. Any thoughts @idreyn ?

client/components/FormattingBar/controlComponents/ControlsTable.tsx
```diff
@@ -32,6 +31,7 @@ const buttonCommands = [
 	{ key: 'table-split-cell', title: 'Split cells', icon: 'split-columns' },
 	{ key: 'toggle-header-cell', title: 'Toggle header cells', icon: 'header' },
 	{ key: 'table-delete', title: 'Remove table', icon: 'trash' },
+	{ key: 'table-toggle-label', title: 'Toggle label', icon: 'tag' },
```
3mcd, line 34: Use of the `tag` icon is tentative here until we can figure out the best symbol/shortcut to use for references.

client/components/Editor/plugins/onChange.ts
idreyn: A selection of a node is a `NodeSelection` — you could cast to that type or write a type guard.

client/components/Editor/plugins/onChange.ts
```diff
@@ -41,7 +45,33 @@ const getInsertFunctions = (editorView) => {
 		}, {});
 };
 
-const getMenuItems = (editorView) => {
+const toggleTableLabel = (editorView: EditorView, editorProps: EditorProps, dryRun = false) => {
+	if (!editorProps.nodeLabels[ReferenceableNodeType.Table]?.enabled) {
```
idreyn, line 49: Is this the first use of optional chaining in our codebase? Nice!

idreyn (approved):
Looks good to me. Should we hide the label checkboxes from the node controls when labeling for that type is not enabled in Pub Settings?

3mcd:
@idreyn Yes absolutely. I'll wrap that into this PR.

Dawit-Tegbaru:
Nice!

gabestein:
> Looks good to me. Should we hide the label checkboxes from the node controls when labeling for that type is not enabled in Pub Settings?

Agree with Ian here. Otherwise, played around with it a bit and it looks great!

isTravis:
🎉 So excited for this to land!

I may have been missing a step, but it was a bit confusing for me at first (jumped into an existed pub, if that matters). I had the reference icon in the toolbar and could add a reference node, but because the default settings had no Block Labels selected, there was nothing for them to point to. 

Curious about:
1. What the defaults are for Block Labels, and 
2. If we should give a heads up when Block Labels are disabled when adding a reference node

3mcd:
Thanks for the feedback!

> 2. If we should give a heads up when all Block Labels are disabled when adding a reference node

Good call. I've been discussing this with Ian. I'll get this added in a follow-up PR.

PR #1084, opened 2020-10-14
feature: Collection Layouts tab

### LINKED ISSUE #1052: [CL] Mount Page layout editor in Collections dashboard tab

idreyn:
Resolves #1052 

At long last, this PR adds a Layout tab to the Collection dashboard, allowing users to specify how the Collection will appear to readers. A few notes:

- Only Pubs in a Collection are eligible to appear in its Layout.
- The Page editor and this tab now share a `LayoutEditor` component whose behavior is very close to identical in both places. The only difference as of now is that new Pubs blocks in a Collection layout will default to sorting by collection order. To that end, a new "Order" control has been introduced to the Pubs block editor.
- The `LinkedPageSelect` that lets you choose a Page to link a Collection to has been removed from the Collection Overview and Settings dashboard tabs, and now lives here.
- The "Visit Collection" button doesn't work yet.

Upcoming PRs will clean up the Pubs block editor so it fits its controls on two lines, and add a new `/<slug>` route to render the Collection on the community site.

_Screenshot:_
image

_Test plan:_
Visit the Layout tab for a given Collection. Verify that:
- Only Pubs in the Collection ever appear in the layout
- You can update and save changes to the layout
- Adding a new Pubs block defaults to sorting by "Collection order"
- You can disable the layout and use a linked Page instead.


gabestein:
Hi, great panel, I have a comment and a question...

No, but seriously, looks great. Two minute ordering nits:

1. Rather than creation date, I think we should probably sort by release date (and if not released, fall back to creation).
1. Would it be possible to support sorting by release date ascending and descending? I've gotten requests for both.

idreyn:
>     1. Rather than creation date, I think we should probably sort by release date (and if not released, fall back to creation).
> 
>     2. Would it be possible to support sorting by release date ascending and descending? I've gotten requests for both.

Yes to both, but I intend to handle them in a separate PR. Sorting by release date instead of creation date may change layouts on existing communities — is there any value in providing options for both?

client/components/LayoutEditor/LayoutEditorPubs.tsx
```diff
@@ -0,0 +1,331 @@
...
+			</InputField>
+		);
+	};
+
+	const renderPinnedPubs = () => {
+		return (
+			<InputField label="Pinned Pubs">
+				<Popover
+					content={
+						<OrderPicker<Pub>
```
3mcd, line 227: Gotta be honest, I've never seen this syntax before! Was TypeScript unable to infer the type argument from the props, e.g. `allItems`?

client/containers/DashboardCollectionLayout/DashboardCollectionLayout.tsx
```diff
@@ -0,0 +1,153 @@
...
+					onClick={handleSaveChanges}
+				/>
+			</React.Fragment>
+		);
+	};
+
+	const renderPageSelectLabel = () => {
+		return (
+			<>
+				Use a Page as the landing page for this Collection: &nbsp;
```
3mcd, line 91: Would `&nbsp;` have the same effect as `{' '}`? Do we prefer one over the other?

3mcd (approved):
Looks good to me!!

client/components/LayoutEditor/LayoutEditorPubs.tsx
idreyn: Actually it looks like it can infer it!

client/components/SettingsSection/SettingsSection.tsx
```diff
@@ -1,29 +1,22 @@
...
 
-const defaultProps = {
-	className: '',
-	title: '',
-	children: undefined,
-};
-
-const SettingsSection = function(props) {
+const SettingsSection = (props: Props) => {
+	const { title, className = '', children } = props;
```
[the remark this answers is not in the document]
idreyn: You're right, we don't.

client/containers/DashboardCollectionLayout/DashboardCollectionLayout.tsx
idreyn: I wish I could say why, but I only know that `&nbsp;` works here while `{' '}` is not treated as visible whitespace (maybe because it abuts an opening tag?)

Issue #1055, opened 2020-09-28
[CL] Update routing scheme to support /<slug> for Collections and Pages


idreyn:
- [x] Change route handler to select between Collection and Page and navigate appropriately
- [x] Add UI to change collection slug in Dashboard
- [x] Update the slug-related Dashboard UI and API routes to forbid slug sharing between these models (ideally update Collection API tests to enforce this)
  - Collection dashboard
  - Page dashboard
  - Collection and Page creation modals
- [x] If there are any slugs in a given Community shared by a Collection and a Page, update the Collection slug to e.g. `my-shared-slug-1` (and make a plan to notify community admins of this change if it seems bad)

PR #1087, opened 2020-10-16
update: Enforce that Collections and Pages have jointly unique slugs

idreyn:
This is related to #1055 

Here we add the following:

- UI in Collection settings to change a Collection's slug
- Checks that enforce that Collections and Pages cannot share slugs (and error states for this in the UI)
- Code that guarantees that a newly created Collection or Page will get a unique slug

There are also some changes (improvements?) to error handling that we already talked about @3mcd 

_Test plan:_
Visit the settings for both a Page and a Collection and verify that you can change their slugs, but not to a slug that another object in the Community already owns. Create a new Collection and Page and verify that they both have valid slugs.

Run tests:
```
npm run test-dev server/collections/
```

Sorry for the alarming number of commits listed below...I need to get good at rebasing again with all of these stacked PRs.


client/utils/usePersistableState.ts
```diff
@@ -0,0 +1,54 @@
+import { useState } from 'react';
+import { useBeforeUnload } from 'react-use';
+
+export const usePersistableState = <T>(
```
idreyn, line 4: This is a pattern we use everywhere, so it'll be good to have a hook for it.

3mcd: This is really cool!

server/utils/errors.ts
```diff
@@ -61,3 +88,18 @@ export const handleErrors = (req, res, next) => {
...
+		}
+	} else if (err instanceof HTTPStatusError) {
+		if (!res.headersSent) {
+			res.status(err.status).send(err.message);
+		}
+	} else if (!res.headersSent) {
+		res.status(500);
+		next(err);
+	}
+};
```
3mcd, line 105: Nice!!

3mcd (approved):
LGTM!! The error code looks great, too. I'll take your word for it that `findAcceptableSlug` is called in each of the sites where it's needed.

PR #1088, opened 2020-10-16
new: Twitter embed

gabestein:
Someone recently asked for better social embedding. It turns out Twitter has recently updated their [oembed](https://developer.twitter.com/en/docs/twitter-for-websites/embedded-tweets/overview) [capabilities](https://developer.twitter.com/en/docs/twitter-for-websites/timelines/overview) to make this quite easy, so, I went ahead and added it. Because Twitter returns "rich" oembeds (ie, not iFrames), it uses the same "hack" as Gists, encoding the html and using a data uri to render it into the iFrame. (As a sidenote, I think we should eventually use this approach – well, in fact, srcdoc because we won't be able to validate the content – to allow people to create arbitrary embeds of html code of their choosing. But that needs a bit more work).

Screen Shot 2020-10-16 at 14 58 27

_Test Plan_
1. Embed a Tweet (e.g. https://twitter.com/pubpub/status/1316132973255766017)
1. Embed a "timeline" (e.g. https://twitter.com/pubpub)

### DIFF
--- a/client/components/FormattingBar/media/MediaTwitter.tsx
+++ b/client/components/FormattingBar/media/MediaTwitter.tsx
@@ -1,0 +1,117 @@
+import React, { Component } from 'react';
+import { InputGroup, Button, Intent, NonIdealState } from '@blueprintjs/core';
+import { isHttpsUri } from 'valid-url';
+
+import Icon from 'components/Icon/Icon';
+import { apiFetch } from 'client/utils/apiFetch';
+import { getIframeSrc, getEmbedType } from 'client/utils/editor';
+
+type Props = {
+	onInsert: (...args: any[]) => any;
+};
+
+const sampleUrl =
+	process.env.NODE_ENV === 'production'
+		? 'https://twitter.com/pubpub'
+		: 'https://twitter.com/kfutures';
+
+type State = any;
+
+class MediaTwitter extends Component<Props, State> {
+	constructor(props: Props) {
+		super(props);
+		this.state = {
+			isValid: false,
+			input: '',
+			embedUrl: '',
+			embedTitle: '',
+		};
+		this.handleInput = this.handleInput.bind(this);
+		this.handleInsert = this.handleInsert.bind(this);
+	}
+
+	handleInput(url) {
+		const input = url;
+		const isValid = isHttpsUri(input) && getEmbedType(input) === 'twitter';
+		this.setState(
+			{
+				input: input,
+				isValid: isValid,
+			},
+			() => {
+				if (!this.state.isValid) {
+					return this.setState({ embedUrl: '', embedTitle: '' });
+				}
+
+				const queryParams = `?type=${getEmbedType(input)}&input=${input}`;
+				// @ts-expect-error ts-migrate(2554) FIXME: Expected 2 arguments, but got 1.
+				return apiFetch(`/api/editor/embed${queryParams}`).then((result) => {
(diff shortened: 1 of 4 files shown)


client/components/FormattingBar/media/MediaTwitter.tsx
```diff
@@ -0,0 +1,119 @@
...
+				}
+
+				const queryParams = `?type=${getEmbedType(input)}&input=${input}`;
+				// @ts-expect-error ts-migrate(2554) FIXME: Expected 2 arguments, but got 1.
+				return apiFetch(`/api/editor/embed${queryParams}`).then((result) => {
+					this.setState({
+						embedUrl: result.html,
+						embedTitle: result.title,
+					});
+				});
```
3mcd, line 53: I'm not totally sure that this need to be in `setState`'s callback function, since we already know the value of `isValid` before the component updates. This could be refactored into something like...

```typescript
const input = url;
const isValid = ...;

this.setState({ input: input, isValid: isValid });

if (!isValid) {
  // this and the above setState call will be batched into the same update
  this.setState({ embedUrl: '', embedTitle: '' });
  return;
}

const queryParams = ...

apiFetch(...).then(result => this.setState(...))

```

client/components/FormattingBar/media/MediaTwitter.tsx
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	render() {
+		return (
+			<div className="formatting-bar_media-component-content">
+				<InputGroup
+					className="top-input"
+					fill={true}
+					placeholder="Enter Twitter URL"
+					large={true}
```
3mcd, line 73: These `attr={true}` can be simplified into just `attr` just like HTML attributes.

3mcd (changes requested):
Looks good other than a couple of TypeScript suggestions and a small refactor of the `handleInput` function.

gabestein:
Thanks! New push on the way up. I should say: I copied this over from the YouTube embed file, so, this is probably homework we're going to have to do eventually on all of these. But your comments were very helpful for continuing to get up to speed on TS, so I appreciate it a lot!

client/components/FormattingBar/media/MediaTwitter.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { InputGroup, Button, Intent, NonIdealState } from '@blueprintjs/core';
+import { isHttpsUri } from 'valid-url';
+
+import Icon from 'components/Icon/Icon';
+import { apiFetch } from 'client/utils/apiFetch';
+import { getIframeSrc, getEmbedType } from 'client/utils/editor';
+
+type Props = {
+	onInsert: (...args: any[]) => any;
+};
```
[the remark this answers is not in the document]
gabestein: `onInsert` expects an argument that contains the type of node to insert. I didn't see a type definition for that yet anywhere, so I just made it a string. But if there is an enum for it you can point me to that I'm missing because I'm new to it, lmk.

client/components/FormattingBar/media/MediaTwitter.tsx
gabestein: good catch! We're gonna need to update this in all the embeds eventually.

3mcd:
Awesome, LGTM!

3mcd (approved):

PR #1096, opened 2020-10-19
update: Enable Pub submission banners in layouts

### LINKED ISSUE #1083: [CL] Update Collection layout submission banner blocks to scope created Pubs to given Collection

idreyn:
Resolves #1083 

Pretty much what it says — we need banner blocks with "New Pub" buttons to work within Collection layouts, and to automatically add the created Pub to the Collection in question.

_Test plan:_
Add a "Submission Button Banner" to a Collection layout and ensure that it allows a non-Member to create a Pub that is inserted into the Collection. Also check on the same kind of block in Page layouts to make sure it still works.


utils/fp.ts
```diff
@@ -0,0 +1,10 @@
+export const withValue = <T, R>(
```
idreyn, line 1: @3mcd I stole this from your branch because I just couldn't wait :D

3mcd: Sweet, glad its useful!!

3mcd (approved):

3mcd:
I tested it. It works!

PR #1098, opened 2020-10-20
update: references feature discoverability and conditional disabling

3mcd:
This PR improves Pub internal references with the following enhancements:

- The references button is disabled when the Pub has no block nodes enabled for labeling.
- When disabled, the button will show a tooltip that contains a link to the references section of the Pub settings.
- The "Hide label" options in node formatting controls is now disabled instead of hidden when referencing is not allowed.
- Each of the "hide label" buttons now shows the same message from the tooltip to help guide the user to Pub settings.

Tooltip:
Screen Shot 2020-10-22 at 6 36 21 PM

Disabled "Hide Label" checkbox:
Screen Shot 2020-10-22 at 6 37 23 PM

I'm not 100% happy with how the hide label checkbox text came out, but it was the most straightforward way I could come up with using Blueprint's `<Checkbox>` component. I'm looking for some feedback (paging @djagdish) on both the position and formatting of the text.

Test Plan:
TBD

### DIFF
--- a/client/components/FormattingBar/controlComponents/ControlsReference.tsx
+++ b/client/components/FormattingBar/controlComponents/ControlsReference.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState, useMemo, useRef, useCallback } from 'react';
+import classNames from 'classnames';
 
 import { ReferencesDropdown } from 'components';
 import { getReferenceableNodes, NodeReference } from 'components/Editor/utils/references';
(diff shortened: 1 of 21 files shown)


utils/dashboard.ts
```diff
@@ -26,7 +33,7 @@ export const getDashUrl = ({ collectionSlug, pubSlug, mode, subMode }: GetDashUr
 				.replace(/ /gi, '-')}`
 		: '';
 
-	return `${baseHref}${modeString}${subModeString}${baseQuery}`;
+	return `${baseHref}${modeString}${subModeString}${section && `#${section}`}${baseQuery}`;
```
idreyn, line 36: ⚠️ The `section` and `baseQuery` need to be swapped here so we don't end up with URLS lik `foo.pubpub.org/dash/blah#blah?blah`

client/components/FormattingBar/types.ts
```diff
@@ -0,0 +1,21 @@
...
+	Always,
+	Disabled,
+}
+
+export type FormattingBarPopover = {
+	component: React.ComponentType;
+	condition: FormattingBarPopoverCondition;
+};
+
+export type FormattingBarButtonData = {
```
idreyn, line 11: Nice!

client/components/FormattingBar/controlComponents/ControlsReference.tsx
```diff
@@ -53,4 +57,44 @@ const ControlsReference = (props: ControlsReferenceProps) => {
...
+			})}
+		>
+			Pub Settings
+		</a>
+	);
+
+	return (
+		<span className={classNames('controls-reference-settings-link', props.dark && 'dark')}>
+			{props.small ? (
+				<>Enable block labeling in {link}</>
```
idreyn, line 82: A bit late to this one but I wonder if we should just call this feature "item labeling"? I think the word "block" may not do much here for someone unfamiliar with the internals of our editor schema.

idreyn (changes requested):
This looks great so far (pending any of my changes + linting etc) and I love the cleanup and typing of some of our hairier formatting bar code. Requesting changes just for one point related to dash URLs.

client/components/FormattingBar/controlComponents/ControlsReference.tsx
3mcd: I like "item labeling" a lot more. I'll go ahead and make the change now with the assumption we'll get more feedback from the team before this gets merged into master.

PR #1102, opened 2020-10-21
update: New Pubs Block ordering options

idreyn:
This PR adds new `publish-date`, `publish-date-reversed`, and `creation-date-reversed` sorting options to layout Pubs blocks. I've also done some CSS tweaking to force the Pubs block options onto two lines.

_It's like Spirit Airlines with the legroom in here, folks,_
image

_Test plan:_
Check that all of the sort options for Pubs blocks work as expected. Check that newly created Pubs blocks in Pages are sorted by publish date.


3mcd (approved):
Code looks good and tested the deployment :+1:

PR #1104, opened 2020-10-22
update: Add the readingCollection URL param to Collection layout PubPreviews

idreyn:
We want to ensure that when a Pub in a Collection layout is clicked, that Pub is loaded in the context of that Collection, meaning it will be shown prominently in the Pub Header and used for Read Next. We accomplish that with the `readingCollection` URL parameter. This PR adds that param where appropriate to links in Pubs blocks in Collection layouts.

_Test plan:_
- Visit a Collection layout and ensure that the `readingCollection` param is specified on Pub hrefs as expected.
- Visit a Page layout and a User profile and ensure the PubPreviews in those views still work.

### DIFF
--- a/client/components/Layout/Layout.tsx
+++ b/client/components/Layout/Layout.tsx
@@ -28,9 +28,8 @@ const Layout = (props: Props) => {
 	const renderBlock = (block: LayoutBlock, index: number) => {
 		if (block.type === 'pubs') {
 			return (
-				<div className="layout-pubs-block">
+				<div className="layout-pubs-block" key={index}>
 					<LayoutPubs
-						key={index}
 						content={block.content}
 						pubs={pubBlocksLists[index]}
 						collectionId={collectionId}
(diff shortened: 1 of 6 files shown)


client/components/Layout/Layout.tsx
```diff
@@ -29,7 +29,12 @@ const Layout = (props: Props) => {
 		if (block.type === 'pubs') {
 			return (
 				<div className="layout-pubs-block">
-					<LayoutPubs key={index} content={block.content} pubs={pubBlocksLists[index]} />
+					<LayoutPubs
+						key={index}
```
3mcd, line 33: Does this `key` prop need to be on the parent div instead?

3mcd (approved):
LGTM other than one small thing that looks unrelated to your changes.

client/components/Layout/Layout.tsx
idreyn: oh, my god, yes! I kept getting key errors and I couldn't for the life of me figure out why because I missed the div. Thanks.

PR #1106, opened 2020-10-23
feature: Collection header block

idreyn:
This PR adds a new type of Layout block for Collections which will appear in default Collection layouts and act as a header. There's also a lot of refactoring in here:

- TypeScript definitions for attribution/contributor related utilities (we really need to reconcile these two names)
- New `ContributorAvatars` component to show a list of contributor avatars like in Pub preview blocks. This has a hover popover showing more details about each contributor, which itself was factored out of the Pub Header details pane.

Feedback on design welcome!

_A fully loaded Collection header block:_
image

_Test plan:_
- Check that Collection header blocks can be added, modified, and removed in the Collection layout editor. Check that they do not appear in the Page layout editor.
- Check these out on the reader side (and in mobile view)
- Test that new Collections have a header block
- Make sure avatars in PubPreviews are still okay (refactored to use shared `ContributorAvatars`)
- Make sure Pub Header details pane contributors list is still okay (and on mobile) (refactored to use `ContributorsListing`)


utils/ensureUserForAttribution.ts
```diff
@@ -0,0 +1,34 @@
+import { Attribution, AttributionWithUser, isAttributionWithUser } from 'utils/types';
+
+import { getPartsOfFullName } from './names';
+
+/**
+ * Attaches a `user` prop to an attribution object if one doesn't already exist, i.e. if the
+ * attributed author isn't present on PubPub.
+ * TODO(ian): I think we should move towards doing this normalization in the database.
+ */
+export default (attribution: Attribution): AttributionWithUser => {
```
idreyn, line 10: This has been moved from `ensureUserForAttribution.js`. If someone can double-check that the logic matches I'd be very appreciative...when this breaks the results are usually catastrophic.

3mcd: Logic checks out!

3mcd (approved):
LGTM!

PR #1112, opened 2020-10-28
dev: Add pubRanks to CollectionPubs

idreyn:
This is the first bit of work to remove `isPrimary` from CollectionPubs and replace it with a ranking-based scheme to determine which Collection should be used in which contexts. Just as the model has a `rank` field that determines its order with respect to its Collection, it will now also have `pubRank` which determines its importance to the Pub.

All this PR does is add the field to the db schema, updates the `createCollectionPub` helper function, and provides a data migration to backfill `pubRanks`: currently-primary CollectionPubs appear earliest in a Pub's list, and the rest are sorted by creation date.

_Test plan:_
Run the CollectionPub unit tests:
```
npm run test-dev server/collectionPub
```

### DIFF
--- a/server/collectionPub/permissions.js
+++ b/server/collectionPub/permissions.js
@@ -22,16 +22,21 @@ export const getPermissions = async ({ userId, communityId, collectionId }) => {
 	};
 };
 
+const canManagePub = async ({ userId, communityId, pubId }) => {
+	const scopeData = await getScope({ loginId: userId, communityId: communityId, pubId: pubId });
+	return (
+		scopeData.activePermissions.canManage &&
+		scopeData.elements.activePub.communityId === communityId
+	);
+};
+
 export const canCreateCollectionPub = async ({ userId, communityId, collectionId }) => {
 	const { activePermissions } = await getScope({
 		communityId: communityId,
 		collectionId: collectionId,
 		loginId: userId,
 	});
-	if (activePermissions.canManage) {
-		return true;
-	}
-	return false;
+	return activePermissions.canManage;
 };
 
 export const getUpdatableFieldsForCollectionPub = async ({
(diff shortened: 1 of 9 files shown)


tools/migrate.js
```diff
@@ -12,9 +12,11 @@ import * as models from 'server/models';
 import { promptOkay } from './utils/prompt';
 
 const {
-	argv: { name },
+	argv: { name, down },
 } = require('yargs');
 
+const fnName = down ? 'down' : 'up';
```
3mcd, line 18: This is great!

server/collectionPub/queries.js
```diff
@@ -48,10 +48,19 @@ export const getPubsInCollection = async ({ communityId, collectionId, userId })
 		);
 };
 
+const getRankInPeers = (requestedRank, ranks, moveToTop = false) => {
+	if (requestedRank) {
+		return requestedRank;
+	}
```
3mcd, line 54: Would `requestedRank` be better as the second argument in this function seeing as it's optional?

server/collectionPub/permissions.js
```diff
@@ -20,3 +21,58 @@ export const getPermissions = async ({ userId, communityId, collectionId }) => {
...
+
+export const canDestroyCollectionPub = async ({ userId, communityId, collectionPubId }) => {
+	const { collectionId } = await CollectionPub.findOne({ where: { id: collectionPubId } });
+	const { activePermissions } = await getScope({
+		communityId: communityId,
+		collectionId: collectionId,
+		loginId: userId,
+	});
+	return activePermissions.canManage;
+};
```
3mcd, line 78: I have a sneaking suspicion this code may have triggered your comment about object shorthand syntax...

3mcd (approved):
Looks good to me!

server/collectionPub/queries.js
idreyn: It's not an optional argument so much as a nullable one (if only there were some way of expressing this in code!)

PR #1129, opened 2020-11-09
fix: Restore narrow widths

gabestein:
Very quick patch to restore narrow widths. Will need to be fixed more upstream when @idreyn returns.


3mcd (approved):
LGTM! Also, I think the test failures are unrelated to your changes.

PR #1140, opened 2020-11-12
fix: dont attempt to resolve collection via /:slug route pattern

3mcd:
Quick fix to prevent the collection route handler from resolving a collection via partial ID when the pattern is `{community root}/:slug`


idreyn (approved):
thanks!

PR #1142, opened 2020-11-13
fix: always enable pointer events in pub figures

### LINKED ISSUE #1134: Interactive elements not clickable

1. Visit https://demo.pubpub.org/pub/2bqfqpud/release/2 and try to download the PDF
2. You can't click on the card
3. Same with playing the video.

The card is clickable. The video is playable.

https://demo.pubpub.org/pub/2bqfqpud/release/1

Mac / Chrome latest
---

**Have any screenshots? (Attach them by dragging the image here)**

**Who reported this? (If not you)**
Scott Jacques

3mcd:
Closes #1134 (for real this time).

### DIFF
--- a/client/components/Editor/styles/figure.scss
+++ b/client/components/Editor/styles/figure.scss
@@ -9,7 +9,6 @@ figure {
 
 	& > * {
 		width: 100%;
-		pointer-events: none;
 	}
 
 	&.ProseMirror-selectednode {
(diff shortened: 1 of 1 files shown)


gabestein:
Thanks! I think this will mostly work...

But, it's slightly different from prior functionality and I wonder if there's a way to get closer to that. Previously, in editor mode, you could click anywhere on an embedded or interactive block to select it. Once it was selected, then you could click to play, etc. Now, the first click will interact with the contents of the block. This means that it's really hard to select media that doesn't have, ie, a caption. The only way to do it is really by cursing over it.

So, I guess my question is, is there a way to only allow the pointer events on `ProseMirror-selectednode` when in editor mode? That seems like it would restore the prior functionality.

3mcd:
@gabestein I think I understand now. Yes this is possible and should be a simple change - I'll push up a fix shortly.

PR #1145, opened 2020-11-16
UI: Navbar, Header styles, Block style selector

djagdish:
**Screenshot showing updates to header styles & block selector dropdown**

image

**Style-set rendered in Figma with font info**
image

**Overflow gradient in navbar-dark**
image

**Overflow gradient in navbar-light**
image

### DIFF
--- a/client/components/FormattingBar/CommandMenu.tsx
+++ b/client/components/FormattingBar/CommandMenu.tsx
@@ -1,54 +1,39 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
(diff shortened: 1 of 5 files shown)


client/components/FormattingBar/CommandMenu.tsx
```diff
@@ -6,14 +6,14 @@ import { Menu, MenuItem, MenuItemDivider } from 'components/Menu';
 
 type commandPropType = {
 	key: string;
-	title: string;
+	title: string | React.ReactNode;
```
idreyn, line 9: FYI: this can just be `React.ReactNode`, since strings are included in that type.

idreyn (approved):
These look awesome!

PR #1146, opened 2020-11-16
fix: default pub connection view to list

3mcd:
This PR restores the previous list view default for the Connection listing at the bottom of Pubs.

_Test Plan_
1. Create a Pub with at least two Connections.
2. Ensure the "Show multiple Connections as" radio group is set to "list" by default.
3. Open the Pub and confirm that the Connection listing defaults to the list view.


tools/updatePubEdgeDefaults.js
```diff
@@ -0,0 +1,22 @@
+import { Pub } from 'server/models';
```
idreyn, line 1: I've been putting data migrations like this in the `migrations` directory. Do you think they should live in their own directory apart from the schema changes instead?

idreyn (approved):
LGTM!

tools/updatePubEdgeDefaults.js
3mcd: I've distinguished the two in the past. Perhaps we have two subdirectories like `migrations/schema` and `migrations/data`.

PR #1147, opened 2020-11-16
fix: Tables in exports and column resizing

### LINKED ISSUE #1128: Tables Not Exporting / Columns and Rows Can't Be Adjusted
I suspect this is related to the recent node ID/reference updates.

1. Create a pub with a table (e.g. https://demo.pubpub.org/pub/2bqfqpud/draft)
2. Try to adjust the column width or row height
3. Export it to any format
4. Table does not appear (in html, it is just a blank table)

Columns and rows should be adjustable. Table should be included in export.

https://demo.pubpub.org/pub/2bqfqpud/draft

idreyn:
Resolves #1128 

This PR fixes two issues related to tables in Prosemirror:

- Tables are not appearing in exports
- Table resizing doesn't work

A lot of the core of what fixes these issues is accounted for in the version bump of `prosemirror-reactive`.

_Test plan:_
- Create tables in a new Pub. Verify that they are labeled correctly and that columns can be resized. Verify that the various table commands accessible from the inline context menu work as expected.
- Export the Pub to HTML and verify that the tables appear.

### DIFF
--- a/client/components/Editor/plugins/table.ts
+++ b/client/components/Editor/plugins/table.ts
@@ -4,10 +4,12 @@ import { Plugin } from 'prosemirror-state';
 // @ts-ignore
 import { columnResizing, tableEditing, goToNextCell, TableView } from 'prosemirror-tables';
 
+import { buildLabel } from '../utils';
+
 function wrapDomSerializer(domSerializer: DOMSerializer) {
 	return Object.assign(Object.create(domSerializer), {
 		// Strip table captions when copying/pasting table elements.
-		serializeFragment(fragment, options) {
+		serializeFragment: function(fragment, options) {
 			const result = domSerializer.serializeFragment(fragment, options);
 			const tableCaptions = result.querySelectorAll('table > caption');
 
@@ -24,19 +26,29 @@ function wrapDomSerializer(domSerializer: DOMSerializer) {
  * Synchronize the reactive id attribute of default prosemirror-tables NodeView.
  */
 class PubTableView extends TableView {
-	constructor(node, whatever) {
-		super(node, whatever);
-		this.syncId(node);
+	constructor(node, ...rest) {
+		super(node, ...rest);
+		this.syncDom(node);
 	}
 
 	update(node, decorations) {
 		const shouldUpdate = super.update(node, decorations);
-		this.syncId(node);
+		this.syncDom(node);
 		return shouldUpdate;
 	}
 
-	syncId(node) {
-		((this as any) as { dom: HTMLElement }).dom.setAttribute('id', node.attrs.id);
+	syncDom(node) {
+		const { dom } = (this as any) as { dom: HTMLElement };
+		const label = buildLabel(node);
+		if (label) {
+			const table = dom.querySelector('table');
+			if (table) {
+				const caption = document.createElement('caption');
+				caption.innerHTML = label;
+				table.append(caption);
+			}
+		}
+		dom.setAttribute('id', node.attrs.id);
 	}
 }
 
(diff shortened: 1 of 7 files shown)


client/components/Editor/plugins/table.ts
```diff
@@ -4,10 +4,12 @@ import { Plugin } from 'prosemirror-state';
 // @ts-ignore
 import { columnResizing, tableEditing, goToNextCell, TableView } from 'prosemirror-tables';
 
+import { buildLabel } from '../utils';
+
 function wrapDomSerializer(domSerializer: DOMSerializer) {
 	return Object.assign(Object.create(domSerializer), {
 		// Strip table captions when copying/pasting table elements.
-		serializeFragment(fragment, options) {
+		serializeFragment: function(fragment, options) {
```
idreyn, line 12: Fixing a lint error here, though I wouldn't mind turning this rule off.

client/components/Editor/schemas/table.ts
```diff
@@ -70,7 +70,7 @@ table.toDOM = (node: ProsemirrorNode) => {
 	return pruneFalsyValues([
 		spec[0],
 		{ id: node.attrs.id },
-		withValue(buildLabel(node), (label) => ['caption', { spellcheck: 'false' }, label]),
```
idreyn, line 73: We don't need `spellcheck` here because this function never returns a `DOMOutputSpec` that's consumed in contenteditable mode.

client/components/FormattingBar/controlComponents/ControlsTable.tsx
```diff
@@ -30,8 +30,8 @@ const buttonCommands = [
 	{ key: 'table-merge-cells', title: 'Merge cells', icon: 'merge-columns' },
 	{ key: 'table-split-cell', title: 'Split cells', icon: 'split-columns' },
 	{ key: 'toggle-header-cell', title: 'Toggle header cells', icon: 'header' },
-	{ key: 'table-delete', title: 'Remove table', icon: 'trash' },
 	{ key: 'table-toggle-label', title: 'Toggle label', icon: 'tag' },
+	{ key: 'table-delete', title: 'Remove table', icon: 'trash' },
```
idreyn, line 34: I just think this should be the last one :)

client/components/Editor/plugins/table.ts
```diff
@@ -24,19 +26,29 @@ function wrapDomSerializer(domSerializer: DOMSerializer) {
...
+		const label = buildLabel(node);
+		if (label) {
+			const table = dom.querySelector('table');
+			if (table) {
+				const caption = document.createElement('caption');
+				caption.innerHTML = label;
+				table.append(caption);
+			}
+		}
+		dom.setAttribute('id', node.attrs.id);
```
3mcd, line 51: Could this code somehow create and append multiple captions to the table? Or is it thrown away/re-created each time table is modified?

3mcd (approved):
LGTM, just left a minor comment.

client/components/Editor/plugins/table.ts
idreyn: Oh, that's a good point. I'll factor it into a separate function that only gets called by the constructor.

idreyn:
ping @gabestein in case you want to give this a once-over before we deploy

gabestein:
Thanks! Did a little testing. I'm able to create a table and change cell widths. Exported fine.

One small bug, although I think this was previously an issue: when I re-import an exported table, I'm unable to change the cell widths. Instead, the cell line stays selected, even if I click away and edit something else, and the cell jumps back to its original width, and occasionally I get an error and a prompt to refresh the page.

Replication:
1. Import this (rename to .md): 
[tabletest.txt](https://github.com/pubpub/pubpub/files/5550173/tabletest.txt)
1. Try to change table width

idreyn:
> One small bug, although I think this was previously an issue: when I re-import an exported table, I'm unable to change the cell widths.

This is very strange! But my hunch is that it's not related to these changes and that we should handle it in a separate ticket.

gabestein:
> This is very strange! But my hunch is that it's not related to these changes and that we should handle it in a separate ticket.

👍

PR #1149, opened 2020-11-17
update: Better Primary Collection handling in Collection Overview

idreyn:
This PR updates the Collection Overview dashboard tab to reflect our new understanding of Primary Collections. On public/non-Tag collections, a (hollow or filled) star icon is used to indicate whether the Collection is Primary for each Pub. If it's not, you can click the star to _make_ the Collection primary for the Pub — under the hood, this moves the Collection to the top of the list of Collections for the Pub by updating `CollectionPub.pubRank` on the relationship. Phew.

This is a big UX improvement since it mostly eliminates the need to go in and manually update the Primary status for constituent Pubs when the Collection is made public — this status is derived from that information and recomputed automatically.

image

Finally, we remove a bunch of code here that manipulates the `isPrimary` field on `CollectionPub`, as it is mostly obsolete. Remaining references will be cleaned up in a subsequent PR.

_Test plan:_
Visit the Overview dashboard for a Collection and verify that:
- You can use the "star" icon control to change the Primary Collection of a Pub.
- The "label" control to the right of that still works as expected.
- The primary status of new Pubs with respect to the Collection is automatically derived as they are added — essentially, they should always become primary unless they are already in another eligible Collection.
- Making a Collection public automatically makes it primary for all eligible Pubs inside it. 
- The Primary Collection controls are not shown on Tags or private Collections.


client/containers/Pub/PubHeader/SmallHeaderButton.tsx
```diff
@@ -1,39 +1,36 @@
...
+		icon,
+		label = null,
+		labelPosition = 'left',
+		onClick = null,
+		tagName = 'button',
+	} = props;
 	return (
 		<Button
-			as={tagName}
+			as={tagName as any}
```
3mcd, line 33: Just out of curiosity, what error were you getting here?

idreyn: I don't know where to find a union of all possible HTML tag names in TS (though I assume React would export one — this is like 70% laziness and 30% confusion)

3mcd (approved):
I wonder if we could make the star button disabled when it is filled in. The filled-in star is still clickable (i.e. it transitions between hover/active states like you'd expect an enabled button to) but clicking on it does nothing. Another alternative could be to keep the button enabled when filled-in and to unset the Pub's primary collection when it is clicked.

Otherwise, the code and behavior looks good to me! I had to test locally because the PR deployment was throwing a server error, which I suspect is our "Failed to bind to port within 60 seconds" issue.

idreyn:
> I wonder if we could make the star button disabled when it is filled in. The filled-in star is still clickable (i.e. it transitions between hover/active states like you'd expect an enabled button to) but clicking on it does nothing.

Yeah, I noticed this too. I'll play around with the interactions a bit before merging, but in general it's not possible to place a tooltip on top of a disabled Blueprint button.

> Another alternative could be to keep the button enabled when filled-in and to unset the Pub's primary collection when it is clicked.

The side effect of this would implicitly be to set another Collection as primary, which I'm not sure we want.

PR #1113, opened 2020-10-29
feature: New Pub Collections management UI supporting pubRanks

idreyn:
As part of our quest to ~eliminate~ rationalize the notion of "Primary Collections", we're overhauling the component that lets you choose what Collections a Pub belongs to, in Pub Settings:

image

image

Collection are reorderable to reflect the notion that the Primary Collection is the first public, non-Tag Collection. A star icon indicates which Collection, if any, is designated as primary based on this ordering. This component uses a new `QueryListDropdown` component (factored out of `PubDropdown`) that I hope will serve all of our querying and dropping down needs admirably in the future.

This component is also designed to be rendered into the Pub Header, where everyone (understandably) seems to want it.

_Test plan:_
Visit the Settings dashboard for a Pub. Ensure that you can use the Collections control to add, reorder, and remove a Pub's Collections, and that the star icon correctly designates the Primary Collection. HARD MODE: Grant manage permissions to a user for a single Pub and one or more Collections within a Community and verify that the user can see and reorder, but not add or remove, Collections that they don't have permissions to manage.

PR #1151, opened 2020-11-18
feature: Add new PubCollectionListing to Pub header

idreyn:
At long last we're bringing the ability to add Pubs to a Collection from the Pub header, using the `PubCollectionListing` component from #1113. That's essentially what this PR does, but it touches a few more files than might seem necessary at first glance to do some TS cleanup and add the `renderDragElementInPortal: boolean` prop to `DragDropListing` — this is necessary to prevent dragged items from appearing offset when the `DragDropListing`'s parent makes use of CSS `transform: translate(...)`, which is...lots of our dialogs and popovers. Love the DOM! 

_Screenshots:_
When there are no Collections, we just show the `QueryListDropdown` control:
image

If there are some Collections in place already, we show the entire `PubCollectionListing`:
image

With Collections in place we show a "+" button without any text. Here it is playing nice with the `OverflowList` tasked with fitting all of these buttons into a single row:
image

_Test plan:_
- On a Pub you can manage, ensure you can use this new control to add, remove, and reorder Collections. Check that the `CollectionBrowser` dropdown, the leftmost of the top Collection buttons, updates as the Primary Collection does.
- On a Pub that you cannot manage, make sure you don't see this control.


client/containers/Pub/PubSyncManager.tsx
```diff
@@ -112,6 +116,8 @@ const idleStateUpdater = (boundSetState, timeout = 50) => {
 
 type State = any;
 class PubSyncManager extends React.Component<Props, State> {
+	idleStateUpdater: ReturnType<typeof idleStateUpdater>;
```
3mcd, line 119: I see you've used `typeof x` in a couple places here and it made me wonder if you have any thoughts on explicitly typing with a type alias or interface vs deriving types with `typeof`. I've tried to avoid using `typeof` to derive types in the past (only using it when necessary, like if a library doesn't export an alias or interface), but I wonder if it's really worth the effort to provision an interface explicitly it if the subject is unlikely to be used polymorphically (is that a word?).

3mcd (approved):
LGTM! :)

client/containers/Pub/PubSyncManager.tsx
idreyn: I really like using `ReturnType<typeof fn>`. The other usage in this file I think is more out of laziness. But I have a sense that we're going to pick apart and completely re-implement this file, which is getting way too big for its britches, in the next six months or so...so I'm being a little loose about these things for now.

PR #1153, opened 2020-11-19
fix: Only show a given Pub once in a PubEdgeListing

idreyn:
Folks are using Pub Connections to build some real complicated "I'm my Own Grandpa" kinds of non-tree graphs between their Pubs, with the result that Pubs can appear more than once in the `PubEdgeListing` of a Pub as a parent, child, or sibling by multiple different parents. No matter what, we really want to make sure that every Pub only appears once in this listing — so that's what this PR does.

_Test plan:_
I ran 

```
PUBPUB_LOCAL_COMMUNITY=hdsr PUBPUB_PRODUCTION=true npm start
```

and then visited [this Pub](http://localhost:9876/pub/cqncbp3l/release/3), verifying that there are no duplicates in its connections listing. I'll probably want to check about a dozen other Pubs, too.

### DIFF
--- a/client/components/PubEdge/PubEdge.tsx
+++ b/client/components/PubEdge/PubEdge.tsx
@@ -3,12 +3,8 @@ import classNames from 'classnames';
 
 import { Byline } from 'components';
 import { usePageContext } from 'utils/hooks';
-import { formatDate } from 'utils/dates';
-import { pubUrl, pubShortUrl } from 'utils/canonicalUrls';
-import { getPubPublishedDate } from 'utils/pub/pubDates';
-import { getAllPubContributors } from 'utils/contributors';
+import { getHostnameForUrl, getValuesFromPubEdge } from 'utils/pubEdge';
 
-import { getHostnameForUrl } from './util';
 import PubEdgeLayout from './PubEdgeLayout';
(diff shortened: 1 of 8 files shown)


client/components/PubEdge/PubEdge.tsx
```diff
@@ -22,53 +18,6 @@ export type PubEdgeProps = {
 	showDescriptionByDefault?: boolean;
 };
 
-const getUrlForPub = (pubData, communityData) => {
```
idreyn, line 25: Moved into `helpers.ts`

client/components/PubEdgeListing/PubEdgeListingCard.tsx
```diff
@@ -25,19 +25,6 @@ export type PubEdgeListingCardProps = {
 	viewingFromSibling?: boolean;
 };
 
-const getIsViewingFromTarget = (pubEdge, viewingFromSibling, isInboundEdge) => {
```
idreyn, line 28: Moved into `helpers.ts`

3mcd:
"I'm my Own Grandpa" brought this movie to mind:
movieposter

3mcd (approved):

PR #1154, opened 2020-11-19
dev: Type improvements, types for pubDates, copyright year

idreyn:
This is a bit of a mishmash of things that I got into while eliminating `isPrimary` checks, finding some copypasted code in two different places that should be in a utility function, adding types to the `pubDates` module, and then getting sidetracked by the need for a few new utility types. Specifically we have:

- `Maybe<X>` is either `X` or a falsy value
- `Some<X>` unwraps `Maybe<X>` back into `X`
- `DefinitelyHas<X, K>` is the object type `X` but with nullable keys in `K` made non-nullable. For instance, `DefinitelyHas<Pub, 'releases' | 'exports'>` is a `Pub` that definitely has its `releases` and `exports` members — which are nullable, because many `Pub` views _won't_ have these. I'm actively looking for a shorter and better name for this utility type.

_Test plan:_
There should be no behavioral changes here (though I seem to have fixed a bug in copyright year calculation). To verify that things work, I set `premiumLicenseFlag = true` on the demo Community and then verified that I could set the license to "copyright" and see a copyright being displayed without WSOD.


client/components/LicenseSelect/LicenseSelect.tsx
```diff
@@ -33,25 +31,14 @@ const LicenseSelect = (props: Props) => {
...
-
-	const currentLicense = getLicenseBySlug(pubData.licenseSlug);
-	// @ts-expect-error ts-migrate(2532) FIXME: Object is possibly 'undefined'.
-	const primaryCollectionPub = pubData.collectionPubs.find((cp) => cp.isPrimary);
-	const collectionPubDate = primaryCollectionPub
-		? primaryCollectionPub.collection.metadata.copyrightYear ||
-		  primaryCollectionPub.collection.metadata.date ||
-		  primaryCollectionPub.collection.metadata.publicationDate
-		: null;
-	const pubCopyrightDate =
```
idreyn, line 45: This fallthrough will never occur because `dateFormat(null)` returns a string representing `Date.now()`. As a result we never actually show the correct Pub publication date.

client/components/Layout/Layout.tsx
```diff
@@ -18,7 +18,7 @@ require('./layout.scss');
 type Props = LayoutOptions & {
 	blocks: LayoutBlock[];
 	id?: string;
-	pubs: Pub[];
+	pubs: DefinitelyHas<Pub, 'releases'>[];
```
3mcd, line 21: What are your thoughts on creating a type like `PubWithReleases` that we can import instead of having to write this type each time?

client/components/Layout/LayoutPubs.tsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
 
 import { PubPreview } from 'components';
-import { Pub } from 'utils/types';
+import { Pub as BarePub } from 'utils/types';
 import { LayoutBlockPubs } from 'utils/layout/types';
 import { createReadingParamUrl } from 'client/utils/collections';
 
+type Pub = DefinitelyHas<BarePub, 'releases'>;
```
3mcd, line 8: The name `BarePub` kind of confused me at first. Do you think it'd be better to keep `Pub` and call `DefinitelyHas<Pub, 'releases'>` something like `PubWithReleases` or `ReleasedPub`?

client/containers/Pub/PubHeader/TitleGroup.tsx
```diff
@@ -4,14 +4,14 @@ import { PubByline, DialogLauncher, PubAttributionDialog } from 'components';
...
 
 import BylineEditButton from './BylineEditButton';
 import EditableHeaderText from './EditableHeaderText';
 
 type Props = {
-	pubData: {
-		title?: string;
-		description?: string;
+	pubData: Pub & {
+		releases: Release[];
```
3mcd, line 14: Could we use `DefinitelyHas` here?

types/global.d.ts
```diff
@@ -0,0 +1,4 @@
+declare type Falsy = false | null | undefined | '' | 0;
+declare type Maybe<X> = X extends Falsy ? never : X | Falsy;
+declare type Some<X> = X extends Falsy ? never : X;
+declare type DefinitelyHas<X extends {}, Keys> = X & { [k in keyof X & Keys]: Some<X[k]> };
```
3mcd, line 4: I thought for 5-10 minutes and couldn't come up with anything better. `Includes` is more concise but TypeScript already has an `Exclude` type which is meant to work with unions, so that might be confusing.

3mcd (approved):
Left a few comments but otherwise looks great.

client/components/Layout/Layout.tsx
idreyn: I don't know yet. I'd actually almost prefer to make `releases` non-nullable on `Pub` since it seems like the kind of thing you need practically every time you query it — thoughts?

3mcd: Yep, I agree it should probably be non-nullable/eagerly loaded as long as the Release objects aren't too hefty, which they don't appear to be.

PR #1155, opened 2020-11-19
update: Remove remaining references to CollectionPub.isPrimary

idreyn:
It's been quite a journey but we're finally here!

_Test plan:_
Run `npm run test` to check everything. There's some decent coverage of the API but we should probably also do a manual QA pass before deploying this one.


3mcd (approved):
Looks good, great work man!

PR #1156, opened 2020-11-20
dev(ts-migrate): move config.js to repo root (1/3)

3mcd:
This PR moves the config.js file to the repo root. The `server` directory will now contain only the TypeScript source code of the API application, and the config file will remain un-transpiled at the root of the project.

_Test Plan_
* Copy your existing local config.js file into the repo root.
* Start PubPub and make sure the app loads and functions properly!


tools/index.js
```diff
@@ -5,7 +5,7 @@
 
 /* eslint-disable global-require */
 if (process.env.NODE_ENV !== 'production') {
-	require('server/config');
+	require('config');
```
idreyn, line 8: Does this resolve?

idreyn (approved):
LGTM!

tools/index.js
3mcd: Good catch, thanks!

PR #1158, opened 2020-11-20
dev(ts-migrate): migrate server javascript to typescript (2/3)

3mcd:
This PR executes the following `ts-migrate` commands against the `server` directory:

1. `rename`: renames the extension of all JavaScript files to `.ts`
2. `migrate`: migrates the files from JS to TS (note: some prop-type transformation was done in the client directories since the server imports client components for SSR).
3. `reignore`: strips all existing `@ts-ignore` and `@ts-expect-error` comments and re-adds them with updated error information or removes them if the error no longer exists.

Any additional relevant changes are tagged with a comment and are visible below.

_Test Plan_
- `npm i && npm start` and verify that PubPub works!


client/components/Menu/Menu.tsx
```diff
@@ -3,8 +3,8 @@ import classNames from 'classnames';
 import { Classes } from '@blueprintjs/core';
 import * as RK from 'reakit/Menu';
 
-import { MenuContext, MenuConfigContext } from './menuContexts';
 import { Placement } from 'popper.js';
+import { MenuContext, MenuConfigContext } from './menuContexts';
```
3mcd, line 7: Interesting, it appears the tool organized imports a bit better (but not quite 100% our convention). Looks like an improvement though.

init.js
```diff
@@ -13,7 +13,7 @@ if (process.env.NODE_ENV === 'production') {
 
 throng({ workers: 1, lifetime: Infinity }, () => {
 	const loadServer = () => {
-		return require('./server/server.js').startServer();
+		return require('./server/server').startServer();
```
3mcd, line 16: Relevant change. Note we're still executing the file directly from `server/` since this branch is still using `@babel/register`.

server/models.ts
```diff
@@ -0,0 +1,116 @@
...
+/* Create standard id type for our database */
+sequelize.idType = {
+	primaryKey: true,
+	type: Sequelize.UUID,
+	defaultValue: Sequelize.UUIDV4,
+};
+
+/* Import and create all models. */
+/* Also export them to make them available to other modules */
+export const Branch = sequelize.import('./branch/model');
```
3mcd, line 39: The main change in this file is that the `.js` extensions were stripped from the `sequlize.import` paths.

server/models.ts
```diff
@@ -0,0 +1,116 @@
...
+			? // @ts-expect-error ts-migrate(2569) FIXME: Type 'Set<any>' is not an array type or a string t... Remove this comment to see the full error message
+			  [...new Set([...attributesPublicUser, ...providedAttrs])]
+			: attributesPublicUser;
+		return {
+			model: User,
+			attributes: attributes,
+			...restOptions,
+		};
+	};
+})();
```
3mcd, line 107: Relevant change. `includeUserModel` was moved into this file due to the odd error I was getting.

idreyn (approved):
I really only skimmed this to get a feel for the kind of changes to expect, but based on what you've explained to me about how this was generated, I expect it to work well!

PR #1159, opened 2020-11-20
fix: Don't use activeCommunity where communityData should be used

idreyn:
`getScope()` currently takes either `communityId` or `communityData` as an input to determine what `scopeData.elements.activeCommunity` is. If the ID is passed, the Community model is fetched; if `communityData` is passed, that object simply becomes `scopeData.elements.activeCommunity`. This presents a problem in `getInitialData`, where we:

1. Load a Community object with all of its constituent pages, collections, etc.
2. Call `getScope` with that Community as an input (as an optimization to avoid fetching it twice)
3. Use the resulting `scopeData.memberData` to call `sanitizeCommunity` on the Community object, eliminating objects that should be invisible to the client.

The problem is that `scopeData.elements.activeCommunity` is _not_ sanitized, and therefore may leak information about private Pages and Collections to the client. The solution is to eliminate the `communityData` argument to `getScope`, and to modify the client to look at the sanitized `initialData.communityData` instead of `initialData.scopeData.elements.activeCommunity` — which is now a bare Community object without any associated models — for things like Pages and Collections.

_Test plan:_
I spot-checked Community and Collection overview and, Collection settings, the Pub header Collections bar, etc. to check interactions that formerly depended on `scopeData.elements.activeCommunity`, making sure they still work.


3mcd (approved):
Tested locally, LGTM.

PR #1162, opened 2020-11-23
feature: Full-resolution image support (including .gif and .webp)

### LINKED ISSUE #1161: Support gifs somehow
We currently do not support gifs, because the resizer doesn't support gifs (I believe this also applies to SVGs, which we also claim to support!)

This is...a big hole, and one we should fix. In particular, HDSR is looking to add some gifs to an upcoming article. It looks like there may be a workaround for newer versions of the resizer (https://github.com/awslabs/serverless-image-handler/issues/153), but that would require updating it, which might be a whole can of worms. If there's any other way to work around it in the short term (parsing the file extension in the figure, maybe, and loading from assets instead of resizer?), we should.

If that workaround seems feasible, lmk and I can implement it next week.

Publish Pubs with gifs.

---

idreyn:
Resolves #1161 

This PR adds (restores?) support for gif and webp images by not passing them through the resizer. We also add a new control for other resizable image formats (jpg, png, did I miss any?) to allow them to skip the resizer as well, so that anyone who really wants a high-resolution image in their Pub can add one.

_New "full resolution" checkbox in `ControlsMedia`:_
image

_Test plan:_
- Add .gif and .webp files to a Pub and verify that they load correctly and the "full resolution" checkbox is hidden for them. 
- Add a .jpg and verify that it loads and that you can choose whether to resize it or not with the checkbox.
- Add a video and verify that the `ControlsMedia` works for it and the checkbox is hidden
- Check an existing image on an older Pub and verify that it loads.

[Here is a Pub](http://localhost:9876/pub/j9oul530/draft) with different image formats under test.

### DIFF
--- a/client/components/Editor/schemas/image.ts
+++ b/client/components/Editor/schemas/image.ts
@@ -1,6 +1,7 @@
 import { DOMOutputSpec } from 'prosemirror-model';
 import { pruneFalsyValues } from 'utils/arrays';
 import { withValue } from 'utils/fp';
+import { imageCanBeResized } from '../utils/media';
 import { buildLabel } from '../utils/references';
 import { renderHtmlChildren } from '../utils/renderHtml';
 import { counter } from './reactive/counter';
@@ -17,6 +18,7 @@ export default {
 			align: { default: 'center' },
 			caption: { default: '' },
 			hideLabel: { default: false },
+			fullResolution: { default: false },
 		},
 		reactiveAttrs: {
 			count: counter({ useNodeLabels: true }),
@@ -39,9 +41,12 @@ export default {
 				},
 			},
 		],
-		// @ts-expect-error ts-migrate(2525) FIXME: Initializer provides no value for this binding ele... Remove this comment to see the full error message
-		toDOM: (node, { isReact } = {}) => {
+		toDOM: (node, { isReact } = { isReact: false }) => {
 			const resizeFunc = node.type.spec.defaultOptions.onResizeUrl;
+			const maybeResizedSrc =
+				resizeFunc && !node.attrs.fullResolution && imageCanBeResized(node.attrs.url)
+					? resizeFunc(node.attrs.url, node.attrs.align)
+					: node.attrs.url;
 
 			return [
 				'figure',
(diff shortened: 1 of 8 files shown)


gabestein:
:( I'm getting an issue with the deployment...

idreyn:
> :( I'm getting an issue with the deployment...

Config vars were out of date — have a go now?

gabestein:
- [ ] Add .gif and .webp files to a Pub and verify that they load correctly and the "full resolution" checkbox is hidden for them.

Looks great! Normal webp work fine through the resizer, and animated webp don't have the resize checkbox. Awesome!

- [ ] Add a .jpg and verify that it loads and that you can choose whether to resize it or not with the checkbox.
- [ ] Add a video and verify that the ControlsMedia works for it and the checkbox is hidden
- [ ] Check an existing image on an older Pub and verify that it loads.
Had a hard time doing this one since there are no old pubs with images on this qubqub instance.

One bug I discovered: it looks like SVGs are failing to upload. I think this may have always been the case? But if so we shouldn't say we support SVGs.

gabestein (commented):
- [x] Add .gif and .webp files to a Pub and verify that they load correctly and the "full resolution" checkbox is hidden for them.

Looks great! Normal webp work fine through the resizer, and animated webp don't have the resize checkbox. Awesome!

- [x] Add a .jpg and verify that it loads and that you can choose whether to resize it or not with the checkbox.
- [x] Add a video and verify that the ControlsMedia works for it and the checkbox is hidden
- [x] Check an existing image on an older Pub and verify that it loads.
Had a hard time doing this one since there are no old pubs with images on this qubqub instance.

One bug I discovered: it looks like SVGs are failing to upload. I think this may have always been the case? But if so we shouldn't say we support SVGs.

idreyn:
> One bug I discovered: it looks like SVGs are failing to upload. I think this may have always been the case? But if so we shouldn't say we support SVGs.

Oh, thanks, just took care of that now.

gabestein:
LGTM! I also found a pub with an older release that had an image and appeared fine. So, approved from my end.

3mcd (approved):
Looks great!

PR #1164, opened 2020-11-23
dev(ts-migrate): remove babel (3/3)

3mcd:
This PR fully removes babel from our test and build systems.

- The `dist` directory now contains two child build directories, `client` and `server`.
- All server code was migrated to TypeScript.
- Babel extensions for jest (babel-jest) and Webpack (babel-loader) were swapped out for their TS counterparts (ts-jest and ts-loader, respectively).
- The NPM scripts were updated to automatically rebuild the server when changed. This is sped up significantly with incremental compilation.
- The workers were also migrated to TypeScript due to their coupled nature with the API server. The main worker processis executed with `ts-node` and thew workers it spawns are automatically run with `ts-node/register`. Watching/reloading in dev mode was tested as well as the following task-related items:
  * Verified HTML export SCSS->CSS compilation still works.
  * Tested a handful of export templates like LaTeX and JATS XML.

_Test Plan_
We will likely want to do a fairly comprehensive smoke test of the application before merging this into `master`, but for now, I'd suggest:
- Make sure to `npm install`.
- Verify the app launches and automatic reloading still works in dev mode.
- Verify the prod builds work.
- Verify imports and exports work.
- Ensure all tests pass.
- Launch Storybook and make sure nothing looks too suspicious.


client/webpack/webpackConfig.prod.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
 	devtool: '#source-map',
 	output: {
 		filename: '[name].[chunkhash].js',
-		path: resolve(__dirname, '../../dist'),
+		path: resolve(__dirname, '../../dist/client'),
```
3mcd, line 28: Client bundle is now output in the `dist/client` directory instead of `/dist`.

workers/queue.ts
```diff
@@ -13,6 +13,36 @@ const maxWorkerTimeSeconds = 120;
...
+		`
+					const wk = require('worker_threads');
+					require('ts-node').register();
+					let file = wk.workerData.__filename;
+					delete wk.workerData.__filename;
+					require(file);
+			`,
+		workerOptions,
+	);
+};
```
3mcd, line 44: **Important change**  — workers are now spawned using `eval` which allows us to inject a call to `require('ts-node').register()` prior to importing the task module.

client/webpack/webpackConfig.dev.js
```diff
@@ -46,7 +46,7 @@ module.exports = {
 			{
 				test: /\.(js|jsx|ts|tsx)$/,
 				include: [resolve(__dirname, '../'), resolve(__dirname, '../../utils')],
-				use: 'babel-loader',
+				use: 'ts-loader',
```
3mcd, line 49: **Important change** — output files into `dist/client` directory and use `ts-loader` to transform TypeScript source. The same change was made for the prod webpack config below.

client/containers/Pub/PubSyncManager.tsx
```diff
@@ -160,8 +159,11 @@ class PubSyncManager extends React.Component<Props, State> {
 	componentDidMount() {
 		const rootKey = `pub-${this.props.pubData.id}`;
 		const branchKey = `branch-${this.props.pubData.activeBranch.id}`;
-		// @ts-expect-error ts-migrate(2569) FIXME: Type 'Reference[] | null' is not an array type or ... Remove this comment to see the full error message
-		initFirebase(rootKey, this.props.pubData.firebaseToken).then(([rootRef, connectionRef]) => {
+		initFirebase(rootKey, this.props.pubData.firebaseToken).then((firebaseRefs) => {
+			if (!firebaseRefs) {
+				return;
+			}
+			const [rootRef, connectionRef] = firebaseRefs;
```
3mcd, line 166: I was able to remove a TS error by inserting an early return here. I didn't do too much digging into the types to figure out why `firebaseRefs` would be null or undefined. @idreyn is this okay for now or should we throw an error instead of failing silently?

.storybook/main.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+	stories: ['../client/**/*.stories.@(tsx|jsx)'],
+	addons: ['@storybook/addon-essentials', '@storybook/addon-viewport', '@storybook/addon-knobs'],
+};
```
3mcd, line 4: **Important change** — Storybook v6 uses a main.js file to configure story paths and addons.

.storybook/preview.js
```diff
@@ -32,5 +25,3 @@ addParameters({
 		showPanel: false,
 	},
 });
-
-configure(loadStories, module);
```
3mcd, line 36: In Storybook v6, `preview.js` is executed for each story, so I moved relevant parts of `config.js` to this file.

.jest/setup-env.js
```diff
@@ -22,40 +22,40 @@ process.env.ALGOLIA_SEARCH_KEY = 'ooo';
...
 }
 
 if (typeof window !== 'undefined') {
-    window.requestIdleCallback = () => {};
+	window.requestIdleCallback = () => {};
 }
 
-global.fetch = () => Promise.resolve();
\ No newline at end of file
+global.fetch = () => Promise.resolve();
```
3mcd, line 61: **Important change** — we overwrite `getSelection`/`createRange` regardless if mocks for those functions exist already or not.

.jest/register-context.js
```diff
@@ -1,8 +0,0 @@
-/**
- * This file is used to setup the require-context-hook plugin, which is used to
- * polyfill the Webpack-specific require.context() utility in non-Webpack
- * environments (e.g. Node, where Jest is running). We use require.context()
- * when compiling a list of all Storybook entries to test in Jest.
- */
-import registerRequireContextHook from 'babel-plugin-require-context-hook/register';
-registerRequireContextHook();
```
3mcd, line 8: **Important change** — I _think_ this code was added for storyshots. The require context hook is no longer necessary after upgrading to Storybook v6.

jest.config.js
```diff
@@ -1,6 +1,7 @@
 const esModules = ['react-stylable-diff'].join('|');
 
 module.exports = {
+	preset: 'ts-jest/presets/js-with-ts',
```
3mcd, line 4: **Important change** — we're using the `js-with-ts` preset to support interop between TS and ESM modules when running tests (since many utils are still written in JS).

init.js
```diff
@@ -13,7 +12,7 @@ if (process.env.NODE_ENV === 'production') {
 
 throng({ workers: 1, lifetime: Infinity }, () => {
 	const loadServer = () => {
-		return require('./server/server').startServer();
+		return require('./dist/server/server/server').startServer();
```
3mcd, line 15: **Important change** — there are many path rewrites, but I just wanted to point out how silly this one is :joy:.

package.json
```diff
@@ -10,11 +10,14 @@
 	},
 	"main": "init.js",
 	"scripts": {
-		"api-dev": "NODE_PATH=./client:./ node init.js --watch ./server --watch ./utils --watch ./client --watch ./node_modules/@pubpub",
-		"api-prod": "NODE_PATH=./client:./ node init.js",
+		"api-dev": "npm run watch-server & NODE_PATH=./dist/server/client:./dist/server:./dist node init.js --watch ./dist/server --watch ./node_modules/@pubpub",
```
3mcd, line 13: **Important change** — `NODE_PATH`s were updated to point to the new dist directories. Automatic reloading of the server still works as well.

package.json
```diff
@@ -10,11 +10,14 @@
...
+		"api-prod": "NODE_PATH=./dist/server/client:./dist/server:./dist node init.js",
 		"build-dev": "webpack --config ./client/webpack/webpackConfig.dev.js --watch",
-		"build-dev-once": "rm -rf ./dist && webpack --config ./client/webpack/webpackConfig.dev.js",
-		"build-prod": "rm -rf ./dist && webpack --config ./client/webpack/webpackConfig.prod.js",
+		"build-dev-once": "rm -rf ./dist/client && webpack --config ./client/webpack/webpackConfig.dev.js && npm run build-server",
```
3mcd, line 16: **Important change** — we only nuke the `dist/client` instead of the entire `dist` folder so the TS build manifest isn't deleted every build (to take advantage of incremental builds).

package.json
```diff
@@ -10,11 +10,14 @@
...
+		"build-prod": "rm -rf ./dist && webpack --config ./client/webpack/webpackConfig.prod.js && npm run build-server",
+		"build-server": "tsc -p server/tsconfig.json",
+		"watch-server": "npm run build-server -- -w",
+		"postbuild-server": "cpy '**/*' '!**/*.ts'  '!**/*.tsx' '!**/*.js' '!tsconfig.json' ../dist/server/server/ --cwd=server/ --parents",
```
3mcd, line 20: **Important change** — This script is run automatically after `build-server`. It copies all non-source, non-config assets into the dist/server directory recursively.

package.json
```diff
@@ -23,43 +26,30 @@
 		"postinstall": "npm run build-prod",
 		"start": "npm run build-dev-once && concurrently --kill-others \"npm run api-dev\" \"npm run build-dev\"",
 		"storybook": "start-storybook -p 9001 -c .storybook -s ./",
-		"test": "npm run lint && npm run check && jest --forceExit --silent",
+		"test": "npm run lint && jest --forceExit --silent",
 		"test-dev": "jest --forceExit",
-		"tools": "NODE_PATH=./client:./ node -r esm --max-old-space-size=12000 ./tools",
-		"workers-dev": "SEQUELIZE_MAX_CONNECTIONS=1 NODE_PATH=./client:./ WORKER=true nodemon workers/init.js --watch workers",
-		"workers-prod": "SEQUELIZE_MAX_CONNECTIONS=1 NODE_PATH=./client:./ WORKER=true node workers/init.js"
+		"tools": "NODE_PATH=./client:./ node -r ts-node/register -r esm --max-old-space-size=12000 ./tools",
```
3mcd, line 31: **Important change** — Tools are now run with `ts-node` since many tools import TS code from the server (namely the sequelize models).

package.json
```diff
@@ -23,43 +26,30 @@
...
+		"workers-dev": "SEQUELIZE_MAX_CONNECTIONS=1 NODE_PATH=./client:./ WORKER=true nodemon --exec 'ts-node' workers/init --watch workers -e js,ts,tsx",
+		"workers-prod": "SEQUELIZE_MAX_CONNECTIONS=1 NODE_PATH=./client:./ WORKER=true node -r ts-node/register workers/init"
```
3mcd, line 33: **Important change** — workers are also now run (in dev and prod) with ts-node. Eventually these could also be compiled into the `dist` directory but it felt less important since the worker's module graph is smaller than that of the server's.

client/containers/Pub/PubSyncManager.tsx
idreyn: `initFirebase` will return `null` if there's an authentication error. My guess is that in either case the failure is equivalently silent to the user, and they'll just see "Connecting..." in the status bar.

client/webpack/webpackConfig.dev.js
idreyn: see ya, Babel!

init.js
idreyn: 😳

server/server.ts
```diff
@@ -102,12 +103,12 @@ passport.deserializeUser(User.deserializeUser());
 /* ---------------- */
 /* Server Endpoints */
 /* ---------------- */
-app.use('/dist', [cors(), express.static('dist')]);
-app.use('/static', express.static('static'));
-app.use('/service-worker.js', express.static('static/service-worker.js'));
-app.use('/favicon.png', express.static('static/favicon.png'));
-app.use('/favicon.ico', express.static('static/favicon.png'));
-app.use('/robots.txt', express.static('static/robots.txt'));
+app.use('/dist', [cors(), express.static(path.join(process.cwd(), 'dist/client'))]);
```
idreyn, line 106: feels like a tiny helper function to create relative path from `cwd` would be helpful here

idreyn (approved):
Looks good, and jeez, what a lot of work! Thanks for taking this on...every LOC in the Github stats that you get for this is well-deserved ;)

PR #1176, opened 2020-12-02
dev: Add a historyKey to every Release

idreyn:
In order to eliminate Merges and treat Releases as simple pointers into a Pub draft, we need to be very sure which key of the draft a given Release points to. Our current understanding of that is surprisingly shaky — we have some Releases that don't correspond to a draft, some missing `branchKey` or `sourceBranchKey` fields that hint at the source of the Release, and some that appear to be duplicates of each other. In practice, when we load `/pub/:slug/release/:x` we just grab the `x-1`th merge without any regard for what Release it "actually" corresponds to, which gets us into trouble sometimes because database Release objects and Firebase merges aren't created atomically with each other during the release process, and sometimes we end up with one but not the other.

This PR contains:
- DB schema migrations to add `Release.historyKey`
- A backfill to assign a `historyKey` to every `Release`

_Test plan:_
I ran the database migration on dev:

```
npm run tools migrate -- --name 2020_12_02_addReleaseHistoryKey
```

Then I ran the backfill:

```
npm run tools migrate -- --name data/2020_12_02_addReleaseHistoryKeys
```

I observed that the only errors were `No merges`, which will occur when the dev and prod Firebase instances aren't in sync — I don't expect to see any of these in production except on Pubs that are already broken anyway. After running this, I expect that every Release with a `sourceBranchKey` (which should supposedly be the corresponding draft key) should now have an identical `historyKey`, so I ran:

```
SELECT * FROM "Releases" WHERE "Releases"."sourceBranchKey" IS NOT NULL AND "Releases"."historyKey" != -1 AND "Releases"."sourceBranchKey" != "Releases"."historyKey"
```

This yielded only 24 results (out of about 18,000 Releases) which tells me that the migration computed the correct `historyKey` for almost all Pubs. I'll write another script to clean these up by comparing entire documents in a separate PR.

### DIFF
--- a/server/release/__tests__/api.test.js
+++ b/server/release/__tests__/api.test.js
@@ -84,7 +84,7 @@ it('will create a Release for admins of a Pub', async () => {
 			}),
 		)
 		.expect(201);
-	expect(release.sourceBranchKey).toEqual(0);
+	expect(release.historyKey).toEqual(0);
 });
 
 it('will create a Release for admins of a Pub', async () => {
@@ -101,7 +101,7 @@ it('will create a Release for admins of a Pub', async () => {
 			}),
 		)
 		.expect(201);
-	expect(release.sourceBranchKey).toEqual(0);
+	expect(release.historyKey).toEqual(0);
 });
 
 it('will not create duplicate Releases for the same draft-key of a Pub', async () => {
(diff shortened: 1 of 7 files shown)


3mcd (approved):
LGTM! The migration is really well-written and easy to follow.

PR #1182, opened 2020-12-08
dev: separate deps from dev deps

3mcd:
This PR separates development dependencies from normal runtime dependencies in the package.json file. This reduces the build size by about 80MB.

_Test Plan_
- `npm i`
- Run unit tests and verify Storybook is working.
- Perform a quick smoke test of the app.


idreyn (approved):
None of the removals from package.json look like runtime dependencies, so LGTM!

PR #1183, opened 2020-12-08
dev: Rename DiscussionNew to Discussion

idreyn:
During our last big migration we created a new Discussion model and named it DiscussionNew, with the expectation that it would eventually be renamed. I am now renaming it. The migration to actually rename the table will have to be somewhat carefully timed to correspond to the few seconds of server downtime during deploys, but I don't expect it will take long.

_Test plan:_
Run the entire unit test suite.


3mcd (approved):
LGTM!

Issue #1171, opened 2020-11-24
Pub node links being covered by Pub scroll header on jump


gabestein:
**What went wrong, step-by-step?**
1. Visit https://projections.pubpub.org/pub/3fa17lcj/release/1#overview-of-the-papers
2. Notice that the section header gets covered by the pub scroll header

**What did you expect to happen?**
The page should jump so that the section header is visible below the pub scroll header

**What URL can we use to see the issue?**
https://projections.pubpub.org/pub/3fa17lcj/release/1#overview-of-the-papers

**What browser/operating system were you using?**
Chrome / OS X
---

**Have any screenshots? (Attach them by dragging the image here)**

**Who reported this? (If not you)**
Alana Brooks

idreyn:
This is the default browser behavior — we'll need to add a custom bit of router-like code to respond to changes in the URL hash by scrolling to the "right" part of the page.

gabestein:
Yeah, that's a bit of a bummer. I wonder if there's a more CSS-y fix...I can do a bit of poking around.

3mcd:
This is partially addressed by #1186, although we still have this issue when clicking on reference links. I'll address this in a follow-up PR.

Issue #1170, opened 2020-11-24
Links to pub nodes work only sporadically on initial page load


gabestein:
**What went wrong, step-by-step?**
1. Visit a link to a pub node, ie https://readingthecomments.mitpress.mit.edu/pub/ujeil296/release/2#r9630153434
2. Browser will typically not jump to the node – sometimes it does if you refresh and it's already loaded the page, indicating it may be some sort of browser order issue.

**What did you expect to happen?**
Browser jumps to the node.

**What URL can we use to see the issue?**
https://readingthecomments.mitpress.mit.edu/pub/ujeil296/release/2#r9630153434

**What browser/operating system were you using?**
Any
---

**Have any screenshots? (Attach them by dragging the image here)**

**Who reported this? (If not you)**

3mcd:
Fixed by #1186

PR #1186, opened 2020-12-09
fix: correctly scroll linked element on Pub load

3mcd:
This is a band-aid fix for #1170 and #1171. This code simply executes a function 500ms after a Pub loads before scrolling to the element who's id is `window.location.hash`, only if that element is not already in the viewport.


idreyn (approved):
I misread the `setTimeout` as `setInterval` when we were chatting about this just now, and in light of that, this looks pretty solid!

PR #1191, opened 2020-12-11
Update pricing page

gabestein:
Updates pricing page for late 2020 based on Community feedback and internal discussions.

screencapture-localhost-9876-pricing-2020-12-11-13_51_11


isTravis (approved):
This is great! Kudos on the super quick turnaround on making these updates. 

I made a couple small tweaks via PR #1194 to simplify the CSS a bit and make font sizes more consistent. Take em or leave em - happy to have this go live with whichever you think is best!

PR #1195, opened 2020-12-14
update: Create Doc model for each Release

idreyn:
This is the thing that's kind of at the heart of our plan to move Releases to Postgres — here, we store the Pub content in a Doc associated with every Release. We're not _reading_ from them, yet, though — that will be a much more complicated set of changes that we'll have to host in parallel with the existing Firebase code for a bit until it's clear that it works. For now we'll just get the codebase in the habit of storing Docs, and run a migration to backfill Docs for existing Releases.

_Test plan:_
Run `npm run test-dev server/releases`


3mcd (approved):
LGTM!

PR #1198, opened 2020-12-15
fix: scroll to top of element and account for sticky pub header

### LINKED ISSUE #1171: Pub node links being covered by Pub scroll header on jump

3mcd:
Fixes #1171

Clicking on internal URLs (with hashes) within a Pub will scroll to the correct element, but the browser doesn't account for the sticky pub header, so the top of the element isn't visible. This PR overrides the default scroll behavior of hash links and scrolls to the element with a 50px offset so the entire element is visible. In addition, code was added to automatically convert internal links to hashes when pasted into the URL field in the anchor controls.

_Test Plan_
- In a Pub with enough content to scroll
- Copy a link towards the bottom of the Pub using the click-to-copy button
- Create a link at the top of the Pub and paste the URL in. The URL's host and path should automatically be removed, leaving only the hash.
- Publish the Pub.
- Click on the link in the release view. The window should scroll to the top of the element, and the entire element should be visible, and not overlapped by the sticky header component.

### DIFF
--- a/client/components/FormattingBar/controlComponents/ControlsLink.tsx
+++ b/client/components/FormattingBar/controlComponents/ControlsLink.tsx
@@ -2,6 +2,9 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { useDebounce } from 'use-debounce';
 import { moveToEndOfSelection } from 'components/Editor';
 import { Button, AnchorButton, InputGroup } from '@blueprintjs/core';
+import { usePubContext, usePubData } from 'client/containers/Pub/pubHooks';
+import { pubUrl } from 'utils/canonicalUrls';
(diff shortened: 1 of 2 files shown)


idreyn (approved):
LGTM!

PR #1200, opened 2020-12-16
fix: Wrap long file names in download cards

gabestein:
Fixes an issue brought to our attention by @pkgw that caused long file names to overflow the download card and push the download button out of it. This still causes some awkwardness for file names with underscores or dashes, since that's not technically a CSS break character, but it's better than the name overflowing the card.

Screen Shot 2020-12-16 at 09 26 39

Screen Shot 2020-12-16 at 09 26 44

_Test plan_
1. Upload a file with a really long name (look here's one: 
[A File With A Bizzarely Long Name That Is Really Too Long Why Is It This Long.txt](https://github.com/pubpub/pubpub/files/5704392/A.File.With.A.Bizzarely.Long.Name.That.Is.Really.Too.Long.Why.Is.It.This.Long.txt)
). 
1. Make sure the name breaks rather than overflowing the card.

### DIFF
--- a/client/components/Editor/styles/file.scss
+++ b/client/components/Editor/styles/file.scss
@@ -16,6 +16,7 @@
 			flex-grow: 1;
 			flex-shrink: 1;
 			margin: 0em 1em;
+			word-wrap: break-word;
 			a {
 				color: inherit;
 				&:hover {


idreyn (approved):
Looks great! Thought I fixed this earlier this year but I forgot to account for filenames with no whitespace.

PR #1201, opened 2020-12-17
fix: sync Prosemirror table captions in update() method of NodeView

idreyn:
Earlier this morning I made a change to `prosemirror-reactive` that cause reactive NodeView wrappers to return the value of `update()` provided from the NodeViews they wrap. This fixes behavior in highly stateful NodeViews like resizable tables, but breaks out table labeling code in our custom TableView subclass.

The fix has two parts:
1. In `prosemirror-reactive`, make sure wrapped NodeViews always receive a reactive copy to their update method. (See https://github.com/pubpub/prosemirror-reactive/commit/026d6135f5b2b3682cc47ba506679a80a308e56b)
2. In our Table view, call `syncCaption` from `update`. Don't forget to remove `caption` nodes left over from previous calls.

All I can say is...thank god for React.

### DIFF
--- a/client/components/Editor/plugins/table.ts
+++ b/client/components/Editor/plugins/table.ts
@@ -28,13 +28,12 @@ function wrapDomSerializer(domSerializer: DOMSerializer) {
 class PubTableView extends TableView {
 	constructor(node, ...rest) {
 		super(node, ...rest);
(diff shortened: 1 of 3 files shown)


3mcd (approved):
Looks good to me :)

PR #1203, opened 2020-12-17
feature: New Firebase unit testing infra

idreyn:
This change integrates the [Firebase Emulator Suite](https://firebase.google.com/docs/emulator-suite) into our testing process, allowing us to write unit tests that read to and write from an actual, locally hosted Firebase database instance. That means we no longer have to stub out these functions when writing tests for features that depend on them, and that we could eventually have unit tests for some client-side Firebase functionality as well. Also included are the beginnings of a new utility for writing Pub data to firebase (`editPub()`) and the application of that utility to the Release tests.

The emulator requires the `firebase-cli` package to be installed on dynos, and the datastore server is written in Java, so I have made changes to the Heroku `test` environment to install this dependency and add a JRE buildpack. I also noticed that the `puppeteer` and `apt` buildpacks are not currently needed to run our tests, so I have removed them. For now, it is necessary to run our tests using `--runInBand` to prevent the Firebase emulator from timing out.

Finally, I have upgraded the `firebase-admin` package from `8.1.0` through a major version to `9.4.2`. I had a look at the [changelog](https://firebase.google.com/support/release-notes/admin/node) and as far as I am concerned there is nothing we need to respond to. Still, we'll want to manually smoke test basic Pub firebase functionality before deploying.

_Test plan:_
If the tests build and pass on Heroku, that's good enough for me!


client/utils/upload.ts
```diff
@@ -1,7 +1,7 @@
 import { isProd } from 'utils/environment';
 import { generateHash } from 'utils/hashes';
 
-const checkForAsset = (url) => {
+const checkForAsset = (url): Promise<void> => {
```
idreyn, line 4: Makes Typescript happy

workers/tasks/import/bulk/cli.ts
```diff
@@ -67,7 +67,6 @@ export const getActor = async (userSlug) => {
 const readPlanFromFile = async (path) => {
 	const contents = await fs.readFile(path);
 	try {
-		// @ts-expect-error ts-migrate(2345) FIXME: Argument of type 'Buffer' is not assignable to par... Remove this comment to see the full error message
```
idreyn, line 70: Makes Typescript happy

3mcd (approved):
All changes LTGM and tested locally.

PR #1205, opened 2020-12-18
dev: Remove a large number of @ts-expect-error directives

idreyn:
I got really carried away removing a `@ts-expect-error` directives that TS would incorrectly flag as unneeded when switching branches...and ended up removing a ton of them. I feel pretty confident that there are no runtime behavior changes here.

This takes us from 1182 `@ts-expect-error` comments to 1129...but before I checked I would have guessed that I removed, like, three hundred of them here 😭


3mcd (approved):
Awesome!!

PR #1206, opened 2020-12-21
dev: Move client and server tsconfig files to root

idreyn:
We currently have a `server/tsconfig.json` and a `client/tsconfig.json`, each of which has an `excludes` array that dictates that test files should be excluded from TS builds. This is all well and good, except in the test files themselves. For a `*.test.ts` file in `server`, VSCode (and presumably other IDEs) look at successive parent directories until it finds `server/tsconfig.json` — which tells it to ignore test files! This doesn't prevent us from running tests, but it means we have no TS language support in these files.

The solution in this PR is to create a `tsconfig.client.json` and `tsconfig.server.json` that the IDE is unaware of, but that the build infrastructure knows about.

_Test plan:_
- I ran `npm run build-prod` and observed that test files were excluded from both `dist/server` and `dist/client`.
- I ran `npm test` and observed that all the tests pass.
- I confirmed that with this change I now have TS language support in test files when editing with VSCode.


3mcd (approved):

server/models.ts
```diff
@@ -92,16 +92,8 @@ export const attributesPublicUser = [
...
-		attributes: providedAttrs,
-		...restOptions
-	}: {
-		attributes?: string[];
-		[k: string]: any;
-	}) => {
-		const attributes = providedAttrs
-			? [...new Set([...attributesPublicUser, ...providedAttrs])]
-			: attributesPublicUser;
+	return ({ attributes: providedAttributes = [], ...restOptions }) => {
```
idreyn, line 95: This is stupid (and deserves more attention) but we cannot actually use TypeScript features in this file because it is required from the `stubstub` global setup without being touched by TS.

PR #1212, opened 2020-12-23
feature: Add support for image alt text and improve a11y of figcaptions

### LINKED ISSUE #1208: Better handling of figcaption and alt for screen readers
Previously, we duplicated figcaptions into alt values in images. This worked okay, but because figcaptions contain html, caused some screen readers to read the contents as HTML. It also meant that some screen readers would read the caption twice in some scenarios. So, we removed it. 

However, in this current scenario, many screen readers will ignore images/graphics altogether, particularly when 

Instead, we should do the following:
- Add a plaintext alt field to the editor image node for cases where images need to be described in more depth than a caption allows, e.g. data visualization.
- When only a `figcaption` is present, we should add an `aria-labelledby` attribute to the `` tag, referencing the `figcaption`. This will cause all major screen readers to read the text of the `figcaption` when browsing both via the down arrow and the graphics shortcut, and to do so in without reading the html as text.
- When `alt` _and_ `figcaption` are present, we should add an `aria-describedby` attribute to the `` tag, referencing the `figcaption`. This will cause all major screen readers to read the text of the `figcaption` and the `alt` when browsing via the down arrow. When browsing via the graphics key, some screen readers will read only the alt, while others will read the entire grouping.
- When only an `alt` text is present, that should suffice – screen readers will read only the alt text with both down arrow and the graphics key.

Source: https://www.hassellinclusion.com/blog/figure-figcaption-extended-alternate-text-screen-readers/
Examples: https://examples.hassellinclusion.com/images/figure-figcaption01.html

Vastly improve the PubPub experience for users using screen readers, in particular properly grouping figcaptions with their images for most readers, and allowing complex graphics to have separate, more descriptive alt texts, than captions alone provide.

idreyn:
Resolves #1208

Our current attempt at screenreader support for images and other captionable media has been to use a `figure`/`figcaption` pair, which should cause the `figcaption` contents to be read in a way similar to an image's alt text. As described in #1208, Gabe (and a Twitter user) found that this is not actually the case for many OS/reader/browser combinations. So this PR has two changes:

- We add an `aria-describedby` (or occasionally `aria-labelledby`) to `img`, `video`, `audio`, `iframe` tags that points at the `figcaption`, to improve the odds of the caption being associated with the media by the SR.
- For instances where SR users would benefit from more detail than is appropriate for a caption, we add an `altText` field to the image node schema, and a corresponding control to modify it in the editor:

image

_Test plan:_
Check that:
- Adding alt text to images works and persists to the document
- Images without any alt text from existing documents don't crash or anything
- The alt text is applied to the image's `alt` tag.
- Images have an `aria-describedby` pointing to the caption if they also have an alt text, and an `aria-labelledby` pointing to the caption otherwise. If there is no caption or alt text provided by the Prosemirror image node, the corresponding DOM node should have no `aria-` attributes and `alt=""` (which will make the image invisible to the screenreader).

### DIFF
--- a/client/components/Editor/schemas/audio.ts
+++ b/client/components/Editor/schemas/audio.ts
@@ -41,6 +41,7 @@ export default {
 			},
 		],
 		toDOM: (node, { isReact } = { isReact: false }) => {
+			const figcaptionId = `${node.attrs.id}-figure-caption`;
 			return [
 				'figure',
 				{
@@ -56,11 +57,12 @@ export default {
 						preload: 'metadata',
 						src: node.attrs.url,
 						alt: node.attrs.caption,
+						'aria-describedby': figcaptionId,
 					},
 				],
 				[
 					'figcaption',
-					{},
+					{ id: figcaptionId },
 					pruneFalsyValues([
 						'div',
 						{},
--- a/client/components/Editor/schemas/iframe.ts
+++ b/client/components/Editor/schemas/iframe.ts
@@ -30,8 +30,8 @@ export default {
 				},
 			},
 		],
-		// @ts-expect-error ts-migrate(2525) FIXME: Initializer provides no value for this binding ele... Remove this comment to see the full error message
-		toDOM: (node, { isReact } = {}) => {
+		toDOM: (node, { isReact } = { isReact: false }) => {
+			const figcaptionId = `${node.attrs.id}-figure-caption`;
 			return [
 				'figure',
 				{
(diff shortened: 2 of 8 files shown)


gabestein (approved):
Approved from a functionality/QA perspective! Looks great, and thank you for jumping on this so quickly. A huge, long overdue improvement. 

Note: I added downloading a PDF and an MS Word to the test plan. Does not have an impact – in Word, the caption is still added to the image description, even if an alt tag is present. I believe that this is being worked on in newer versions of Pandoc (https://github.com/jgm/pandoc/issues/6782), so we might want to consider upgrading soon, especially given their recent table work. But I digress...

3mcd (approved):
Looks good code-wise and functionality-wise. One UX suggestion: the state of the text areas should be persisted when changing between tabs, and the update button should update both values. I had a brief moment of confusion when I filled out the alt text first, then the caption, then pressed update, and only the caption was saved.

3mcd:
> Looks good code-wise and functionality-wise. One UX suggestion: the state of the text areas should be persisted when changing between tabs, and the update button should update both values. I had a brief moment of confusion when I filled out the alt text first, then the caption, then pressed update, and only the caption was saved.

It looks like this only happens when the image has neither a caption nor alt text. If you fill out the alt text first, caption second, then press update, only the caption is saved.

idreyn:
That's odd but good catch, I'll look into it

PR #1215, opened 2020-12-29
dev: Split Editor.tsx into more files and add stronger typing

idreyn:
I wanted to make some changes to the editor (in an upcoming PR) but I didn't feel comfortable doing so because the file has gotten a bit too big for its britches, so I decided to split it out into a few smaller files. There's no big rethink of how things work here and there should be _almost_ no behavioral changes — just smaller files and more types.

The one thing I _have_ changed here is that the a couple of the editor props — specifically `onError`, `onKeypress`, `onScrollToSelection`, and `isReadOnly` now follow the React model and update when the props update. We accomplish this by calling `editorView.setProps` within a set of `useEffect` hooks. One consequence of this is that the callbacks passed in here should be memoized using `useCallback`, which is good practice anyway. In the future, `<Editor>` props that can be updated without needing to reconstruct a new `EditorView` should behave this way, and we may eventually want a more scalable pattern for ensuring that this happens.

_Test plan:_
I'm chickening out of writing a specific test plan here for now and I'm going to come up with a list of things to playtest in the editor that we can use to test this release (and others that modify the editor in the future). For now, you should just be able to check out this branch and ensure that everything, including the collaborative features, seems to be working as expected.

### DIFF
--- a/client/utils/suggestions/suggestionManager.ts
+++ b/client/utils/suggestions/suggestionManager.ts
@@ -110,6 +110,7 @@ export default class SuggestionManager<T> {
 	getPosition() {
 		if (!this.isSuggesting()) {
 			return {
+				position: 'absolute' as const,
 				top: 0,
 				left: 0,
 			};
@@ -120,6 +121,7 @@ export default class SuggestionManager<T> {
 		const parent = document.body.getBoundingClientRect();
 
 		return {
+			position: 'absolute' as const,
 			top: bounds.bottom - parent.top,
 			left: bounds.left - parent.left,
 		};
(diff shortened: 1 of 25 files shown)


client/utils/suggestions/suggestionManager.ts
```diff
@@ -110,6 +110,7 @@ export default class SuggestionManager<T> {
 	getPosition() {
 		if (!this.isSuggesting()) {
 			return {
+				position: 'absolute' as const,
```
3mcd, line 113: Wow, I didn't know about `as const`. That's super useful.

3mcd (approved):
LGTM

PR #1223, opened 2021-01-05
dev: PubPage type and other (mostly TypeScript) improvements

idreyn:


isTravis (approved):
👍 
35 `@ts-expect-error`s down, 3.9e13 to go.

PR #1224, opened 2021-01-05
dev: Add some testing utils for Firebase draft (and some new tests)

idreyn:
Breaking this out of a larger PR, this one adds an `editFirebaseDraft` utility that lets us write Prosemirror steps to a Pub in the Firebase emulator for later retrieval. Also included is the `getStepsInChangeRange` utility that we'll be using in the near future.

Test plan:
```
npm run test
```


server/utils/firebase/changes.ts
```diff
@@ -0,0 +1,31 @@
+import { Step } from 'prosemirror-transform';
+import { Reference } from '@firebase/database-types';
+import { uncompressStepJSON } from 'prosemirror-compress-pubpub';
+
+import { editorSchema } from '../firebaseAdmin';
+
+export const getStepsInChangeRange = async (
```
idreyn, line 7: We will use this in an upcoming PR but it felt thematically appropriate to include here

package.json
```diff
@@ -169,6 +169,7 @@
 		"node": "12.13.0"
 	},
 	"devDependencies": {
+		"@firebase/database-types": "^0.6.1",
```
idreyn, line 172: This is the only dependency I added here, the rest of this just appears to be alphabetization.

PR #1225, opened 2021-01-05
update: Add DiscussionAnchor model and associated code

idreyn:
This PR adds a `DiscussionAnchor` model which will keep track of discussion positions in Releases, replacing the Anchor model, which will be removed. The `createRelease` query has been rewritten to create Release, Doc, and DiscussionAnchor models atomically in a transaction, but it will also continue to execute the Firebase merges we want to deprecate until we change how the Pub route handler loads content. This really gets us over a hump in this project since we now have all the new models required for Firebase-free Release content.

Included is a data migration to move content from  `Anchor` models and Firebase discussion content into `DiscussionAnchor` models. I ran this on the dev database and encountered only a handful of irregularities, which I'll follow up with by hand.

_Test plan:_
I ran the new tests with `npm run dev`. I also verified manually that I can make a Release through the UI, and that doing so creates new DiscussionAnchor models that appropriately map discussions to text in the Release doc.


server/discussionAnchor/model.ts
```diff
@@ -0,0 +1,18 @@
+export default (sequelize, dataTypes) => {
+	return sequelize.define(
+		'DiscussionAnchor',
+		{
+			id: sequelize.idType,
+			isOriginal: { type: dataTypes.BOOLEAN, allowNull: false },
+			discussionId: { type: dataTypes.UUID, allowNull: false },
+			historyKey: { type: dataTypes.INTEGER, allowNull: false },
+			selection: { type: dataTypes.JSONB, allowNull: true },
```
isTravis, line 9: Quick check just to verify my mental model how this is working:

- We used to store this `selection` value in Firebase. When the document changed, one of the clients would write a new value to the `selection` field that matched the `currentKey` of the doc.
- Each client manually managed all their local discussion selections though - so that value would only be used for new pageloads, not to keep decorations in sync across client instances.
- We now store this information in postgres, and update it at a similar cadence, and again, it's only relevant for page loads.

I think I'm following from the code below, but just to be explicit, what is the intention of when this field gets updated, and who triggers it?

isTravis (changes requested):
Playing with the user-facing functionality, it worked for me to:

- Make a discussion on the draft, have that position remain as the document changed.
- Create a release (draft discussion did not appear on release)
- Add discussion to release
- Edit draft (moving the draft and release discussion positions)
- Publish release
- Confirm release discussions and draft discussions maintained their position.

However, I can't get two browsers to synchronously edit. Other qubqub deployments don't seem to have this problem (e.g. https://qubqub-pr-1223.herokuapp.com/pub/na4f0cos/draft).

server/discussionAnchor/model.ts
idreyn: This doesn't replace the Firebase discussion anchoring information — just the `Anchors` table that currently exists. The primary difference between that table and this one is that the `selection` field for a given `DiscussionAnchor` never gets updated — instead, when a Release is created, we create a new `DiscussionAnchor` with a future history key that associates it with that Release. Keeping the old one around will help us anchor comments on previous Releases.

isTravis (approved):
Synchronous editing working for me now after you reset the dyno 👍

PR #1234, opened 2021-01-14
update: Rewrite Prosemirror discussion plugin

idreyn:
This PR replaces the existing Prosemirror discussions plugin with a rewritten version. Both plugins expose mostly the same interface — they create the same decorations which are used to mount discussions UI, and read and write the same objects from Firebase. The new one is a little more resilient to conditions that would cause a discussion to become un-anchored, in that it will detect and rebuild widget discussions that have been destroyed by Prosemirror mappings, and it can "fast-forward" discussions with obsolete discussion anchors in the same way that the `restoreDiscussionMaps` utility on the server does. It also fixes a longstanding problem where steps written during `EditorView` initialization — like new node IDs — would trigger a race condition that prevented this plugin from syncing discussions. Mostly, it provides a more stable surface to build on in this area with types and a new separation between data and presentation.

This is not ready to be deployed to production as it still needs to be adapted to mount the Release-side `DiscussionAnchors` when they are provided to the plugin — that code has mostly been written, but will appear in a separate PR. I would also like to write a handful of unit tests in a different PR.

_Test plan:_
For now our manual testing will have to suffice. This involved several of us opening the same Pub and simultaneously making content changes and adding anchored discussions, and verifying that there were no editor errors and the discussions did not get out of sync.


isTravis (approved):
This is great. Really wonderful to have cleaner file structure and types (and obviously the anchor fast-forwarding, etc).

I played with it a bit more in this deployment given that you said you had some cleanup to do (e.g. I see the highlights don't stay visible as they did when we were testing, which is good), and all still works as expected on my end.

The one caveat I have in mind is that the bugs in this space have been tough-to-find race condition issues. The movement of the discussions plugin out of the collab module and cleaner file/function splitting should help, but that'd be the only thing I'd keep my out for in the few days after launch.

PR #1238, opened 2021-01-14
dev: Some refactoring in anticipation of new Discussion plugin

idreyn:
This is a bit of a grab bag of changes that the new Discussions plugin (#1234) depends on, but that I wanted to separate into a different PR. Highlights:

- The `server/utils/firebase` directory has been moved to `client/components/Editor` since we need its functionality on the client
- The `getFirebaseDoc` utility now returns an un-serialized Prosemirror Node.
- The `copyDiscussionMapsToBranch` utility has been removed. This was off-label functionality we sometimes used to support content production and it will need to be rewritten under our new Release architecture.

I don't expect any changes in behavior from this PR.


client/components/Editor/utils/changes.ts
```diff
@@ -1,14 +1,13 @@
+import { Node, Schema } from 'prosemirror-model';
 import { Step } from 'prosemirror-transform';
-import firebase from 'firebase-admin';
 import { uncompressStepJSON } from 'prosemirror-compress-pubpub';
 
-import { editorSchema } from '../firebaseAdmin';
-
 export const getStepsInChangeRange = async (
 	branchRef: firebase.database.Reference,
```
isTravis, line 6: Is `firebase` here in the global namespace?

isTravis (approved):
Looks good! No unexpected behavior in manual testing with discussions, releases, and anchor mapping.

Just the one question about the `firebase` reference I put inline - but it seems to compile - so might not be an issue?

client/components/Editor/utils/changes.ts
idreyn: Yeah. I actually don't know why but TypeScript seems to accept it — and ultimately this may be our only option since writing:

```
import firebase from 'firebase';
```

throws a warning on the client recommending that we don't do this and use `firebase/app` instead. Likely it's dragging in a bunch of Firebase features that we don't use.

isTravis: Interesting. Might be worth an explicit comment hinting at something there (or even just pointing back to this PR) in case we do bump into firebase weirdness at some point.

PR #1239, opened 2021-01-15
feature: Paste and drag images into the editor

### LINKED ISSUE #1217: Copying an image into a Pub does not work
When? Why? How?

idreyn:
Does what it says on the tin, and resolves #1217! Pasting and dragging images into the editor should now work — the process creates a placeholder decoration which is removed when the image finishes uploading. Check it out:

[screencap of drag+drop](https://user-images.githubusercontent.com/2208769/104791286-3c9bf080-5768-11eb-849f-cc82404463e2.mp4)

Big thanks to the author of [this Prosemirror forum post](https://discuss.prosemirror.net/t/image-plugin-with-uploading-image-title-and-a-few-other-things/3408) for pointing us to their repo that provides a plugin with similar functionality — they did all the hard work of learning about the DOM clipboard APIs so I didn't have to.

_Test plan:_
Try adding an image from the following sources:
- Your clipboard, copied from the browser
- Your clipboard, copied from your file system
- A file dragged in from your file system

Might be possible to get files dragged in from other browser tabs but that might also require downloading and re-uploading foreign URLs so this is good for now.


isTravis (approved):
Amazing! So nice to finally have this functionality. Pasting and dragging both now sync across browsers for me nicely.

I had some syncing issues on qubqub, but I think it might be the deployment because I don't see where in this code a problem like that could've been introduced. Before you promote to production, could we mash keys and test out the syncing together just for a sanity check?

PR #1240, opened 2021-01-15
dev: Email all users tool

gabestein:
To discuss next week. Mostly to be run locally for privacy policy etc updates, but wanted to get the PR up so y'all can look at it Tuesday.


gabestein:
- Order query and count offset
- Add language explaining why no opt-out

idreyn (approved):
Questions about whether the Mailgun library can handle some of the parallelism internally aside, this looks good to me!

PR #1245, opened 2021-01-21
dev: Load Releases from Postgres

idreyn:
As of this PR, Releases are now loaded from Postgres. Releases and drafts of Pubs should work properly, but I'm about to embark on an even bigger piece of refactoring to remove Branches, so I'm not too picky about whether this works — whatever testing we do now will have to be done about sevenfold after the next PR.


isTravis (approved):
Great to see so many typescript fixes and updates coming in this. 

I think I have my head wrapped around the meat of this refactor, but if there are nuances that you think are particularly subtle, it might be helpful for me to walk through those with you on a call. Perhaps we do that in the context of the pending Branch PR.

PR #1246, opened 2021-01-21
dev: Add Draft model for Pubs (and more?)

idreyn:
With Postgres Releases in place and no Branch-related features exposed to users any more, it is time to remove Branches as a feature of our data model. Currently they serve two purposes: they tell us where in Firebase to find the draft content for a Pub, and when that content was last updated. As a first step in this process, this PR introduces a new `Draft` model for this specific purpose. Eventually we will probably also use this model to point to non-Pub collaborative content in Firebase, like Page blocks or whatever we cook up.

_Test plan:_
- I ran the schema and backfill migrations on the dev database and verified the result
- I ran `npm run test-dev server/pub/` to check that creating a Pub now creates an associated Draft.


isTravis (approved):
👍 Looks pretty straightforward and good to me.

PR #1247, opened 2021-01-22
dev: Remove Branches

idreyn:
This PR removes Branches from our codebase and data model.  Going to deploy this to duqduq for easier testing!

Things I think are temporarily broken here to be addressed later:
- [ ] Discussion anchors on Releases

Things to test:
 - [ ] Pub dashboard Overview and Settings tabs render
 - [ ] Pub direct download `/pub/:slug/download/:format` route works as expected
 - [ ] Pub history on draft works
 - [ ] Creating/retrieving exports work
 - [ ] `/pub/:slug/branch` redirects work
 - [ ] Searching for draft and released Pubs works
 - [ ] Newly created Pubs appear in Algolia index and can be searched
 - [ ] Both existing and newly created Pubs can be edited
 	- [ ] Adding checkpoints works (just keep typing)
 	- [ ] Adding discussion anchors works  

Before deploy:
- [ ] Run migration to backfill draft models
- [ ] Run migration to remove branch-related columns from database
- [ ] Copy new firebase rules to production

After deploy:
- [ ] Re-run draft backfill to catch stragglers


isTravis (approved):
C'est fantastique. Un monde sans branches. Un monde de paix. 

Seriously - this is a huge step in cleaning up the codebase and I'm hopeful it will simplify our lives immediately. Running through the tests, here are some of the things I found:

- On the draft history slider, the fast-forward and fast-rewind buttons (two arrows) are both greyed out when I wouldn't expect them to be. I wonder if it's not getting the start/end point of branches properly. 
cursor

- The search continues to be a bit confusing to me (this could be something also problematic in production, or an algolia caching issue like we've seen before). I can confirm that new pubs I create wind up in the algolia dev index. However, the search page doesn't return the pub I can see in the algolia dashboard. Further, on the qubqub test community, the search results seem to be from all communities? Maybe that's a dev/review-app flag that's intentional?

- In the migration plan, I just wanna confirm that re-running the backfill to catch stragglers will still work after `Run migration to remove branch-related columns from database`.

I looked through the files, but it's a huge number, so I'm more comfortable with doing lots of QA testing, rather than trying to comb through and find logic errors. Also happy to sync in person if you wanna give an overview of the changes and approach! 👍

PR #1264, opened 2021-02-01
update: Anchor Release discussions with new plugin

idreyn:
This PR finally makes discussions visible on Releases. It also ensures that for now, discussions made on either the draft or on a Release will stay on that side of the Pub based on their visibility — regardless of the user's permission to see them. We can relax this constraint once we have UI that helps users understand and control discussion visibility.

I have also removed the old `Anchor` model here since it's been entirely supplanted by `DiscussionAnchor`.

_Test plan:_
On my devserver:
- I opened this Pub: http://localhost:9876/pub/afqs9x3a/release/2 and observed that Discussions are visible and anchored. 
- I also checked the previous Releases and noted that it had Discussions anchored as well.
- I returned to the latest Release and made a new Discussion there, and verified that it was anchored
- I checked the draft of this Pub and verified that a different set of Discussions are anchored there.
- I made some changes to the draft and created a new Release, and verified that the Release-side discussions were mapped through those changes.


client/components/Editor/plugins/discussions/discussionsState.ts
```diff
@@ -172,6 +159,19 @@ export const createDiscussionsState = (options: Options) => {
 		});
 	};
 
+	const addDiscussion = (discussionId: string, selection: DiscussionSelection) => {
```
idreyn, line 162: This needs to call `asynchronouslyUpdateDiscussion`, not `updateDiscussion`, hence the move

client/components/Editor/plugins/discussions/util.ts
```diff
@@ -32,15 +32,23 @@ export const mapDiscussionSelectionThroughSteps = (
...
-	const mapping = new Mapping(steps.map((step) => step.getMap()));
 	const { from, to } = getRangeFromSelection(selection!);
 	// This matches the default mapping of inline Decorations, which we wish to emulate here.
 	// The default Selection mapping behavior is too greedy; for instance, if a selection is made
 	// at the beginning of the document, any text inserted before that selection is also mapped
 	// into that selection, which is not what we want for anchored Discussions.
 	// See: https://github.com/ProseMirror/prosemirror-view/blob/49889c576bc450a3008117249f9ef9beb6bf2969/src/decoration.js#L36
-	const nextFrom = mapping.map(from, 1);
-	const nextTo = mapping.map(to, -1);
+	const { from: nextFrom, to: nextTo } = steps.reduce(
```
idreyn, line 41: Sadly we cannot use this approach from the previous revision:
```
const mapping = new Mapping(steps.map((step) => step.getMap()));
// map positions with mapping.map()
```
Instead, we need to map the positions through each Step, one at a time, or we'll lose important information. For instance, the above `Mapping` treats the act of erasing content and then adding new content as a simple replacement, which improperly preserves the highlight on the old content where it may be entirely irrelevant. I _think_ that this is the behavior we want.

(This is only important for Release-side discussions, which get mapped through many steps at once — draft-side discussions go through the same function, but typically `steps.length === 1` in that case and the result is the same)

server/release/__tests__/api.test.ts
```diff
@@ -219,8 +219,8 @@ describe('/api/releases', () => {
 			createDiscussionForPub({
 				pubId: pub.id,
 				userId: rando.id,
-				from: 0,
```
idreyn, line 222: Changed since this is not a valid position range (`0` is outside the first block)

client/components/Editor/plugins/discussions/util.ts
isTravis: Agreed that that is the behavior we want.

client/components/Editor/plugins/discussions/util.ts
```diff
@@ -32,15 +32,23 @@ export const mapDiscussionSelectionThroughSteps = (
...
+		(state, step) => {
+			const { from: currentFrom, to: currentTo } = state;
+			const stepMap = step.getMap();
+			return {
+				from: stepMap.map(currentFrom, -1),
```
isTravis, line 46: Is this replicating the functionality described in the comment block above? And if so, is the -1 intentional (The removed line had a positive value for from: `const nextFrom = mapping.map(from, 1);`

idreyn: Good eye, that would have been a nightmare to track down

isTravis: 👍 Testing it out, I can see that it produced the wrong behavior too - so we have a nice condition to validate against once this fix lands on qubqub. This comment was made when "Hello" was the first line in the doc, and adding content before it shows the mapping being off.
Screen Shot 2021-02-01 at 3 53 05 PM

isTravis:
I'm getting a bug that only seems to happen on draft. On creating a new anchored discussion, two identical discussions seem to populate on the right-hand margin. This goes away on page reload, so I think it's just at the local rendering layer (not persisted in a duplicate manner to the db). 

Screen Shot 2021-02-01 at 3 53 31 PM

client/components/Editor/plugins/discussions/util.ts
idreyn: ++ I added a unit test case for this as well

PR #1269, opened 2021-02-03
dev: Fix some accumulated lint errors

idreyn:
When we did our TS migrations last year we forgot to tell the linter to check `.ts(x)` files before allowing code to be pushed, so there are quite a few errors that have accumulated since that time. This PR fixes them all, hopefully without any behavioral changes.

_Test plan:_
Just generally clicking around to make sure things are okay, with special attention on the Pub header and footer where I did some PropTypes -> TS conversion.


isTravis (approved):
What a satisfying PR. I find fixing lint errors so satisfying when I have the time to sit and do it in bulk.

Poked around with the duqduq deployment and couldn't find anything amiss. Also looked through the Pub Header/Footer files and things looked pretty clean. So good from my end!

PR #1271, opened 2021-02-04
Add tool to crawl pubs for invalid DOIs

gabestein:
This adds a tool that allows you to crawl Pubs within a community for invalid DOIs. Useful for checking and finding which Pubs need to be resubmitted.

Also, removes a list of email addresses we accidentally checked into the email tool and fixes some lint errors. Ugh.


isTravis (approved):
👍 Looks like a helpful tool. I could see eventually caching this functionality into something client-side visible so admins can test/debug themselves. But for now, this is obviously a big first step.

gabestein:
Yup, I can see us running this periodically against all DOI'd pubs and adding a doi_resolved column to the DB or something that warns admins when Pubs have a DOI but aren't resolved.

PR #1277, opened 2021-02-11
new: Show pandoc warnings when importing LaTeX

gabestein:
Improves PubPub's import UX to show Pandoc warnings when importing LaTeX documents. This is helpful because it shows the places where Pandoc has had to skip or modify LaTeX functions to successfully import the document, making it easier for production teams to understand why and how they'll need to fix errors introduced due to the unavoidable incompatibilities of the LateX schema, the pandoc schema, and PubPub's prosemirror schema.

I've tried to keep this incredibly simple, with the understanding that at some point we might want to parse these messages to show more info or give an indication of severity, but my sense is we actually won't quite know how to do that until we get more user feedback on the feature, and that this additional info, combined with a link to our new [help documentation](https://help.pubpub.org/pub/2gfph6h8/release/2), is already a huge improvement.

Having said that, I think there could be more thought put into how and where the data is passed to the client, parsed, and displayed, so please feel free to make suggestions there (and obviously on anything else, just calling that out in particular).

_Test Plan_
1. Import a complex latex doc on prod and make sure it imports without error.
2. Import the same complex latex doc on qubqub (not local, as your local pandoc version may vary) and make sure it imports but shows useful information.
3. Import a docx, an epub, and a txt file on prod and make sure they import.
4. Repeat on qubqub and make sure they import without showing information.

### DIFF
--- a/client/containers/Pub/PubDocument/PubFileImport/FileImportDialog.tsx
+++ b/client/containers/Pub/PubDocument/PubFileImport/FileImportDialog.tsx
@@ -80,7 +80,7 @@ const FileImportDialog = ({
 
 	const isImportDisabled = !hasDocumentToImport || incompleteUploads.length > 0 || isImporting;
 	// @ts-expect-error ts-migrate(2339) FIXME: Property 'doc' does not exist on type '{}'.
-	const { doc, warnings = [], pandocInfos, error, proposedMetadata } = importResult;
+	const { doc, warnings = [], pandocErrorOutput, error, proposedMetadata } = importResult;
 	const hasProposedMetadata = proposedMetadata && Object.keys(proposedMetadata).length > 0;
 
 	useKeyPressEvent('/', (evt) => {
@@ -186,15 +186,7 @@ const FileImportDialog = ({
 			const missingCitations = warnings
 				.filter((w) => w.type === 'missingCitation')
 				.map((w) => w.id);
-			const pandocWarnings = pandocInfos.split('\n').map((info) => {
-				return (
-					<span>
-						{info}
-						<br />
-					</span>
-				);
-			});
-			if (missingImages.length > 0 || missingCitations.length > 0 || pandocInfos) {
+			if (missingImages.length > 0 || missingCitations.length > 0 || pandocErrorOutput) {
 				return (
 					<Callout
 						aria-live="assertive"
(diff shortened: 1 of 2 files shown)


client/containers/Pub/PubDocument/PubFileImport/FileImportDialog.tsx
```diff
@@ -80,7 +80,7 @@ const FileImportDialog = ({
 
 	const isImportDisabled = !hasDocumentToImport || incompleteUploads.length > 0 || isImporting;
 	// @ts-expect-error ts-migrate(2339) FIXME: Property 'doc' does not exist on type '{}'.
-	const { doc, warnings = [], error, proposedMetadata } = importResult;
+	const { doc, warnings = [], pandocInfos, error, proposedMetadata } = importResult;
```
idreyn, line 83: Could you give this value a more informative name, like `pandocErrorOutput`?

client/containers/Pub/PubDocument/PubFileImport/FileImportDialog.tsx
```diff
@@ -186,8 +186,15 @@ const FileImportDialog = ({
 			const missingCitations = warnings
 				.filter((w) => w.type === 'missingCitation')
 				.map((w) => w.id);
-
-			if (missingImages.length > 0 || missingCitations.length > 0) {
+			const pandocWarnings = pandocInfos.split('\n').map((info) => {
```
idreyn, line 189: React will complain that each of these children needs a `key` prop, which could just be the index from the second parameter of the map callback. Alternatively, though, you can accomplish the effect of splitting this at newlines by using the `white-space: pre` or `white-space: pre-wrap` (or maybe just `pre` given the width of the container) on an enclosing element.

client/containers/Pub/PubDocument/PubFileImport/FileImportDialog.tsx
```diff
@@ -221,6 +228,18 @@ const FileImportDialog = ({
 									</details>
 								</li>
 							)}
+							{pandocInfos && (
+								<li>
+									<i>
+										Conversion from LaTeX to HTML succeeded, with information (
+										<a href="https://help.pubpub.org/pub/2gfph6h8/release/2">
```
idreyn, line 235: Perhaps we should not link to a specific release (and maybe also give this Pub a nice slug?)

idreyn (changes requested):
A welcome and long overdue addition! I have a few requested changes but nothing major.

client/containers/Pub/PubDocument/PubFileImport/FileImportDialog.tsx
gabestein: wow how did I not know pre was a thing?!

idreyn (commented):
LGTM!

idreyn (approved):

PR #1284, opened 2021-02-16
fix: Allow long sticky TOC to scroll

### LINKED ISSUE #1241: Extra long TOCs unscrollable

1. Open long TOC, can't scroll to bottom!
...

Can scroll to bottom

https://openscientist.pubpub.org/pub/play/release/1#virtue-and-values-of-open-science

Any!

===

idreyn:
Resolves #1241

Some very long Pubs like [this one](https://openscientist.pubpub.org/pub/play/) have tables of contents which overflow from the `PubHeaderSticky` which appears below the fold, making it impossible to select its lower elements. This PR fixes that.

_Before:_
Screen Shot 2021-02-16 at 8 28 16 AM

_After:_
Screen Shot 2021-02-16 at 8 28 24 AM

_Test plan:_
Visit a Pub with a loooong TOC. I ran:

```
PUBPUB_PRODUCTION=true PUBPUB_LOCAL_COMMUNITY=openscientist
```

and then navigated to localhost:9876/pub/play. I observed that I was able to scroll the `PubHeaderSticky` TOC (the one in the Pub header itself remains unchanged)

### DIFF
--- a/client/containers/Pub/PubHeader/PubHeaderSticky.tsx
+++ b/client/containers/Pub/PubHeader/PubHeaderSticky.tsx
@@ -22,10 +22,9 @@ const PubHeaderSticky = (props: Props) => {
 			<div className="sticky-buttons">
 				{pubHeadings.length > 0 && (
 					<React.Fragment>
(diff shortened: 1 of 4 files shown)


isTravis (approved):
Great! Works for me and couldn't find any adverse side effects playing around with it.

PR #1296, opened 2021-02-22
dev: Add types and a few new utilities to arrays.ts

idreyn:
This quick PR was pulled out of dashboard perf work. I (finally) added type definitions to `arrays.ts` and snuck in a few utilities that will help us later.


isTravis (approved):
👍

PR #1297, opened 2021-02-22
dev: Types for apiFetch

idreyn:
Another bit of refactoring pulled out of dashboard perf changes — this adds type definitions for `apiFetch` and its method-specific helpers (`apiFetch.post` etc). This cuts a lot of `@ts-expect-error` comments from the client, which accounts for the rest of the noise in this PR.


isTravis (approved):
Awesome - looks clean!

PR #1298, opened 2021-02-22
dev: Add utilities to query, sort, and filter many Pubs

### LINKED ISSUE #1289: Write queryManyPubs utility
- Filter by collection membership, scoped IDs, excluded IDs
- Order by publication date, creation date, collection rank ASC/DESC
- Limit and offset

idreyn:
Resolves #1289 

This PR addresses the core need of our dashboard performance work — we need a way to ask the server for a list of Pubs with certain properties without actually loading those Pubs from the database. We will accomplish this with the included `queryPubIds` function, which builds a large SQL query for Pubs filtered on these parameters:

- Scoping and exclusion based on a list of Pub IDs
- Collection membership
- Released status

As well as the ability to sort Pubs by these fields:

- Publication date
- Creation date
- Update date (of the draft and Pub model, as a heuristic for recent activity)
- Collection order

Plus the ability to offset and limit these queries, allowing for pagination. To prevent this from turning into string soup I've brought in the `knex` query builder to add some guardrails on the SQL that we produce. I've been having a great experience with it so far.

_Test plan:_
Nothing yet. I may go back and add some unit tests for `queryPubIds`, but the PR is reviewable as is.


isTravis (approved):
Nice - I haven't used `knex` before, but it looks super useful. Not sure I wanna imagine what this file would look like without it.

`queryMany.ts` reads cleanly - though not having a list of pubs to test against, I'm not super confident I can vouch for the logic of it. I'll make a note to test that when testing the overall feature with UX, etc.

PR #1299, opened 2021-02-22
update: Load Pubs on demand into Layout editor

### LINKED ISSUE #1290: Update LayoutEditor to stream Pubs from /api/layouts instead of front-loading

idreyn:
Resolves #1290 

This PR is basically plumbing leading from the `queryPubIds` utility introduced in #1298 to the Page/Collection layout editor. We introduce a new API route, `/api/layout`, which accepts a list of layout blocks and a list of already-acquired Pubs, and it returns the Pub IDs used to build this layout, and any associated Pubs the client doesn't already have. These are gated by the usual Pub-sanitizing process, so it may give different results to different users.

This returns us to the desirable state where the Layout editor is really WYSIWYG, since it uses the exact same query mechanism as the backend, with practically no frontend manipulation of Pubs.

_Test plan:_
Don't even bother looking at this one — the subsequent PR (I will guess #1300) will have these changes plus a new and improved UX for Pub blocks.


isTravis (approved):
> Don't even bother looking at this one

Done! Appreciate looking over the structure though. The final nail in the "let's just send all pubs down for every Page to layout" coffin.

Issue #1292, opened 2021-02-22
Update PubSelect to use queryManyPubs


idreyn:
For use wherever we currently need to have a list of all Pubs on page load, including:

- The "pinned Pubs" dropdown in layout editor
- The Collection dashboard overview tab

PR #1300, opened 2021-02-22
update: Pubs block editor design refresh

### LINKED ISSUE #1291: Re-style LayoutEditorPubs controls to fit and flow better in container

idreyn:
Resolves #1291 

This PR updates the layout of the various controls used to query and style Pubs in the Pubs block editor. I made these changes because the layout was becoming very busy, and looked sort of overgrown especially at narrower widths. There's a bit of a convenience tradeoff in hiding some of these options behind dropdowns, but I am fairly happy with the compromise here.

https://user-images.githubusercontent.com/2208769/108775559-394a1080-752f-11eb-9297-f02f93b5cbcc.mov

_Test plan:_
This would be a good time to start testing this out for real with new and existing layouts across communities. Try:

- Adding Pubs blocks
- Re-ordering Pubs blocks
- Removing Pubs blocks
- Changing the query/styling options in the new Pubs block editor and making sure they all work

Note that the Pinned Pubs control currently only shows you a subset of Pubs that happen to be lying around. This is a consequence of #1299, and will be addressed when #1292 is finished and the results are incorporated into this control. I may even try to fix that weird dragging offset bug!


client/components/LayoutEditor/LayoutEditorPubs/LimitPubs.tsx
```diff
@@ -0,0 +1,64 @@
...
+		}
+	};
+
+	return (
+		<div className="layout-editor-pubs_limit-pubs-component">
+			<RadioGroup
+				onChange={(evt) => handleRadioSelect(evt.currentTarget.value)}
+				selectedValue={limit ? 'limit' : 'all'}
+			>
+				<Radio value="all" label={`Show up to ${maxPubsPerBlock} Pubs`} />
```
isTravis, line 40: Makes sense to keep the `all` vs `limit` separation since that's how we model it at the moment. I think a language update could make it feel less redundant though: "Show max number of pubs (100)".

isTravis (approved):
Such an improvement! Really looking forward to these updates going live.

One small comment on copy - but otherwise functions as expected when I tested.

PR #1311, opened 2021-03-12
Refactor getResizedUrl function and calls to use new resizer

### LINKED ISSUE #1257: Update resizer to v5
On April 1, AWS's execution environment will be changing in a way that effects our CloudFormation-orchestrated image resizer app. It will need to be upgraded from V2 to V5, which includes breaking changes and will require an update to PubPub's codebase.

isTravis:
This PR updates our Image Resizer stack to use version [5.2 of the Serverless Image Handler](https://github.com/awslabs/serverless-image-handler) and resolves #1257. 

The new Image Resizer lives as `resizer-v3.pubpub.org` and has a different API given that it's based on [Sharp](https://sharp.pixelplumbing.com/) rather than [Thumbor](http://thumbor.org). I find the documentation and formatting of Sharp much more modern and ergonomic - so that's a plus! The PR refactors the `getImageResizer` function to use the new API and also applies the slight modification to all invocations of `getImageResizer`. The main update is in `utils/images.ts` and most of the other changes are just function call refactors.

I'm hopeful that the more modern library also improves some of the compression issues we had previously been having. One positive sign in that direction is that the color performance seems to be much better in certain cases:

Frame 1

Like our previous stack, Sharp does not support .gif resizing without a custom compilation that includes a few additional libraries. There seems to be decent documentation on doing that if we eventually want to add that functionality ([Sharp documentation](https://sharp.pixelplumbing.com/install#custom-libvips), [AWS documentation](https://github.com/awslabs/serverless-image-handler)).

You can play with the library's parameters and see a preview of the request body in [this demo](https://d3g904gs8h9bzj.cloudfront.net/index.html). You can use `assets.pubpub.org` as the bucket name and `g6k2yuu3/71567617067103.jpg` as the image key.

If we merge and everything seems to be working well after a couple days, I'll spin down the previous resizer's CloudFormation and remove its DNS records. It may feel a bit slower for the first couple days as the resizer cache will be built back up from scratch. 

Test plan:
Upload a .jpg (or .jpeg) or .png and verify that resizing behaves as expected and with sufficient quality.

### DIFF
--- a/utils/images.ts
+++ b/utils/images.ts
@@ -1,8 +1,20 @@
 import { Maybe } from './types';
+import { btoaUniversal } from './strings';
 
-const validExtensions = ['jpg', 'jpeg', 'png', 'gif'];
+type validFits = 'cover' | 'contain' | 'fill' | 'inside' | 'outside';
 
-export const getResizedUrl = (url: Maybe<string>, type: string | null, dimensions: string) => {
+const validExtensions = ['jpg', 'jpeg', 'png'];
+
+export const getResizedUrl = (
+	url: Maybe<string>,
+	fit: validFits,
+	width?: number,
+	height?: number,
+) => {
+	/* Our resizer uses the AWS Serverless Image Handler CloudFormation */
+	/* (https://github.com/awslabs/serverless-image-handler). 			*/
+	/* That Handler uses sharp (https://sharp.pixelplumbing.com) as the */
+	/* underlying engine and thus relevant documentation source 		*/
 	if (!url || url.indexOf('https://assets.pubpub.org/') === -1) {
 		return url || '';
 	}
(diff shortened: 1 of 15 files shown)


utils/images.ts
```diff
@@ -1,8 +1,20 @@
 import { Maybe } from './types';
+import { btoaUniversal } from './strings';
 
-const validExtensions = ['jpg', 'jpeg', 'png', 'gif'];
+type validFits = 'cover' | 'contain' | 'fill' | 'inside' | 'outside';
```
idreyn, line 4: Nit: capitalized and singular (`Fit`)

idreyn (approved):
Looks good to me! Thanks for setting this up.

utils/images.ts
isTravis: Ah! Good catch. The `validExtensions` array right above tricked me!

Issue #1123, opened 2020-11-05
Eliminate FormattingBarLegacy


gabestein:
In Pages and Discussion editor

Occasional source of bugs/confusion

PR #1320, opened 2021-03-22
dev: Clean up Prosemirror command API for FormattingBar

idreyn:
In preparation for #1123 and an impending PR that adds text alignment commands to our Prosemirror editors, here is some cleanup work around rich text formatting and Prosemirror commands.

Most changes are related to how the `FormattingBar` gets information about actions it can take:
- The `onInsert` properties of most node specs are copies of the same boilerplate and these have been removed in favor of a generic `insertNodeIntoEditor` function, which falls back to`onInsert` if it is needed in a special case. This also allows us to remove the `insertFunctions` list in `EditorChangeObject`.
- The list of `menuItems` that was previously generated during `EditorChangeObject` generation was also ripe for an update. This list had lots of duplication, was re-generated from scratch on every transaction, and had only a weak association with the `FormattingBar` based on matching string keys. There is now a `commands` directory within the Editor package that abstracts over these menu items in a way that TypeScript can reason about statically.
- Usages of `menuItems` and `insertFunctions` across the React UI have been replaced with a new `useCommandStates` hook which takes a set of commands from the `commands` directory to watch, and provides their current state (`{ isActive, canRun, run }`) as a function of editor state. Moving this logic out of the `onChange` plugin and into React will help us manage performance since these computations can now be debounced or simply ignored when they are not required.
- `FormattingBar` and `FormattingBar/CommandMenu` now make use of `useCommandState`. The custom `PubInlineMenu` logic has mostly been removed and replaced with a generic `FormattingBar` instance.
- The heavy-duty custom hooks in `FormattingBar.tsx` have been moved to their own files. ~~The call site to `usePendingAttrs` has been moved to the child component `FormattingBarPopover` as an optimization.~~ _(this is breaking on some node types so I'm reverting it for now)_

A few other changes will make it easier to replace `FormattingBarLegacy` across PubPub:
- Information about `NodeLabels` is moved to a new plugin that abstracts it away from Pubs — so we no longer need to have a `PubData` in scope to mount a `FormattingBar`
- Other references to Pubs specifically should be removed from helper text in `FormattingBar`.

_Test plan:_
Check that
- All of the Pub formatting bar buttons work
- All of the floating table menu commands work
- Adding and manipulating media nodes still works as expected
- The `PubInlineMenu` that appears when selecting text, and all of its buttons, still work


isTravis (changes requested):
This is wonderful! Beyond the typescript wonders, moving a lot of every-cycle logic into a more performant hook is gunna make a big difference. Good thinking!

Most formatting bar options and table menus  worked as expected. I did find two things that seem to need fixing though:
- Strikethrough doesn't seem to work
- I can't get the expanded menu to stay open for inline items (e.g. inline equation, footnote, citation). I can get it to stay open for block nodes, and sometimes when clicking from a block node to an inline node, it'll stay open, but otherwise can't find a way (click, key over, etc) to get inline node menus to stay open.
LV3foXHyXk

idreyn:
> Strikethrough doesn't seem to work

Ah, I wrote `strikethrough` instead of `strike`

> I can't get the expanded menu to stay open for inline items...

Me neither...I guess moving `usePendingAttrs` broke it, so I'll just revert that change.

isTravis (approved):
Looking good - working on my end now!

PR #1321, opened 2021-03-22
feature: Text alignment in Prosemirror content

idreyn:
Since I am migrating the Layout editor away from `FormattingBarLegacy`, and the rich text Layout block editor has its own sort of hacked-in text alignment option, I thought it would be neat to try to support text alignment across all PubPub content. So, here it is. Paragraphs, headers, and lists can all be centered. The document body, blockquotes, and table cells all act as containers whose immediate children may be aligned — so we don't end up with contradicting alignment rules between parent nodes and their children.

_Screenshot:_
image

_Test plan:_
- Ensure that the following elements can be aligned and re-aligned: paragraphs, headings, lists
- Ensure that table cells and table heading cells have their own alignment separate from that of the document
- Ensure that the alignment buttons in the formatting bar work properly: if all of the currently selected content shares an alignment, that alignment is highlighted
- Ensure that alignment is carried through copy+paste operations if you select an entire paragraph or other alignable node (this will not work if your selection is entirely contained within the paragraph)
- Ensure that alignment works in PDF and HTML exports


isTravis (approved):
Really nice! 

I've been scratching my head trying to remember why we didn't have alignment like this much earlier. I had some vague sense that there was a blocking reason, but maybe that's coming from the days of PubPub using markdown. Everything works well and I think this will be a welcome option for many (not to mention a much simpler approach for Pages). 

I wasn't sure about the desired functionality, but I noticed that when you switch types (e.g. from paragraph to header, or list to paragraph, header to paragraph, etc) it automatically assumes a left-alignment, rather than using the alignment of the previous block-type. 

Google Docs and Word both appear to keep alignment setting in tact as the type is changed, so if there's a quick way to preserve that, that may be the expected behavior.

idreyn:
> I wasn't sure about the desired functionality, but I noticed that when you switch types (e.g. from paragraph to header, or list to paragraph, header to paragraph, etc) it automatically assumes a left-alignment, rather than using the alignment of the previous block-type.

Ideally I think we'd be able to persist alignment across block types, but the Prosemirror command API makes that a little tricky. Going from paragraph to paragraph is (sorta) simple because you can piggyback the `splitBlock` command that is run when you press enter to create a new paragraph off the old one...I'm certain there's a way to do this across block types, but less sure it's worth the effort since folks aren't making bake sale flyers in PubPub (yet) and probably don't need/want to center-align the entire document.

PR #1322, opened 2021-03-22
bug: permit comments with image but no text

howebenhowe:
Comments without text but with an image triggered the 'post button' to be disabled. No longer.

_Test Plan_
create and submit a comment to any pub with an image and no text

### DIFF
--- a/client/components/Editor/utils/view.ts
+++ b/client/components/Editor/utils/view.ts
@@ -20,6 +20,14 @@ export const getJSON = (editorView) => {
 export const getTextFromDoc = (doc: Node, separator = '\n') => {
 	return doc.textBetween(0, doc.nodeSize - 2, separator);
 };
+
+export const getImages = (editorView) => {
+	if (!editorView || !getJSON(editorView)) {
+		return [];
+	}
+	return getJSON(editorView).content.filter((con) => con.type === 'image');
+};
+
 export const getText = (editorView, separator = '\n') => {
 	if (!editorView) {
 		return null;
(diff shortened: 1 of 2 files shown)


client/components/Editor/utils/view.ts
```diff
@@ -20,6 +20,14 @@ export const getJSON = (editorView) => {
 export const getTextFromDoc = (doc: Node, separator = '\n') => {
 	return doc.textBetween(0, doc.nodeSize - 2, separator);
 };
+
+export const getImages = (editorView) => {
+	if (!editorView || !getJSON(editorView)) {
+		return [];
+	}
+	return getJSON(editorView).content.filter((con) => con.type === 'image');
```
idreyn, line 28: Let's call `getJSON` once and store the result in a variable.

client/components/Editor/utils/view.ts
```diff
@@ -20,6 +20,14 @@ export const getJSON = (editorView) => {
 export const getTextFromDoc = (doc: Node, separator = '\n') => {
 	return doc.textBetween(0, doc.nodeSize - 2, separator);
 };
+
+export const getImages = (editorView) => {
```
idreyn, line 24: I would name this something like `getTopLevelImages`, since images may appear more deeply nested inside the Prosemirror document, and this will not pick those up.

idreyn (changes requested):
Looks mostly good to me — just a few requests in the comments.

idreyn (approved):
LGTM 👍

Issue #1274, opened 2021-02-09
Add srcset to images across PubPub


gabestein:
Two options:

1. Solve only for pixel density
2. Also solve for loading image widths appropriate to browser width

PR #1324, opened 2021-03-24
feature: Enable srcset for pub images

howebenhowe:
addresses #1274
Generate a srcset for resized images in pubs to accommodate devices of any pixel density.
Does not address banners and instances where resized images are applied as background-image via css.

_Test plan_
+ In inspector, enable devices of at least two different pixel densities (e.g., Chrome inspector's emulated 'laptop with HiDPI' and 'laptop with MDPI')
+ disable cache in inspector
+ with inspector open and lower DPI device activated, navigate to any pub with an image in pub body
+ note image size in bytes
+ switch emulated device to HiDPI device
+ reload page
+ note + compare image size, HiDPI size should be greater


idreyn (approved):
Tried it out and it worked great for me. Lots of people are going to be very happy to see this!

PR #1327, opened 2021-03-26
dev: Attack and dethrone FormattingBarLegacy

idreyn:
Although we upgraded to a new FormattingBar for the Pub editor about a year ago, there are still a number of places where we still use the old version. This PR replaces all of those instances with the new formatting bar. I have also made a number of vaguely editor-related improvements:

- Pressing <kbd>Enter</kbd> will now only open the indicated node controls popover if the editor view is focused (annoying otherwise when you're typing into a comment box with the editor open)
- We changed the `toDOM` definition of `block_equation` last year; I have updated the definition of `parseDOM` to match it so that equations can be copy+pasted (and render correctly in caption fields)
- The media caption input scrolls correctly again instead of overflowing the FormattingBarPopover
- `Editor` now takes a `debounceEditsMs` prop that debounces `onEdit` callbacks for tight coupling to React without slow re-renders on every ProseMirror transaction.

While I was hoping we'd be able to get the full suite of media controls embedded in the discussion editor, there's just too much design work that needs to be done there, so I backed off for now.

_Screeshots:_
In the Layout editor:
image

In the ThreadComment editor:
Screen Shot 2021-03-26 at 4 04 54 PM

In the Review UI:
image

_Test plan:_
- Check that the formatting bar works in the following places: Pub, Text layout block, Discussion input, ThreadComment editor, ThreadInput (in the Review interface). 
- Check that the embedded formatting bar inside of the image controls caption editor works
- Check that those formatting bars can put up their own nested popovers (e.g. by adding an equation to an image caption). 
- Check that equations in captions render correctly when the popover is closed and then opened again
- Check the positioning of floating FormattingBar controls, like the link editor (in every context) and the table controls (in the Pub and Layout editing contexts)


isTravis (approved):
Looks great!

Couldn't find any issues working through the test cases. So awesome to see all the LegacyFormattingBar files deleted finally!

PR #1335, opened 2021-04-01
fix: Use display:contents on li > p to fix text-algined lists

idreyn:
We want center-aligned lists to look like this:

```
     - a
     - b
     - c
```

and not like this:

```
-      a
-      b
-      c
```

Our first few attempts:

1. Setting `display: inline-block` on `li > p`. This turned out not to work in Safari, and also created problems with bullet placement on tall `li` elements.
2. Setting `list-style-position: inside` on `ul, ol`. This caused multi-line `li` content to appear on a line beneath the bullet.

Ultimately, the problem is that our Prosemirror schema, all text content is contained within paragraphs, but `ul > li > p` adds an extra box within the list item that we could do without. So, our third attempt: use `display: contents` on `li > p` to remove the `p` from the box model entirely. Here is what this looks like:
image

Note that the final list item actually contains two paragraphs.

_Test plan:_
Inside Pubs and Page rich-text blocks, create ordered and unordered lists with every text alignment and check that they look okay. Pay attention to the behavior of text within an `li` element that spans multiple lines.

### DIFF
--- a/client/components/Editor/styles/main.scss
+++ b/client/components/Editor/styles/main.scss
@@ -33,6 +33,12 @@ h6 {
 	}
 }
 
+li {
+	> p {
+		display: contents;
+	}
+}
+
 .footnote {
 	vertical-align: super;
 	font-size: 0.85em;
--- a/client/containers/Pub/PubDocument/pubBody.scss
+++ b/client/containers/Pub/PubDocument/pubBody.scss
@@ -82,15 +82,11 @@ $pub-body-font-size: if-is-export(14px, 20px);
 		font-style: normal;
 		line-height: 1.7;
 		letter-spacing: -0.003em;
-		margin: 0em 0em 1em;
 		word-break: break-word;
 	}
 
-	li {
-		margin: 0em;
-		& > p {
-			margin: 0;
-		}
+	p {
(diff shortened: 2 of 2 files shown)


isTravis:
Was going to leave two quick notes:
1. The line height seems to change (and thus the position of the bullet) when a list is empty vs has content.
2. Google docs and Word allow alignment per list-item, rather than per-list. 

Neither of these are as problematic as the line-wrap issue, so good to get this up quickly. But if there are quick css fixes for those, I'd advocate for them!

idreyn:
> The line height seems to change (and thus the position of the bullet) when a list is empty vs has content.

Ah, I actually thought I fixed that here but I ended up leaving out the fix (styling `li::marker` directly to change its size). I think `::marker` has come into more general use since we decided against targeting it a year or two ago, so I'll add that in.

> Google docs and Word allow alignment per list-item, rather than per-list.

That's wild! I'm sure we could enable that but it's definitely out of scope for now.

> The left-padding seems to still exist on lists, leaving the left edge of the document not aligned.

Per our Slack discussion I'm going to restore `padding-left: 1em`. Looks much better that way!

PR #1344, opened 2021-04-09
update: Overhaul Community and Collection overviews with new design and Pub lazy-loading

idreyn:
This PR re-implements the Community and Collection dashboard overviews to use Deepak's new designs, and with the ability to lazy-load more Pubs from the client instead of needing them all up front. For very large communities this will make the dashboard usable, and — along with a few other tweaks — this should allow us to return to serving PubPub from Heroku 2x dynos without worrying about memory pressure from extremely large query payloads.

What's **not** here yet, that's still on my to-do list:
- Scope summaries ("x Pubs, y Collections, z Discussions") in several places. These need to be re-implemented as a pre-cached/materialized computation since they can no longer be accurately computed from the data loaded with the page.
- The recently viewed Pubs section
- Some sorting and filtering options. These are still a little under-defined and I think we can launch without them. I do think a way to see recently updated Pubs would be especially useful, though. 
- Incorporating author information (and other things?) into search
- A final decision on the search UX — currently you have to press enter to search. I'm hitting performance problems with search-as-you-type, and I'm not convinced that it's a better experience anyway, but there's also something missing with the current approach
- Some design hints, especially on Collection overviews — we no longer mention the kind of a Collection on its overview, which is a little silly. I may try adapting the sidebar from the Community overview to hold some of this information
- Totally complete mobile styling — this might as well wait for a focused effort on mobile dashboard styles, since the Dashboard is functionally unusable on mobile until we collapse the left-hand menu somehow.

All that aside, I think this is completely usable, is an improvement over what we currently have, and we could consider shipping this in its current state.

_Test plan:_
On a Community overview (good to try on a community like `baas` with tons of Pubs), make sure that:
- Pubs load and are alphabetized, are clickable, etc.
- Collection rows can be disclosed and their header sticks properly
- You can explicitly load more Pubs from a disclosed Collection row
- Lazy-loading works by scrolling to the bottom of the page
- Searching and filtering by type works

On a Collection (non-Tag) overview, make sure that:
- Pubs can be added, removed, and re-ordered
- The primary and hint controls on the right side of Pub rows work as before
- More Pubs can be lazy-loaded by scrolling down

_New Community overview:_
Screen Shot 2021-04-12 at 1 49 42 PM

_Disclosed Collection in Community overview:_
Screen Shot 2021-04-12 at 1 50 12 PM

_Searching in Community overview:_

Screen Shot 2021-04-12 at 1 50 26 PM

_New Collection overview:_
Screen Shot 2021-04-12 at 1 50 36 PM


client/containers/DashboardOverview/CollectionOverview/DashboardCollectionOverview.tsx
```diff
@@ -0,0 +1,178 @@
...
+		if (pubs.length === 0 && hasLoadedAllPubs) {
+			if (isSearching) {
+				return <SpecialRow>No matching Pubs.</SpecialRow>;
+			}
+			return (
+				<NonIdealState
+					className="collection-empty-state"
+					icon="clean"
+					title="This Collection looks brand new!"
+					description="Try creating a Pub or Collection from above."
```
isTravis, line 137: This default language seems a bit misleading. Perhaps just, "Try creating a Pub using the button above"

isTravis (approved):
Still play-testing now on the call :) 

But code reformats and high-level scan of code review looks clean. Did my best to skip over files that looked like refactors or typescript edits, but I've almost certainly missed some. 

Good on my end to push up and wait for any error reports.

client/containers/DashboardOverview/CollectionOverview/DashboardCollectionOverview.tsx
idreyn: Oh! Just copied that from the Community side. Nice catch.

Issue #1067, opened 2020-09-30
Citation upgrades


ThomasJorna:
**What's your enhancement request?**
Two things:
-Sort citations by author last name instead of by appearance. 
-Set multiple "standard" inline citation styles, e.g. (With Brackets, 2020) and Without Brackets (2020)

**What would this enhancement help you accomplish?**

Our journal uses APA-style for references, which requires alphabetically sorted bibliographies. I quite like PubPub's built in citation system as it gives you a pop-up showing you the full citation on hover, but we cannot use it at the moments since the bibliography/citations aren't formatted correctly. 
The second issue could be fixed manually, but that makes adding citations an even bigger hassle. One make this easier could be to have the selected text be automatically used as the custom citation input. This would be especially useful to us, as most of our authors prefer not to write their article in PubPub.

---

**Any attachments or reference images? (Drag them here)**
![image](https://user-images.githubusercontent.com/21983833/94733636-5d326d00-0368-11eb-8bee-9f3792ac407b.png)
Ideally I could have it look like this in the citations list.

Suggestion for citation quality of life changes:
![image](https://user-images.githubusercontent.com/21983833/94733857-b1d5e800-0368-11eb-99a5-70089558b59b.png)
![image](https://user-images.githubusercontent.com/21983833/94733945-cade9900-0368-11eb-9465-6db56713dbed.png)

**What's the URL to your PubPub community?**
https://jtrialerror.pubpub.org
**Who requested this? (If not you)***

ThomasJorna:
A further illustration of the specific interaction I'm looking for:

This is how PubPub interprets LaTeX's \textcite{Collins1972, Collins etc...}, which "should" be rendered as Collins (1972, 19...) but instead is interpreted as \parencite{Collins1972}\parencite{Collins...} and thus rendered as 4 separate citations, rendering it pretty much unusable for more than 1 citation.
![image](https://user-images.githubusercontent.com/21983833/94931167-b95fd300-04c7-11eb-9dee-e68a89bff192.png)

On a positive note: it's fantastic that PubPub is able to convert a tex and bib file to such nice citations so seemlessly, everything is accounted for!

gabestein:
Thanks, Thomas. I suspect you've seen this given your proposed solutions, but as a workaround for now, even though it can be time-consuming, you can use the custom citation text feature to strip out the extra parens.

[...]

PR #1349, opened 2021-04-12
Alphabetize citation list

howebenhowe:
Addresses #1067
After checking all citation styles we offer, it appears they all alphabetize by the entry itself, and typically that means the author's last name, though if no author is present, the title. Here we alphabetize the list, remove numbers from the citation list, and refactor the handling of citations across the editor.

Screenshot:
Screenshot from 2021-04-12 17-08-21

Test plan:

+ create [pub with broad variety of citations and footnotes](https://demo.duqduq.org/pub/ple0quof/draft)
  + structured value only footnote
  + unstructured value only footnote
  + both values footnote
  + error-producing structured value footnote
  + structured value only citation
  + unstructured value only citation
  + both values citation
  + error-producing structured value citation
+ ensure alphabetical sort is applied to text of the citation, whether that be the author's last name or the title's first word, as the case may be.


client/containers/Pub/PubDocument/PubBottom/Notes.tsx
```diff
@@ -74,13 +68,12 @@ const Note = (props: NoteProps) => {
...
 		<li className="note">
-			{/* @ts-expect-error ts-migrate(2339) FIXME: Property 'number' does not exist on type 'noteProp... Remove this comment to see the full error message */}
-			<div className="number">{note.number}.</div>
+			{note.number && <div className="number">{note.number}.</div>}
 			<div className="inner">
 				<PubNoteContent
 					ref={contentRef}
-					structured={citation && citation.html}
-					unstructured={note.unstructuredValue}
+					structured={linkifyHtml(note.renderedStructuredValue?.html)}
```
idreyn, line 75: ❗️ I think it's likely that we want the call to `linkifyHtml` to live inside the `PubNoteContent` component, so this behavior will get picked up in the `ControlsFootnoteCitation` as well. Might be worth wrapping these computations in `useMemo` as well.

client/containers/Pub/PubDocument/PubBottom/Notes.tsx
```diff
@@ -109,11 +102,11 @@ const Note = (props: NoteProps) => {
 
 const Notes = (props: NotesProps) => {
 	const { notes, ...restProps } = props;
+	const hasNumbering = notes.some(({ number }) => !!number);
 	return (
-		<ul className="notes-component">
+		<ul className={classNames('notes-component', !hasNumbering && 'no-padding')}>
```
idreyn, line 107: I'd suggest `has-numbering` instead. Helps when you're looking at the SCSS side of the code and you see:

```
.has-numbering { padding: 0; }
```
instead of 
```
.no-padding { padding: 0; }
```
and you get some context about why this class exists at all, without having to look elsewhere.

client/containers/Pub/PubDocument/PubBottom/PubBottom.tsx
```diff
@@ -40,14 +36,14 @@ const PubBottom = (props: Props) => {
 				{footnotes.length > 0 && (
 					<SearchableNoteSection
 						title="Footnotes"
-						items={footnotes}
+						notes={footnotes}
 						searchPlaceholder="Search footnotes..."
 					/>
 				)}
-				{citations.length > 0 && (
+				{Object.keys(citations).length > 0 && (
```
idreyn, line 43: ❗️ `citations` is still an array so this should probably be reverted.

client/containers/Pub/PubSyncManager.tsx
```diff
@@ -33,7 +33,7 @@ const shimPubContextProps = {
 	updateLocalData: null as any,
 	updatePubData: null as any,
 	// @ts-expect-error ts-migrate(2554) FIXME: Expected 2-3 arguments, but got 0.
-	citationManager: new CitationManager(),
+	noteManager: new NoteManager(),
```
idreyn, line 36: You might consider passing in the same defaults as in `usePageState` here so we can remove the `@ts-expect-error` comment. Runtime-wise it's no big deal since this instance should absolutely never be touched, since consumers will get the one created in the constructor below.

client/utils/notes.ts
```diff
@@ -0,0 +1,22 @@
...
+				method: 'POST',
+				body: JSON.stringify({
+					structuredValues,
+					inlineStyleKind,
+					citationStyleKind,
+				}),
+		  })
+		: Promise.resolve([]);
+
+export class NoteManager extends NoteManagerCore {
```
idreyn, line 18: Were you planning to merge this with `NoteManagerCore` since we no longer have a server/client split? I'm 100% okay leaving this as-is, it's a nice abstraction.

client/containers/Pub/pubHooks.ts
```diff
@@ -19,3 +22,43 @@ export const usePubData = () => {
...
+export const usePubNotes = () => {
+	const {
+		noteManager,
+		collabData: {
+			editorChangeObject: { view },
+		},
+	} = usePubContext();
+	const { citations = [], footnotes = [] } = view ? getNotes(view.state.doc) : {};
+	const [notes, setNotes] = useState(noteManager.notes);
+	useEffect(() => noteManager.subscribe(setNotes), [noteManager]);
```
idreyn, line 40: $0.02: I like to write `useEffect` calls after all of the variable declarations (except handlers and render helpers) inside of a component or hook, roughly in this order.

```
const Component = (props) => {
    const {x, y, z} = props
    const [a, setA] = useState(...)
    const [b, setB] = useState(...)
    const {q, r, s} = useContext(...)
    const w = fn(x, y, z);

    useEffect(...)
    useEffect(...)

    const handleX = useCallback(() => ...)

    const renderX = () => ...
    const renderY = () => ...

    return ...
}
```

So it ends up roooooughly in this order:

1. Variable declarations (except nested functions) including those that call hooks
2. Effect hooks that don't return a value (e.g. `useEffect`, our `useInfiniteScroll`)
3. Event handler declarations (e.g. `handleClick`)
4. Render helpers (e.g. `renderInnerText`)
5. The return value

You're not required to use this ordering, but I'm curious if you have any of your own headcanon conventions for how these things should be laid out. For instance, it makes sense that you'd have the `useState` and `useEffect` hooks related to `notes` in the same place — it's just not where I'd expect to find that `useEffect`.

client/containers/Pub/pubHooks.ts
```diff
@@ -19,3 +22,43 @@ export const usePubData = () => {
...
+	const isNumberList = noteManager.citationInlineStyleKind === 'count';
+	const renderedFootnotes = footnotes.map((footnote, index) => ({
+		...footnote,
+		...notes[footnote.structuredValue],
+		number: index + 1,
+	}));
+	const renderedCitations = citations
```
idreyn, line 47: I'd like to suggest writing this in a way that:

1. Doesn't trigger a `sort()` every render
2. Doesn't do any computation inside of the `sort()` body

The general idea would be:

1. Wrap the whole thing in `useMemo`. I'd expect this hook body to be run hundreds or thousands of times for every time this value _needs_ to be re-rendered (because `citations` or `notes`) changed, so this will really add up.
2. Pre-compute the values that you'll sort by, associated to the values you want to sort by `unstructuredValue`
3. Sort by those values

I remember that we talked about how it would be nice to have a lodash-style utility that lets you sort an array by a key (either a property name or a computed value from each object). If you're so inclined this could be a good place to create and use that utility, since it would accomplish steps 2 and 3.

Here's a sketch of what I have in mind (apologies if this doesn't run!)

```ts
const renderedCitations = useMemo(() => {
    const unsortedRenderedCitations = citations.map((citation, index) => ({
        ...citation,
        ...notes[citation.structuredValue],
        ...(isNumberList && { number: index + 1 }),
    }));

    // You could also use a Map here and hold the rendered citations themselves
    // instead of their structured values.
    const sortableValuesByStructuredValue: Record<string, string> = {};
    unsortedRenderedCitations.forEach(citation => {
        sortableValuesByStructuredValue[citation.structuredValue] = getRenderedValue(citation)
            .replace(/(<([^>]+)>)/gi, '')
            .toLowerCase()
            .trim();
    });

    return unsortedRenderedCitations.sort((a, b) => {
        const preppedA = sortableValuesByStructuredValue[a.structuredValue];
        const preppedB = sortableValuesByStructuredValue[b.structuredValue];
        return preppedA > preppedB ? 1 : -1;
    });
}, [citations, notes]);
```

utils/citations.ts
```diff
@@ -1,9 +1,30 @@
+export type CitationStyleKind =
```
idreyn, line 1: Thank you so much for adding these types!

idreyn (changes requested):
Thanks for all your hard work on this, and for the various bits of cleanup and improvement along the way. I have a few things I'd like you to address (marked with ❗️), otherwise this looks about ready to go!

idreyn (approved):
LGTM!

PR #1353, opened 2021-04-14
feature: Custom CSS and JS for Communities

idreyn:
Though we're planning to roll this out verrrry slowly, one Community needs this functionality starting next week so I went ahead and added it today. This is a bit of an MVP and we can always revisit some of these design decisions:

- Custom scripts are stored directly in the database and inlined with the request rather than stored as a separate asset in S3. I think we'll probably want to change this.
- We're using the `@monaco-editor/react` editor as a code editor here instead of CodeMirror or something else.
- Right now this is only suitable for use in Communities that we trust. For untrusted Communities we'll need to do some heavy-duty CSS sanitization, and I sort of doubt we'll ever want to turn on custom JS by default.

_Test plan:_
I wrote a few basic tests:
```
npm run test-dev server/customScript
```
To actually test-drive this, make sure you're in the `demo` Community where this is enabled, and then:

- Visit `/dash/scripts`
- Add and save some JS and CSS (I know, I want meta+S keybindings too)
- Refresh the page and make what you added is still there
- Then visit a Page and a Collection layout on the Community and make sure that the custom scripts are working!

_Screenshots:_

The editor:
Screen Shot 2021-04-14 at 3 23 04 PM

_The result:_
Screen Shot 2021-04-14 at 3 32 22 PM

### DIFF
--- a/client/containers/App/paths.ts
+++ b/client/containers/App/paths.ts
@@ -7,6 +7,7 @@ import {
 	DashboardCollectionOverview,
 	DashboardPubOverview,
 	DashboardCollectionLayout,
+	DashboardCustomScripts,
 	DashboardDiscussions,
 	DashboardEdges,
 	DashboardImpact,
@@ -102,6 +103,10 @@ export default (viewData, locationData, chunkName) => {
 		},
 		DashboardCollectionLayout: {
 			ActiveComponent: DashboardCollectionLayout,
+			isDashboard: true,
+		},
+		DashboardCustomScripts: {
+			ActiveComponent: DashboardCustomScripts,
 			isDashboard: true,
 		},
 		Explore: {
--- a/client/containers/DashboardCustomScripts/CustomScriptPanel.tsx
+++ b/client/containers/DashboardCustomScripts/CustomScriptPanel.tsx
@@ -1,0 +1,63 @@
+import React, { useState } from 'react';
+import { useBeforeUnload } from 'react-use';
+import { Button } from '@blueprintjs/core';
+
+import { apiFetch } from 'client/utils/apiFetch';
+import { CustomScriptType } from 'utils/types';
+import { usePageContext } from 'utils/hooks';
+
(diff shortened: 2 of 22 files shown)


isTravis (changes requested):
Nice job getting this together so quickly! The CSS side is going to be such a welcome site for those willing to take the risks.

I'm still not convinced we ever want to allow custom JS. Happy to discuss it further and explore the idea after this PR lands - but just stating that opinion explicitly.

For this PR, trust isn't quite enough to let an external community write custom JS. Perhaps an explicit flag on the `CustomScript` model that is set `false` on every saved edit, requiring someone on the dev team to approve and flip the flag before it gets injected into `Html.js`. It's too easy for a user to unintentionally include JS (e.g. a facebook widget) that creates a security risk or causes us to break the ToS contract we have with each user.

gabestein:
Tests out well for me!

This is gonna be a game-changer for a number of folks. Very exciting.

I'm with @isTravis, mostly. In particular, I think we'll want a way to enable css only, which a number of communities (sonic scope, invisible culture, sar, hdsr, etc.) could make good use of today.

As long as that's possible, I'm a little less concerned about the JS issue Travis brought up, because my guess is that in those rarer cases, we'll be managing most of the JS on behalf of folks for the time being, and the feature is pretty well hidden.

idreyn:
Good points all around. I agree that we'll want to modify this to provide a CSS-only option to Communities in general sometime soon. We do however need to spin up the custom JS feature for Contours (or quickly change direction on how we plan to support some of their requests), and I can take responsibility for auditing that code for security and privacy. My understanding is that they mostly want it to add some lightbox features to Page-level images.

> we'll be managing most of the JS on behalf of folks for the time being, and the feature is pretty well hidden

Just for clarity of discussion: the feature is not merely well-hidden; it is impossible to use for Communities other than the ones we've currently approved (just `demo` and `contours` at this point)

isTravis (approved):

PR #1355, opened 2021-04-16
update: give everyone all the cc licenses

gabestein:
Makes a long overdue change to give all communities all the CC licenses.

_Test plan_
- Visit a community without the premium flag
- Make sure they can now select from all 5 cc license options


isTravis (approved):
Looks good!

PR #1360, opened 2021-04-21
update: Add ScopeSummary model and calculations

### LINKED ISSUE #1294: Create ScopeContentSummary and associated queries/triggers
To be able to say how many Discussions/Reviews/Pubs etc are in a Community/Collection/Pub without loading them all up front. Not sure what the best mechanism to keep these up to date is quite yet.

### LINKED ISSUE #1359: Reviews tab no longer shows number of reviews
Must be related to the recent dash pushes.

1. Requested publication.
2. Review shows up on the review tab.
3. No number shows up in the "reviews" tab link indicating an open review.

A number shows up in the "reviews" tab link.

https://demo.pubpub.org/dash/reviews

Chrome Latest / Mac Latest

===

Screen Shot 2021-04-20 at 13 46 32

Scott / crimEmpire

idreyn:
We want to have a sense of how many elements each scope (`Community`, `Collection`, `Pub`) contains without having to calculate this information on the fly during requests. To accomplish this I've added a `ScopeSummary` model associated to each of the scope models above, which will hold this information. I'm making use of Sequelize lifecycle hooks (I believe for the first time in this codebase) to ensure that the summaries are updated when they need to be, i.e. during the following events:

- A Pub, Collection, or Community is created
- A Pub is added or removed from a Collection
- A Discussion or Review is added to a Pub

The real goal here is to add these summaries to the dashboard overview tabs, which I'll do in another PR. In the meantime, we can get this merged into master and have the migration backfill the summaries for existing content.

Resolves #1294
Resolves #1359

_Test plan:_
Run the unit tests:
```
npm run test-dev server/scopeSummary
```
Don't forget to also run the schema migration before deploying:
```
PUBPUB_PRODUCTION=true npm run tools migrate -- --name 2021_04_21_addScopeSummaryId
```
and then the backfill:
```
PUBPUB_PRODUCTION=true npm run tools migrate -- --name data/2021_04_21_backfillScopeSummaries
```


howebenhowe (approved):
Looks good to me.

isTravis (approved):
Looks good. Thanks for taking the scary first step of using hooks - they seem like a good fit here.

For context, I think one of the reasons we didn't use them elsewhere (e.g. updating Algolia on Pub update/creation) is that often those functions needed more data than was returned by the sequelize hook. We could query for that extra data from the hook, but if I recall correctly, we had it already in the API function, so it felt duplicative. 

The way you've broken out the hooks and queries here though make it feel cleaner than hoping you can re-use something that exists in the API query, so I could see us moving to this approach for all the side-effecty function calls we want to make.

PR #1362, opened 2021-04-21
feature: Enable filter/sort tabs in dashboard overview

idreyn:
This PR adds a set of tabs to the Community and Collection overview where we can put tabs to do sorting and filtering of Pubs. Since there's not really a release/draft concept with Collections (yet?) all tabs other than "All" will only show Pubs for the time being. In addition to "Draft" and "Released" from the mockups, I've added a "Latest" tab that I think folks may find useful.

The tabs appear to be taking up a frustrating amount of vertical space for now given their size; soon we'll add an "Explore" label per the design spec to balance things out.

I would kinda like to unify the `searchTerm` on each overview component with the `Partial<PubsQuery>` state that the `OverviewSearchGroup` now provides, but I can't quite find the right abstraction/am having a bad time with hooks.

_Test plan:_

Check that...
- The appropriate Pubs load on each of the tabs, and that lazy-loading works as expected (there are some issues with the Latest tab loading stuff in the correct order that you may encounter — I'm working it out)
- The Pubs/Collections controls are hidden/disabled as appropriate when only Pubs should be shown
- You can search within tabs
- Pubs in a Collection cannot be re-ordered when any tab other than "All" is selected.

_Screenshots:_

Community overview:
Screen Shot 2021-04-21 at 4 05 22 PM

Collection overview:
Screen Shot 2021-04-21 at 4 05 45 PM


client/containers/DashboardOverview/CommunityOverview/CommunityItems.tsx
```diff
@@ -86,7 +91,7 @@ const CommunityItems = (props: Props) => {
 				);
 			}
 			if (pubs.length === 0 && hasLoadedAllPubs) {
-				if (searchTerm.length > 0) {
+				if (isSearchingOrFiltering) {
 					return <SpecialRow>No matching Pubs or Collections.</SpecialRow>;
```
howebenhowe, line 95: Why not `if (pubs.length === 0 && hasLoadedAlPubs && isSearchingOrFiltering) `?

howebenhowe (commented):
Looks good to me with the one code style question.
Likewise, testing on qubqub checked out.

PR #1365, opened 2021-04-23
Cite and deposit quick fix

gabestein:
Simplifies the license check as suggest by @idreyn in #1365


idreyn (changes requested):
Thanks for this change. But we still need the `!licenseSlug` and `community.premiumLicenseFlag` checks!

gabestein:
d'oh, of course. Added back...

PR #1380, opened 2021-04-29
feature: Query Pubs by author name

idreyn:
A quick one to enable searching for Pubs by author name wherever we use `queryMany`, including in the dashboard overview and some assorted Pub selection widgets.

_Test plan:_
- From the Dashboard overview, try searching for Pubs by author name. Test for authors with PubPub accounts as well as "shadow authors" without accounts
- Check on the Pinned Pubs picked to make sure this also works here


howebenhowe (approved):
Looks good,  though some of this is hieroglyphics to me. I'll need to bone upon query building before I can provide meaningful review on

```
		if (term) {
			builder.whereRaw(
				'"title" ilike ? OR EXISTS (SELECT FROM unnest("authorNames") name WHERE name ilike ?)',
				[`%${term}%`, `%${term}%`],
			);
		}
```

PR #1388, opened 2021-05-03
update: Load Connections dashboard tab without all Pubs

### LINKED ISSUE #1358: Load connections dashboards async

idreyn:
This PR addresses one of the last places where we load all Pubs into scope on the Dashboard: the Connections tab.

_Test plan:_
Visit the Connections tab for a Pub and make sure:
- You can search for a Pub from within the Community and add a Connection
- You can also add a Connection by pasting in an external URL or DOI

Resolves #1358

### DIFF
--- a/client/containers/DashboardEdges/NewEdgeInput.tsx
+++ b/client/containers/DashboardEdges/NewEdgeInput.tsx
@@ -86,21 +86,19 @@ const NewEdgeInput = (props: Props) => {
 		if ('indeterminate' in item && item.indeterminate) {
 			return indeterminateMenuItem;
 		}
-		if ('targetPub' in item) {
+		if ('targetPub' in item && item.targetPub) {
 			const { targetPub } = item;
-			if (targetPub) {
-				return (
-					<PubMenuItem
-						key={targetPub.id}
-						title={targetPub.title}
-						contributors={targetPub.attributions.filter((attr) => attr.isAuthor)}
-						image={targetPub.avatar}
-						active={modifiers.active}
-						onClick={handleClick}
-						showImage={true}
-					/>
-				);
-			}
+			return (
+				<PubMenuItem
+					key={targetPub.id}
+					title={targetPub.title}
+					contributors={targetPub.attributions.filter((attr) => attr.isAuthor)}
+					image={targetPub.avatar}
+					active={modifiers.active}
+					onClick={handleClick}
+					showImage={true}
+				/>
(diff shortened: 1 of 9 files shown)


client/containers/DashboardEdges/NewEdgeInput.tsx
```diff
@@ -47,59 +51,63 @@ const indeterminateMenuItem = (
...
+						active={modifiers.active}
+						onClick={handleClick}
+						showImage={true}
+					/>
+				);
+			}
 		}
-		if (externalPublication && externalPublication.title) {
-			const { title, contributors, avatar } = externalPublication;
+		if (
```
howebenhowe, line 105: Is this condition mutually exclusive with the targetPub condition above? In that case, `else if`?
Given each if clause contains a return statement, they seem mutually exclusive, `else` makes the logic easier to track.

client/containers/DashboardEdges/NewEdgeInput.tsx
```diff
@@ -47,59 +51,63 @@ const indeterminateMenuItem = (
...
-					contributors={targetPub.attributions.filter((attr) => attr.isAuthor)}
-					image={targetPub.avatar}
-					active={modifiers.active}
-					onClick={handleClick}
-					showImage={true}
-				/>
-			);
+		if ('targetPub' in item) {
+			const { targetPub } = item;
+			if (targetPub) {
```
howebenhowe, line 91: Here the if-conditions are split into two statements, below they are chained with `&&`. Can we use the same approach in both places?

howebenhowe (approved):
Only style suggestions. Otherwise, looks good and behaves correctl.

client/containers/DashboardEdges/NewEdgeInput.tsx
idreyn: yup!

client/containers/DashboardEdges/NewEdgeInput.tsx
idreyn: I agree that this would potentially add clarity, but our linter actually has [a rule](https://eslint.org/docs/rules/no-else-return) that breaks `if/else if/else` clauses into disjoint `if` statements if every branch has a `return` statement and it can statically verify that they're mutually exclusive. It's a style that I've gotten used to, but if it bothers you, you have the go ahead to turn it off.

PR #1389, opened 2021-05-03
update: Load Reviews dashboard tab without all Pubs

idreyn:
This address the actual final place where we load all Pubs on each request: the Reviews tab.

_Test plan:_
For each of a (Community, Collection, Pub) visit the Reviews tab and ensure that all the scoped Reviews are visible.


howebenhowe (approved):
LGTM

PR #1391, opened 2021-05-03
fix: alphabetize structured citations

howebenhowe:
Fixes citations with structured values not sorting alphabetically in pub bottom citation list.

### DIFF
--- a/client/containers/Pub/usePubNotes.ts
+++ b/client/containers/Pub/usePubNotes.ts
@@ -30,9 +30,9 @@ export const usePubNotes = () => {
 	const renderedCitations = useMemo(
 		() =>
 			citations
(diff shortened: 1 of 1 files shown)


client/containers/Pub/usePubNotes.ts
```diff
@@ -30,6 +30,7 @@ export const usePubNotes = () => {
 	const renderedCitations = useMemo(
 		() =>
 			citations
+				.map((citation) => ({ ...citation, ...notes[citation.structuredValue] }))
 				.map((citation, index) => ({
 					...citation,
 					...notes[citation.structuredValue],
```
idreyn, line 36: Since you're spreading this value in above, this line can be removed.

idreyn (approved):
Nice!

PR #1394, opened 2021-05-04
dev: Split utils/types.ts into a number of smaller files

idreyn:
When we started using TypeScript I made a file called `utils/types.ts` that had a couple of useful type definitions in there. It's now a few hundred lines long and I'm looking to add a bunch more types for Activity, so I figured I'd split it up. The definitions now live in files within the `types/` directory, and are all exported at the top level...so anything you used to get from `utils/types` will now be accessible from `types`.

_Test plan:_
Make sure the devserver works:
```
npm start
```
...and the production server:
```
npm run build-dev-once && npm run api-prod
```
...and Storybook:
```
npm run storybook
```
...and the `tools` directory which still uses `ts-node`:
```
npm run tools devshell
```
...and the tests:
```
npm run test-dev
```


howebenhowe (approved):
Outstanding.

PR #1403, opened 2021-05-10
fix: Correctly chunk Pub data

### LINKED ISSUE #1339: Some pubs not properly chunked in algolia
1. Search for text in [this article](https://hdsr.mitpress.mit.edu/pub/kxks56er/release/4) on HDSR (e.g. "football")
2. Articles does not appear in search

It appears we're having issues with chunking in algolia.

Article appears in search.

https://hdsr.mitpress.mit.edu/search?q=football

===

idreyn:
Resolves #1339

We use Algolia to index Pub content for easy searching. Algolia has a 10kb size limit for each searchable record, so we break Pub content into many chunks and create one record for each chunk; these are later de-duplicated by `pubId` so users only see one record per Pub when searching. 

This PR fixes a bug I introduced in #1247 where we're actually inserting a single record with a `string[]` of chunks instead of many records with one `string` chunk each. Algolia is good at searching nested data so this doesn't really impair searching, but we do occasionally hit the 10kb ceiling on the total content size, preventing us from making the Pub searchable as noted in the issue above.

While debugging this I made a few other change which I think are worth keeping:

- Retiring our own text chunking code in favor of Algolia's `chunk-text` library
- Dropping the `stopwords` library; since we pass in our own list of stopwords, we are not using any of its functionality.

_Test plan:_

I opened the devshell:
 
```
PUBPUB_PRODUCTION=true npm run tools devshell
```

then I deposited search data for the Pub referenced above:

```js
pub = await Pub.findOne({where: {slug: "kxks56er"}});
await setPubSearchData(pub.id)
```

I observed that this completed without error. Then I checked the Algolia index and verified that this created records with stopwords removed as expected.


howebenhowe (approved):
LGTM

PR #1407, opened 2021-05-12
fix: Let FormattingBar controls float detached from sticky header 

### LINKED ISSUE #1354: Table formatting wonkiness

1. The table does not want to change
2. I fight with the table
3. When I highlight a column to merge cells within it, the page scrolls and the options menu is hard to navigate
...

https://recordit.co/iKqyqDf4yk

Chrome

===

https://recordit.co/iKqyqDf4yk

idreyn:
Resolves #1354 (kinda, sorta...)

In the bug report above we see an issue with the floating table controls (`ControlsTable`) component where, when its menu is opened near the bottom of the viewport and moused over, it will attempt to scroll the page down, creating a jittering effect. The reason it scrolls at all is down to a Chrome-specific implementation of `element.focus()`, which is outside our ability to resolve for now. But the jittering is due to the fact that the controls are mounted as a child of the `FormattingBar`, which is stuck to the top of the viewport as the Pub scrolls — and thus scrolling the page has no effect on the position of the menu.

We can solve the jittering by mounting the `ControlsTable` outside of the `FormattingBar`. There are provisions to do this built into the component, but the API has become pretty hard to understand at this point and many changes have unintended side effects. So the bulk of this PR is refactoring that will hopefully make it easier for readers to understand what's happening in different formatting bar configurations.

The changes at a glance:

- There is a new type called `ControlsConfiguration` which holds...configuration for how the controls should appear. Another type, `ResolvedControlsConfiguration`, extends this first one and is what gets passed out of `useControlsState`. This gives a name to a bundle of conceptually related values that float around this code and reduces the number of props we have to pass back and forth, in addition to making it easier to understand which states it's possible for the formatting bar to be in.

- One new degree of freedom in `ControlsConfiguration` is the `isAbsolutelyPositioned` attribute. This tells the positioning code whether the controls should have `position: absolute`. The ability to toggle this allows us to support several different use cases, in particular the one where controls components need to launch their own control components (like in the image caption editor) without being contained visually by the cramped bounds of the controls wrapper.

- I renamed `FormattingBarPopover` to `FormattingBarControlsContainer`. This is the component that is responsible for positioning formatting controls, e.g. the citation editor, the hyperlink popup, and the table controls in question. There's a competing notion of a "formatting bar popover" referring to the thing that appears when you mouse over certain disabled formatting bar buttons, like the References ("@") button. I figured this new name removes that confusion and is a little bit more descriptive. Other references to "popovers" in this conceptual space have been renamed to refer to "controls" as well.

_Test plan:_
- Check that Pub controls (e.g. image and citation editor, link popover) work.
- Check that you can still make changes e.g. to image caption content. Check that the controls components within the image caption editor are still mounted and positioned correctly (appearing next to their target nodes and not contained within the tiny caption editor)
- Check that the issue in #1354 is resolved, since the table controls component floats freely instead of sticking with the Pub header.
- Check that other formatting bars, in the Page Layout editor (which had the same sticky problem) and the Discussion/ThreadComment editors are working as expected. In particular, the link popovers should position correctly and selecting an image node should _not_ launch a controls component.


howebenhowe (approved):
LGTM.

PR #1409, opened 2021-05-12
fix: Match NoteManager state to initialStructuredCitations

### LINKED ISSUE #1393: Too many calls to /api/citation-format during Pub load
These are loaded in bulk on the server and we don't need to ask the API to fetch them when the client mounts.

idreyn:
Resolves #1393 

`NoteManager` currently holds notes in this shape:

```ts
Record<StructuredValue, {structuredValue: StructuredValue, renderedStructuredValue: RenderedStructuredValue}>
```

but the `initialStructuredCitations` it receives from the server has the shape:

```ts
Record<StructuredValue, RenderedStructuredValue>
```
This causes us to call `/api/citation-format` for each of the citations in a Pub when it's loaded. My chosen fix is to change `NoteManager` to use the type provided by `initialStructuredCitation`, since I don't think we're using `note.structuredValue` anywhere in `NoteManager`.

_Test plan:_
Add some citations to a Pub and verify that they render correctly. Then refresh the page and verify that the same citations appear in the document with their rendered values intact, but that `/api/citation-format` has not been called.

### DIFF
--- a/server/utils/citations/structuredCitations.ts
+++ b/server/utils/citations/structuredCitations.ts
@@ -2,10 +2,12 @@ import fs from 'fs';
 import path from 'path';
 import crypto from 'crypto';
 import Cite from 'citation-js';
+import { Node } from 'prosemirror-model';
 
 import { getNotes } from 'components/Editor';
 import { citationStyles, CitationStyleKind, CitationInlineStyleKind } from 'utils/citations';
 import { StructuredValue, RenderedStructuredValue } from 'utils/notesCore';
+import { Pub } from 'types';
 
 /* Different styles available here: */
 /* https://github.com/citation-style-language/styles */
@@ -49,7 +51,7 @@ const extractAuthorFromApa = (apaStyleCitation: string) => {
 const getInlineCitation = (
 	citationJson: any[],
 	citationApa: string,
-	inlineStyle: string,
+	inlineStyle: CitationInlineStyleKind,
 	fallbackValue: string,
 ) => {
 	if (inlineStyle === 'authorYear') {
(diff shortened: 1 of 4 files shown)


server/utils/citations/structuredCitations.ts
```diff
@@ -2,10 +2,12 @@ import fs from 'fs';
 import path from 'path';
```
idreyn, line 2: These changes and the ones in `types/pub.ts` aren't part of the fix, I just added them while debugging since we have better types available for these values now!

utils/notesCore.ts
```diff
@@ -39,7 +34,7 @@ export class NoteManagerCore {
 	) {
 		this.citationStyleKind = citationStyleKind;
 		this.citationInlineStyleKind = citationInlineStyleKind;
-		this.notes = notes;
+		this.notes = { ...notes };
```
howebenhowe, line 37: I don't understand what spreading into brackets does. Creates new object instead of just copying a reference to the same object?

howebenhowe (changes requested):
Citations on the qubqub for this PR don't render properly.
See [this demo citation/footnote pub](https://qubqub-pr-1409.herokuapp.com/pub/kzspnp07/draft) and/or screenshot below.

Screenshot from 2021-05-17 13-51-05

idreyn:
@howebenhowe  great catch, just made a change in `usePubNotes` that should fix this

utils/notesCore.ts
idreyn: Exactly — I noticed that the note manager was actually modifying the `initialData` object, which made debugging a little difficult. Not entirely necessary but good hygiene.

howebenhowe (changes requested):
Alphabetization is now busted again, I think because non-rich-text (i.e. unstructured value) got dropped from the object passed for sorting comparison. See inline comment.

Screenshot from 2021-05-18 09-53-11

howebenhowe (approved):

PR #1414, opened 2021-05-18
fix: add .editor to print styles to prevent display block override

### LINKED ISSUE #1172: figure > img elements in exports should have display: flex
Our CSS for HTML/PDF exports gives `figure` elements a max-height of 800px. We want images (and videos, etc, I guess) inside of those figures to have `display: flex` so they don't vertically overflow that, but they currently have `display: block` because of a more specific CSS selector. This can cause Paged to freak out when very tall images are added to a Pub.

### LINKED ISSUE #1400: PDF download broken again.... will not download entire document <https://doi.org/10.21428/8bbb7f85.35a0e14b>

1.   NOT the formatted PDF... the auto-generated one.
2.  When I download, only 147 pages download, not the whole document.
...

The whole document PDF to download.

https://doi.org/10.21428/8bbb7f85.35a0e14b

firefox also Chrome

===

gabestein:
Resolves #1172 
Resolves #1400

Adds an `.editor` directive to the `figure`s in printDocument.scss styles so that all figures display as flex rather than block. This makes it so that figures (and their captions) never display larger than 800px height, which prevents pagedjs errors caused by unsplittable multi-paged images.

_Test plan_
1. Create a doc with a really tall image and some surrounding text (e.g. http://demo.duqduq.org/pub/zk4hj3js/draft) and make sure the PDF generates with the correct image.
2. For good measure, download a really long Pub with lots of multi-sized images (e.g. https://openscientist.duqduq.org/pub/play/release/2) and make sure that the PDF downloads with all its pages and the images all display well in their different sizes.

Gotchas:
- Forgetting to delete a cached `printDocument.css` from `dist/workers/tasks/export/styles/`
- Forgetting to delete cached PDF exports from the dev db

### DIFF
--- a/workers/tasks/export/styles/printDocument.scss
+++ b/workers/tasks/export/styles/printDocument.scss
@@ -8,12 +8,12 @@ https://gitlab.pagedmedia.org/tools/pagedjs-cli/issues/11 */
 @import 'components/Editor/styles/latex.scss';
 
 .editor {
-    @import 'components/Editor/styles/main.scss';
-    @import 'components/Editor/styles/collab.scss';
-    @import 'components/Editor/styles/figure.scss';
-    @import 'components/Editor/styles/file.scss';
-    @import 'components/Editor/styles/table.scss';
-    @import 'components/Editor/styles/textAlign.scss';
+	@import 'components/Editor/styles/main.scss';
+	@import 'components/Editor/styles/collab.scss';
+	@import 'components/Editor/styles/figure.scss';
+	@import 'components/Editor/styles/file.scss';
+	@import 'components/Editor/styles/table.scss';
+	@import 'components/Editor/styles/textAlign.scss';
 }
 
 /* Import some global variables and styles, including fonts */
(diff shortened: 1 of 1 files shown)


workers/tasks/export/styles/printDocument.scss
```diff
@@ -23,206 +23,206 @@ https://gitlab.pagedmedia.org/tools/pagedjs-cli/issues/11 */
...
+	}
+
+	tr,
+	th {
+		break-inside: avoid;
+		max-height: 90vh;
+		overflow-y: hidden;
+	}
+
+	.editor figure {
```
gabestein, line 143: this is the only actual line changed. not sure why it did this to the entire file

gabestein:
sorry about the formatting, and thanks for resetting. Lmk if you need anything else from me to move forward!

idreyn (approved):
Looks good to me, and thank you for handling this whole thing. I think ideally we'd nest all of the styles under `.editor` in a more consistent way whether it's needed for precedence or not, but that would require careful attention so I'm happy to make the smallest change that fixes the problem here!

gabestein:
Thanks! And agreed. I looked at that, I _think_ we could probably get away with doing it for everything under print media, tbh. And it's not that long of a stylesheet, so, wouldn't be more than a few days to go line-by-line. Anyhow, we should do it at some point. That point is not now!

PR #1421, opened 2021-05-21
Activity Item create helper functions

### LINKED ISSUE #1435: TypeScript data model for ActivityItem
- Figure out what kinds of `ActivityItem` there will be
- Write them out in TypeScript

### LINKED ISSUE #1436: Model and creation helper functions for ActivityItem
We are going to be creating lots of `ActivityItems` from all over the backend. Having a set of `create[X]ActivityItem` helper functions that do the heavy lifting in constructing them will minimize the footprint of this feature on critical path code.

howebenhowe:
Resolves #1435
Resolves #1436

Implements suite of functions to facilitate creating ActivityItems for most notification-warranting actions.
Includes activity items for:
Collections
Members
Pubs
Communities

Notably omitted activity items (to implement later):
PubReviews
PubEdge

Testing to follow with implementation of PubReview and PubEdge


server/activityItem/queries.ts
```diff
@@ -0,0 +1,176 @@
...
+
+export const createPubActivityItem = async (
+	kind: 'pub-created' | 'pub-updated' | 'pub-removed' | 'pub-released',
+	userId: string,
+	oldPub: types.Pub,
+	communityId: string,
+	releaseId: string,
+	pubId: string,
+) => {
+	const pub: types.Pub = Pub.findOne({ where: pubId });
```
idreyn, line 55: Need an `await` here (and in a few other equivalent spots)

server/activityItem/queries.ts
```diff
@@ -0,0 +1,176 @@
...
+	const diffs = getDiffsForPayload(pub, oldPub, ['title', 'doi']);
+	createActivityItem({
```
idreyn, line 57: Likewise we should `await` (or actually just `return`) here

server/activityItem/queries.ts
idreyn: Also, food for thought...it may be desirable to avoid making this call in the `pub-created` and `pub-removed` cases, since nothing will actually be diffed here.

server/activityItem/queries.ts
```diff
@@ -0,0 +1,176 @@
...
+): Flags => {
+	return keys.reduce(
+		(memo: Flags, key: EntryKey) =>
+			oldEntry[key] === newEntry[key] ? memo : { ...memo, [key]: true },
+		{} as Flags,
+	);
+};
+
+export const createPubActivityItem = async (
+	kind: 'pub-created' | 'pub-updated' | 'pub-removed' | 'pub-released',
```
idreyn, line 48: `pub-released` probably justifies a separate function here. It's the only one of these where the `releaseId` argument is required, which is possible(?) but grisly to express in TS. I'd suggest removing it from this list along with the `releaseId` parameter and adding this functionality to the second batch.

server/activityItem/queries.ts
```diff
@@ -0,0 +1,176 @@
...
+	userId: string,
```
idreyn, line 49: I suggest calling this `actorId` since it's passed right along to that field of the `ActivityItem`.

types/activity/pub.ts
```diff
@@ -0,0 +1,105 @@
...
+	kind: 'pub-released';
+	payload: {
+		releaseId: string;
+	};
+};
+
+type PubEdgeActivityItemBase = PubActivityItemBase & {
+	payload: {
+		pubEdgeId: string;
+		target: {
```
idreyn, line 45: Not important to fix this now but after reviewing the play this should read:

```ts
target:
| {
        pub: {
            id: string;
            title: string;
            slug: string;
        };
  }
| {
        externalPublication: {
            id: string;
            title: string;
            url: string;
        };
  };
```
since we're guaranteed to have one or the other but not both.

idreyn (changes requested):
Looking pretty good so far — requesting changes mostly for the missing `await`/`return` statements, but there are some other thoughts worth thinking in here about the design of these functions as you continue to build them out.

server/activityItem/model.ts
```diff
@@ -0,0 +1,11 @@
+export default (sequelize, dataTypes) => {
+	return sequelize.define('ActivityItem', {
+		id: sequelize.idType,
+		kind: { type: dataTypes.TEXT, allowNull: false },
+		pubId: { type: dataTypes.UUID },
+		payload: { type: dataTypes.JSONB },
+		communityId: { type: dataTypes.UUID, allowNull: false },
+		actorId: { type: dataTypes.UUID, allowNull: false },
+		collectionId: { type: dataTypes.UUID },
```
idreyn, line 9: Oh, yeah, could we add a field here:

```js
timestamp: {
    type: dataTypes.DATETIME,
    defaultValue: sequelize.Sequelize.NOW,
    allowNull: false,
}
```
been thinking that we probably want to have an explicit statement of the moment in time these items refer to besides the `createdAt` and `updatedAt` that Sequelize creates, since some will be created after the fact with a backfill, and might but updated later in the future.

server/activityItem/queries.ts
```diff
@@ -0,0 +1,323 @@
...
+export const createPubDiscussionActivityItem = async (
+	kind: 'pub-discussion-comment-added',
+	actorId: string,
+	communityId: string,
+	isReply: boolean,
+	pubId: string,
+	discussionId: string,
+	threadCommentId: string,
+) => {
+	const pub: types.Pub = await Pub.findOne({ where: { id: pubId } });
```
idreyn, line 297: These could be parallelized with `Promise.all()`

idreyn (approved):
Looking really great!

Once we start putting these into place I think we will find some spots where passing in an entire object we already have in scope may be more appropriate than passing in its ID and querying for it again...this is particularly true of the functions we'll call from Sequelize hooks. But, no way to know for sure which of these that applies to, so best to cross that bridge if we come to it.

server/activityItem/queries.ts
```diff
@@ -0,0 +1,340 @@
...
+};
+
+export const createPubActivityItem = async (
+	kind: 'pub-created' | 'pub-updated' | 'pub-removed',
+	actorId: string,
+	oldPub: types.Pub,
+	communityId: string,
+	pubId: string,
+) => {
+	const pub: types.Pub = await Pub.findOne({ where: pubId });
```
idreyn, line 220: This should be `{where: {id: pubId}}`

server/activityItem/queries.ts
```diff
@@ -0,0 +1,340 @@
...
+};
+
+export const createPubReleaseActivityItem = async (
+	kind: 'pub-released',
+	actorId: string,
+	communityId: string,
+	releaseId: string,
+	pubId: string,
+) => {
+	const pub: types.Pub = await Pub.findOne({ where: pubId });
```
idreyn, line 243: Same here

server/activityItem/queries.ts
```diff
@@ -0,0 +1,340 @@
...
+	oldMember: types.Member,
+) => {
+	const member: types.Member = await Member.findOne({ where: { id: actorId } });
+	const diffs = getDiffsForPayload(member, oldMember, ['permissions']);
+	return createActivityItem({
+		kind,
+		actorId,
+		communityId,
+		payload: {
+			userId: memberId,
```
idreyn, line 54: We want the ID of the `User` associated with the `Member` object — so `member.userId`

server/activityItem/queries.ts
```diff
@@ -0,0 +1,340 @@
...
+		communityId,
+		payload: {
+			...diffs,
+			pub: {
+				title: pub.title,
+			},
+			threadId: review.threadId,
+			isReply,
+			threadComment: {
+				id: review.threadId,
```
idreyn, line 170: This should be `threadCommentId`

server/activityItem/queries.ts
```diff
@@ -0,0 +1,340 @@
...
+) => {
+	const threadComment: types.ThreadComment = await ThreadComment.findOne({
+		where: { id: threadCommentId },
+	});
+	const review: types.Review = await ReviewNew.findOne({ where: { id: reviewId } });
+	const pub: types.Pub = await Pub.findOne({ where: { id: review.pubId } });
+	const diffs = getDiffsForPayload(review, oldReview, ['status']);
+	return createActivityItem({
+		kind,
+		pubId: threadComment.id,
```
idreyn, line 159: this should be `pubId`

idreyn, line 159: This should be `pubId`

server/activityItem/queries.ts
```diff
@@ -0,0 +1,340 @@
...
+				id: threadComment.id,
+				text: threadComment.text,
+				userId: actorId,
+			},
+			pub: { title: pub.title },
+		},
+	});
+};
+
+export const createPubReviewUpdatedActivityItem = async (
```
idreyn, line 142: We don't need any `Thread`-related payload for this item (it's related strictly to changes in the `Review` model itself) so you can simplify this considerably!

PR #1462, opened 2021-06-07
update: Community ActivityItem backfill tool

### LINKED ISSUE #1441: ActivityItem backfill for existing Communities
It will be helpful for Communities (as well as us in testing) to have some historical `ActivityItems` that we can reconstruct from the Community's history, for instance from the creation dates of `Pub` and `Member` objects. With #1436 this should hopefully not be too much work and a pretty good bang for buck — though I do expect it will take a while to run.

idreyn:
Resolves #1441

In this PR we add a tool that creates as many `ActivityItems` as can be inferred from the history of a Community. Also note:

- I made the `actorId` property of an `ActivityItem` nullable because it's not possible to infer an actor in many cases here
- There are some changes to some of the `ActivityItem` types and creation queries to best reflect their actual use (for instance, I discovered that not every created Review comes with a ThreadComment).

_Test plan:_

I ran
```
npm run tools backfillCommunityActivity -- --community kfg-notes
```
and observed that it completed, creating about 4,000 items in the database. I did the same thing with `hdsr` for good measure.

### DIFF
--- a/server/activityItem/queries.ts
+++ b/server/activityItem/queries.ts
@@ -1,22 +1,23 @@
 import * as types from 'types';
 import {
 	Collection,
-	Member,
+	CollectionPub,
 	Community,
-	PubEdge,
 	Discussion,
 	ExternalPublication,
-	ReviewNew,
+	Member,
 	Pub,
-	CollectionPub,
-	ThreadComment,
+	PubEdge,
+	Release,
+	ReviewNew,
 	Thread,
+	ThreadComment,
 } from 'server/models';
 
 import { getDiffsForPayload, getChangeFlagsForPayload, createActivityItem } from './utils';
 
 const resolvePartialMemberItem = async (member: types.Member) => {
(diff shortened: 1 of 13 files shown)


server/activityItem/queries.ts
```diff
@@ -1,22 +1,23 @@
...
+	ReviewNew,
 	Thread,
+	ThreadComment,
 } from 'server/models';
 
 import { getDiffsForPayload, getChangeFlagsForPayload, createActivityItem } from './utils';
 
 const resolvePartialMemberItem = async (member: types.Member) => {
-	if ('pubId' in member) {
+	if ('pubId' in member && member.pubId) {
```
howebenhowe, line 20: same as
```
if (member.pubId) {
...
}
```
right? Same below for `collectionId`

howebenhowe (commented):
Looks good. Only suggested change was on if conditions, but seems like that's a more explicit style.

PR #1477, opened 2021-06-17
feature: Alignment options for Pages & Collections blocks

### LINKED ISSUE #1459: Ability to center Page/Collection blocks
Enhancement requests are requests to make changes to an existing PubPub feature. If you'd like to suggest a new feature, please post in the ideas section of our [https://github.com/pubpub/pubpub/discussions](discussion forum).

_Before filing your enhancement, please search existing issues to make sure a similar one doesn't already exist. If one does, please leave a comment or reaction in support of the issue rather than creating a new one._

_To fill out the request, please leave the below headers (**) in tact, and fill out each section. If a section isn't applicable, leave it blank or write n/a under it. Please do not delete any template sections._

Would be really really great if we had the option to center these

Would look a lot better on certain pages and with a certain number of pages in the block

---

qweliant:
Resolves #1459 

This feature allows users to change the alignment of pages and collections in the Collections & Pages Block editor. More selections can be made in regard to justification (see [here](https://css-tricks.com/almanac/properties/j/justify-content/)).

Test plan:

Visit the Pages dashboard of a Community and verify that:

Collections & Pages Block has a Justify Content section
justification can be be changed in the editor
justification setting persist after saving changes by going to the home page
a user that is not logged in also sees the changes

Screen Shot 2021-06-21 at 11 43 42 AM

Screen Shot 2021-06-21 at 11 58 46 AM

### DIFF
--- a/client/components/LayoutEditor/LayoutEditorPagesCollections.tsx
+++ b/client/components/LayoutEditor/LayoutEditorPagesCollections.tsx
@@ -1,12 +1,13 @@
 import React, { useCallback, useMemo } from 'react';
 import { Button, FormGroup } from '@blueprintjs/core';
 
-import { InputField, Popover, OrderPicker, PubMenuItem } from 'components';
+import { InputField, Popover, OrderPicker, PubMenuItem, MenuSelect, MenuSelectItems } from 'components';
 import LayoutPagesCollections, {
 	Content,
 	BlockItem,
 	PageOrCollection,
 } from 'components/Layout/LayoutPagesCollections';
+import { CollectionsPagesJustifyType } from 'utils/layout/types';
 
 type Props = {
 	onChange: (index: number, block: Content) => any;
@@ -45,6 +46,13 @@ const getAllItems = (
 	].sort((a, b) => (a.title > b.title ? 1 : -1));
 };
 
+const justifiedContent: MenuSelectItems<CollectionsPagesJustifyType> = [
+	{ value: 'center', label: 'Center' },
+	{ value: 'space-between', label: 'Space Between' },
+	{ value: 'space-around', label: 'Space Around' },
+	{ value: 'left', label: 'Left' },
+];
+
 const LayoutEditorPagesCollections = (props: Props) => {
 	const { layoutIndex, onChange, content, collections, pages } = props;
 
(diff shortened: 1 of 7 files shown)


qweliant:
Screen Shot 2021-06-21 at 12 00 48 PM


The width of the select menu should match the label.(?) Any thoughts @idreyn @howebenhowe to make this happen or if it should.

client/components/LayoutEditor/LayoutEditorPagesCollections.tsx
```diff
@@ -84,28 +108,44 @@ const LayoutEditorPagesCollections = (props: Props) => {
...
-							<OrderPicker
-								availableItems={availableItems}
-								selectedItems={selectedItems}
-								onSelectedItems={setSelectedItems}
-								renderItem={(item, handleClick) => (
-									<PubMenuItem title={item.title} onClick={handleClick} />
-								)}
+
+				<div className="dropdown-grouping">
+					<div className="dropdown">
```
idreyn, line 113: I think I found a fix for the way this is jumping around as the `MenuSelect` width changes — if you add:
```tsx
buttonProps={{fill: true, alignText: 'left'}}
```
to the `MenuSelect`, it'll cause the `Button` that it renders to (1) fill the width of the container and (2) push its text to the left and the caret to the right. (Eventually we may find that this should be the default behavior). By the way, I'm reading this off the [Blueprint docs](https://blueprintjs.com/docs/#core/components/button) which are extremely handy.

With this done all that's left is to set the width or min-width of this div to something like 150px. I'd suggest giving it a new class name and referencing it in `layoutEditor.scss` (nested under `.layout-editor-pages-component`)

idreyn (approved):
Looks good to me but see the one comment inline!

client/components/LayoutEditor/layoutEditor.scss
```diff
@@ -46,6 +47,13 @@
 			padding-top: 12px;
 		}
 	}
+	.dropdown-grouping {
+		display: flex;
+		// align-items: flex-end;
+		.dropdown {
+			margin: 0 0 5px 20px;
```
idreyn, line 54: I don't think any further `margin-bottom` is required here (but interested to hear your thoughts on it)

qweliant: it doesn't, I think it was a copy paste thing or maybe i freaked before adding the 20px to the left. its removed

PR #1479, opened 2021-06-23
dev: Remove keen for good and update admin dashboard

gabestein:
This removes all remaining Keen call-sites and references, including:

- hydrateWrapper
- cookie settings (text and actual cookie name)
- admin dashboard

The admin dashboard stopped functioning when we removed Keen, so I updated it to use the new Metabase dashboard.

_Test plan_
- Make sure there are no Keen calls on any pageload
- Make sure you can toggle cookie preferences, and when doing so Heap cookies still appear/disappear, and the cookie opt-out/in we drop is still set properly
- Visit `/admin` from a community (e.g. demo) and make sure it 404s
- Visit `/admin` on the www community while logged out and make sure it 404s
- Visit `/admin` on the www community while logged into your personal account or the pubpub team account and make sure the dashboard loads
- Visit `/admin` on the www community while logged into a test account that is not your personal or the pubpub team account and make sure it 404s


client/containers/AdminDashboard/AdminDashboard.tsx
```diff
@@ -1,58 +1,42 @@
...
-	}
-}
+type Props = {
+	impactData: any;
+};
+
+const AdminDashboard = (props: Props) => {
+	const { impactData } = props;
+	const { baseToken } = impactData;
+	const genUrl = (token) => {
```
idreyn, line 13: This could probably just be a constant instead of a function since it's called once and its argument is available right upstairs.

idreyn (approved):
Looks awesome!

PR #1480, opened 2021-06-23
update: Use tristate instead of boolean for subscription mute logic 

idreyn:
In #1472 I added a `UserSubscription` with a `muted` field that would allow users to elect not to get notifications for a Pub or a Thread if they no longer wish to do so. My intention is that `muted` should override subscription states at higher scopes — so if a user is subscribed to a Pub but a Thread within that Pub is muted (thinking of a conversation which has become unproductive that an author no longer wishes to participate in) then they should not receive notifications for that Thread.

Unfortunately, in this case, the opposite of "muted" is not "subscribed" — if the user changes their mind and unmutes this Thread, _then_ unsubscribes from the Pub, this would leave them with a `UserSubscription` instance with `mute = false` hanging around keeping them subscribed to the Thread. I think we can manage this by carefully deleting `UserSubscription` instances but I'd much rather just let the model accomodate this. So instead of `muted,` `UserSubscription` now has a `status` property with three values:

- `active`: the subscription is actively generating notifications
- `inactive`: the subscription is not actively generating notifications. from the user's perspective it should behave as though it's not there, but from our perspective it's easier to manage than continually creating/destroying `UserSubscription` instances as they change their mind.
- `muted`: the subscription is muted and users will not receive any notifications from it.

_Test plan:_
```
npm run test-dev server/userSubscription
```


server/visibility/queries.ts
```diff
@@ -23,9 +18,36 @@ export const getParentModelForVisibility = async (
 	return null;
 };
 
+type UpdateOptions = {
+	visibilityId: string;
+	access: types.Visibility['access'];
```
howebenhowe, line 23: hadn't realized (remembered?) you could access type properties (that is, sub-types?) this way. Oh, Typescript, you never cease to humble.

howebenhowe (approved):

PR #1481, opened 2021-06-23
update: Notifications model, queries, and api 

idreyn:
This PR introduces the `UserNotification` model and associated code. Every `UserNotification` is associated with an `ActivityItem` from which it derives its text content. We will currently only generate notifications for activity items of kind `pub-discussion-comment-added` and `pub-review-comment-added`.

_Test plan:_
Run the unit tests at:
```
npm run test-dev server/userNotification
```


server/activityItem/fetch.ts
```diff
@@ -135,7 +135,7 @@ const fetchActivityItemModels = async (
 
 const getActivityItemAssociationIds = (
 	items: types.ActivityItem[],
-	scope: Scope,
+	scope?: Scope,
```
qweliant, line 138: is this setting Scope to a null value if Scope has not been set?

qweliant: i suppose so: https://www.educba.com/typescript-question-mark/

qweliant, line 138: What does an optional scope resolve?

server/activityItem/fetch.ts
```diff
@@ -152,12 +152,14 @@ const getActivityItemAssociationIds = (
...
-	if ('collectionId' in scope) {
-		collection.add(scope.collectionId);
+	if (scope) {
+		community.add(scope.communityId);
+		if ('pubId' in scope) {
+			pub.add(scope.pubId);
+		}
+		if ('collectionId' in scope) {
+			collection.add(scope.collectionId);
+		}
```
qweliant, line 162: does this add activity notifications for users in some "scope"? is the scope a pub and/or collection

server/activityItem/fetch.ts
```diff
@@ -277,12 +279,19 @@ const fetchAssociations = (
...
+};
+
 export const fetchActivityItems = async (
 	options: FetchActivityItemsOptions,
 ): Promise<ActivityItemsFetchResult> => {
 	const { offset = 0, limit = 50, scope, filters = [] } = options;
 	const activityItems = await fetchActivityItemModels({ offset, limit, scope, filters });
-	const associationIds = getActivityItemAssociationIds(activityItems, options.scope);
-	const associations = await fetchAssociations(associationIds);
+	const associations = await fetchAssociationsForActivityItems(activityItems, options.scope);
```
qweliant, line 295: does this mean, instead of fetching all activity, you are fetching activity per association to a pub or collection

server/activityItem/fetch.ts
idreyn: The `Scope` here provides some IDs for a Community/Collection/Pub trio — if we pass them in, those models will be included in the result, otherwise they won't. We pass a `Scope` in for the Activity dashboard (which belongs to a scope) but not for the notifications tray (which doesn't).

server/activityItem/fetch.ts
idreyn: On this line we've already fetched the `ActivityItems` and we're now requesting models associated with those items — for instance a `pub-discussion-created` activity item will want to load the associated `Discussion` model. We're just changing how the `scope` gets passed around because we now want to expose the `fetchAssociationsForActivityItems` function for callers that _already_ have activity items.

howebenhowe (approved):
LGTM, test came back clean.
Gonna need a lesson on augmenting Modelize one day, perhaps.

PR #1482, opened 2021-06-23
update: Derive UserNotifications from ActivityItems 

idreyn:
This PR creates notifications for users when comments are created! We accomplish this by generating appropriate `ActivityItems` when comments are added using a Sequelize hook on `ThreadComment`, and then using a second hook to respond to `ActivityItem` creation and generate `UserNotifications`.

This PR introduces a `defer()` utility that we should use to wrap code executing inside an async function that we would like to defer until after the function returns — in this case, we use twice: first to defer the creation of activity items from comments, and second to defer the creation of notifications from activity items. This function doesn't do much other than bookkeeping — we could accomplish the same thing just by not using `await` in front of a function call — but we use it to keep track of outstanding deferred tasks so we can wait for them to complete in tests. And maybe someday, somewhere else.

I considered using our worker queue to run deferred tasks, but I found that it might be a somewhat heavyweight solution here — the time it takes for an item to make it from the server to a worker can be ~5 seconds, which is a long enough time for these tasks to complete. The more important factor here is idempotency; worker tasks are designed to be retried on failure, where here I wanted to avoid generating notifications multiple times for users if there is a partial failure in the function that does so.

_Test plan:_
There are new tests that test notifications being created:
```
npm run test-dev server/userNotification/__tests__/hooks.test.ts
```


server/userNotification/create.ts
```diff
@@ -0,0 +1,84 @@
...
+		...unmutedThreadSubscriptions,
+		...unmutedPubSubscriptionsNotSupersededByThreadMute,
+	];
+
+	const subscriptionsByUserId = indexByProperty(
+		subscriptionsThatMayProduceNotifications,
+		'userId',
+	);
+
+	const userIdsToNotifty = await filterUsersWhoCanSeeThread({
```
howebenhowe, line 56: surely, typo for notify

server/threadComment/hooks.ts
```diff
@@ -1,12 +1,41 @@
...
+};
+
 ThreadComment.afterCreate(async (threadComment: types.ThreadComment) => {
 	await setUserSubscriptionStatus({
 		userId: threadComment.userId,
 		threadId: threadComment.threadId,
 		setAutomatically: true,
 		status: 'unchanged',
 	});
+	defer(() => createActivityItems(threadComment));
```
howebenhowe, line 40: should this be singular instead of plural? It either creates one or the other activityItem, but still only one at a time, no?

howebenhowe (approved):
One typo, and one plural -> singular naming change, otherwise groovy.

server/userNotification/create.ts
idreyn: Fixed not here but in the `notifications` branch.

server/threadComment/hooks.ts
idreyn: Also fixed in `notifications`

Issue #1451, opened 2021-05-28
User notification UI


idreyn:
At long long last with #1450 done we can build a little widget to show users their unread notifications (relating exclusively to Threads for the time being). I imagine this having a little chat bubble icon and living in the global nav somewhere.

PR #1483, opened 2021-06-23
update: UserNotificationPreferences model, queries, and api 

idreyn:
This PR includes a new model, `UserNotificationPreferences`, which...lets users set their notification preferences. In lieu of more actionable testing information, here is the information this model stores:

-  `receiveNotifications`: whether to receive notifications at all (defaults true)
- `subscribeToThreadsAsCommenter`: whether to automatically subscribe to a thread when commenting in it (defaults true)
- `subscribeToPubsAsMember`: whether to subscribe to Pubs when you are added to them directly as Members (defaults false for now)
- `subscribeToPubsAsContributor`: whether to subscribe to Pubs when you are added to them as a contributor (defaults false)
- `notificationCadence`: a number of minutes across which we are guaranteed not to show you new notifications twice. For instance, if this is set to `60` and we show you new notifications at 8:07 AM, we will not do so again until 9:07
- `markReadTrigger`: a strategy for marking notifications read, taking on possible values:
    - `seen`: mark the notification read when it comes into view on the frontend
    - `clicked-through`: mark the notification read when the user clicks a link within it
    - `manual`: don't mark notifications read; let users do this themselves

_Test plan:_
Sadly I have not written any API tests here yet and you'll have to head over to #1451 to verify that this works.


howebenhowe (approved):
LGTM, nice and straight forward

PR #1484, opened 2021-06-23
dev: Move PubHeaderBackground to components directory

idreyn:
This is a quick one to move the `PubHeaderBackground`, which renders the unique accent color and background for a Pub, out of the `Pub` container and into the `components` directory. We use this in several places already and we will add one more in #1451.

_Test plan:_
It should suffice to visit the Overview dashboard tab for a Pub and verify that it loads the `PubHeaderBackground` behind the Pub title.


client/utils/activity/titles/user.ts
```diff
@@ -27,13 +26,7 @@ const getUserTitleFromUserIdAndContext = (
 };
 
 export const actorTitle: TitleRenderer<InsertableActivityItem> = (item, context) => {
-	const { otherActorsCount } = context;
```
idreyn, line 30: Please ignore these last few files here — these are a result of a mistake when splitting up a branch into many PRs.

client/components/PubHeaderBackground/pubHeaderBackground.scss
```diff
@@ -1,4 +1,4 @@
-@import './themes.scss';
+@import '../../containers/Pub/PubHeader/themes.scss';
```
qweliant, line 1: why do the pubheader themes stay in containers if imported here?

qweliant (commented):
Just file locations and import changes. Only curious about why theme.scss imports from container/Pub/PubHeader are made

qweliant (approved):

PR #1485, opened 2021-06-23
dev: Split underlying Avatars component out of ContributorAvatars

idreyn:
We have a `ContributorAvatars` component that shows the avatars of users who contributed to a Pub. We want a generic `Avatars` component that can be used to show any list of users. This PR does that, and changes `ContributorAvatars` so that it is based on the more generic component.

_Test plan:_
Visit a Page with a Pubs block (the `demo` homepage should work fine) and verify that in Pub previews, user avatars appear correctly and have a popover on hover.


idreyn:
Returning this to draft mode because I just found a new layout bug

client/components/Avatar/Avatar.tsx
```diff
@@ -12,7 +12,7 @@ const propTypes = {
 	avatar: PropTypes.string,
 	instanceNumber: PropTypes.number,
 	borderColor: PropTypes.string,
-	borderWidth: PropTypes.string,
+	borderWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
howebenhowe, line 15: Why should `borderWidth` be either string or number? Seems like one or the other could do the trick, so why accept both?

qweliant: i too would like to know

qweliant (approved):

client/components/Avatar/Avatar.tsx
idreyn: It would make the most sense for it to be a number but there are a number of existing call sites that pass in a string (which gets implicitly cast to a number) and I don't want to change and test those right now.

PR #1486, opened 2021-06-23
feature: Pub/Thread subscription button

idreyn:
This PR provides a new `SubscriptionButton` component which renders a dropdown list that is used to choose a subscription status for a Pub or Thread.

_Screenshot:_
image

_Test plan:_
When logged in, visit a Pub and scroll down to the discussions section at the bottom. Use the new bell icon to update the Pub's subscription status and check that it persists across refreshes. When logged out, the button should not appear.


howebenhowe (approved):
LGTM

PR #1495, opened 2021-06-29
fix: Enforce that Pages have a non-null layout

### LINKED ISSUE #1494: Pages created with layout = null don't load
Currently pages can have a null `layout` property — when they are loaded the UI uses `getDefaultLayout()` as a a fallback. Since we stopped loading every single Pub onto the client to render pages, this confuses the client because the server doesn't understand that this is the layout the client expects to render.

We'll solve this problem by enforcing that Page layouts are non-nullable and backfilling existing null layouts with the default one.

idreyn:
Resolves #1494 

We currently allow Pages to be created with `layout = null`. When we load such a Page, the client uses the value of `getDefaultLayout()` as a fallback layout, but the server doesn't send down Pubs for that layout, which results in an error. We solve this problem by enforcing that Pages do have a non-null `layout`, by backfilling existing null layouts to the default value, and using `getDefaultLayout()` — renamed to `generateDefaultPageLayout()` — to give Pages a layout when created.

This also fixes two minor bugs:
- The existing default layout blocks have an invalid `content.size` which has been replaced with `content.pubPreviewType`
- There is currently the possibility of adding a `null` value to `Community.navigation` during Page creation.

I ran the following to backfill null layouts (it's harmless but unnecessary to repeat these)
```
npm run tools migrate -- --name data/2021_06_29_backfillPageLayouts
```
and observed that a few hundred rows were affected. Then I ran:
```
npm run tools migrate -- --name 2021_06_29_makePageLayoutNonNullable
```
which successfully updated the database schema to prevent null Page layouts. These commands will need to be repeated at deploy time with `PUBPUB_PRODUCTION=true`.

_Test plan:_

I created a new Page and observed that it took on the default generated layout and was able to load at both `/dash/pages/:slug` and `/:slug`.

### DIFF
--- a/stubstub/modelize/builders.js
+++ b/stubstub/modelize/builders.js
@@ -7,6 +7,7 @@ import { Community, Member, Release, User } from 'server/models';
 import { createPub } from 'server/pub/queries';
 import { createCollection } from 'server/collection/queries';
 import { createDoc } from 'server/doc/queries';
+import { createPage } from 'server/page/queries';
 
 const builders = {};
 
@@ -64,6 +65,8 @@ builders.Pub = async (args) => {
 
 builders.Collection = ({ title = 'Collection ' + uuid.v4(), kind = 'issue', ...restArgs }) =>
 	createCollection({ title, kind, ...restArgs });
+
+builders.Page = createPage;
 
 builders.Member = async ({ pubId, collectionId, communityId, ...restArgs }) => {
 	const getTargetArgs = () => {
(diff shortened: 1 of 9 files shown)


stubstub/modelize/builders.js
```diff
@@ -65,6 +66,8 @@ builders.Pub = async (args) => {
 builders.Collection = ({ title = 'Collection ' + uuid.v4(), kind = 'issue', ...restArgs }) =>
 	createCollection({ title, kind, ...restArgs });
 
+builders.Page = createPage;
```
idreyn, line 69: This overrides what happens when we start a `Page {}` entry in Modelize.

howebenhowe (approved):

Issue #1434, opened 2021-05-28
Include publisher field and custom publication name in metadata (at community level, probably?)


gabestein:
### What's your enhancement request?
Be able to specify the publisher and custom publication name at the community level.

### What would this enhancement help you accomplish?
- Generate publisher metadata tags for citations, Google Scholar, crossref, and others.
- Specify a different name for citations

### Features

1. New community settings text field under "details": "Cite as." If filled out, this field replaces the community name as the text in journal and conference (but not book chapter) collections:
-  The name of the `container-title` in all citationjs callsites (cite as dropdown, pub settings, pub dashboard overview)
- The name of the `citation_journal_title` in HTML meta tags on Pub and collection pages
- The name of the publication in exports (cover page and headers)
- The name of the `<journal_metadata><full_title>` in crossref deposits

2. New community settings text field under "details": "Publisher." If filled out, this field is used in all types of collections for:
- The name of the `publisher` in all citationjs callsites (cite as dropdown, pub settings, pub dashboard overview)
- A new `citation_publisher` HTML meta tag on Pub and collection pages
- A new publisher name in export cover pages
- The name of the `<publisher><publisher_name>` in crossref deposits

howebenhowe:
[publisher crossref reference](https://data.crossref.org/reports/help/schema_doc/4.4.2/schema_4_4_2.html#publisher)

PR #1502, opened 2021-07-09
feature : Add "Cite as" and "Publish as" fields to community model

qweliant:
addresses issue #1434 

This PR introduces the ability for community moderators to add citation and publishing information in the community settings page.

This will also be reflected in the exportation of journals and in citation information

Test plan:

View cite as and publisher info in the dashboard for community settings

Make sure that the same info is present in pub overview and pub details

go to pub details and view the crossref deposit for each

books should have the publisher listed, while pubs should have the journal name listed

export a pub as an PDF that is part of an issue. It should list the journal at the top. 

do the same for a pub in a book and a conference. it should only have the publisher


gabestein:
Open Qs:
- Can we deposit to crossref without a publisher? I'll test on crossref and post to forum for guidance if no publisher.

gabestein:
For depositing to Crossref sans publisher: what I found is it doesn't seem like we're depositing a publisher at all right now, despite the schema in `utils/crossref/schema/publisher.js`

Good news: we can continue to default to not including a publisher.

Bad news: I think we're going to need a bit more work to wire up the publisher field.

qweliant:
Not sure if booktitle is the publisher or actual book and title is the chapter

image

gabestein:
IIRC, booktitle = collection title, title = chapter title, and publisher should be a separate field passed into citation.js.

gabestein (changes requested):
Sweet! This turned out to be a pretty darn complex one, touching a ton of systems, and I think we're mostly there save for a few small UX improvements now that I've had a chance to see things live, and some places downstream where the new fields aren't quite getting populated correctly. That's suggesting to me that our metadata structure is probably getting unwieldy and will likely need an overhaul at some point. But for now, we must carry on.

## UX
- Aesthetically, I think both of the new fields should be `input`s rather than `textarea`s, as they're usually short and limited to one line. 
- "Publish your book as" might be a tad confusing for the field label, since (I think, anyway) it applies to both books and conferences, and it could also be misinterpreted as a being the title of the book. So, I think it should probably be "Publisher", which, though old-fashioned and staid, is pretty widely  recognized. 
- As we discussed on our Tuesday call, let's add a tooltip to each to help clarify the fields. 

For "Cite As", what about something like:
"When filled out, this field will be used as the journal title for issue collections in citations, PDF exports, and Crossref deposits. By default, the Community Title is used."

For "Publisher", I propose:
"When filled out, this field will be used as the publisher for book and conference proceedings collections in citations, PDF exports, and Crossref deposits."

## Publisher
- The `citation_publisher` meta field appears to be duplicated, with both "PubPub" and the field value showing up in meta. The field value should effectively overwrite "PubPub" if filled out, rather than duplicating it.
- Publisher doesn't seem to appear in citations when filled out. I think you may need to add a new `publisher` entry to `generateCitationHtml.ts.` Once we figure it out, it will will show up as a `publisher={Text}` entry in the generated bibtex.
Screen Shot 2021-07-15 at 16 51 19
- Similarly, Publisher isn't showing up in eligible Crossref deposits, either in the preview, the generated XML, or final deposits. Staring at the code, I'm not exactly sure why. I'm happy to pair with you next week to take a look, or you may find it more efficient look it over with @idreyn or @3mcd when he arrives, since he built the system in the first place! Also, we should introduce you to the arcana of QAing crossref deposits, which is, let me tell you, so much fun.
- Looking at PDF exports, I'm thinking we may just want to put the publisher name on the cover page, and not subsequent pages. And rather than doing so in that title and community string, I think we should probably do it at the bottom of the cover page. Let's chat about some options.

## Cite As
- Metadata and citations look good!
- It looks like this is currently appearing on PDFs of all types. It should only appear on PDFs where the Pub's primary collection is an Issue. For books and conferences, only the Collection title should appear.
- Like Publisher, it looks like Cite As isn't making it to the Crossref deposits for some reason. As before, I'd be happy to pair next week to help noodle it out, although it might be more efficient for you to consult with Ian or Eric.

client/containers/DashboardSettings/dashboardSettings.scss
```diff
@@ -24,4 +24,7 @@
 	.bp3-card {
 		margin-bottom: 15px;
 	}
+	.settings-section-component .bp3-label .bp3-popover-wrapper {
```
qweliant, line 27: how specific do css selector tags need to be?

server/community/model.ts
```diff
@@ -18,6 +18,8 @@ export default (sequelize, dataTypes) => {
 				unique: true,
 			},
 			title: { type: dataTypes.TEXT, allowNull: false },
+			citeAs: { type: dataTypes.TEXT },
```
qweliant, line 21: curious how db migrations to prod get handled?

server/utils/citations/generateCitationHtml.ts
```diff
@@ -66,8 +66,9 @@ export const generateCitationHtml = async (pubData, communityData) => {
 		type: 'article-journal',
 		title: pubData.title,
 		...authorsEntry,
-		'container-title': communityData.title,
+		'container-title': communityData.citeAs || communityData.title,
```
qweliant, line 69: I think I should check what kind of collection this is, but i'm not sure. the same for publisher. that way, this data only gets set for collections that are a book, issue, or conference instead of everything.

server/utils/ssr.tsx
```diff
@@ -221,8 +235,7 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 			<meta key="pa1" property="article:published_time" content={String(publishedAt)} />,
 			<meta key="pa2" property="dc.date" content={dcPublishedAt} />,
 			<meta key="pa3" name="citation_publication_date" content={googleScholarPublishedAt} />,
-			<meta key="pub1" name="citation_publisher" content="PubPub" />,
-			<meta key="pub2" property="dc.publisher" content="PubPub" />,
+			<meta key="pub1" property="dc.publisher" content="PubPub" />,
```
qweliant, line 238: should i add meta data for conference as well?

client/containers/DashboardSettings/CommunitySettings.tsx
```diff
@@ -222,6 +228,65 @@ const CommunitySettings = () => {
...
+										title for
+										<br />
+										issue collections in citations, PDF exports, and Crossref
+										deposits.
+										<br />
+										By default, the Community Title is used.
+									</span>
+								}
+								// @ts-expect-error ts-migrate(2322) FIXME: Type '{ children: Element; content: Element; toolt... Remove this comment to see the full error message
+								tooltipClassName="bp3-dark"
```
3mcd, line 248: I think we might be able to remove these `tooltipClassName` props (and the corresponding `// @ts-expect-error` comments) while we're in this file. I removed them in a local branch and the tooltips seem to render exactly the same.

client/containers/DashboardSettings/CommunitySettings.tsx
```diff
@@ -222,6 +228,65 @@ const CommunitySettings = () => {
...
+								<Icon icon="info-sign" />
+							</Tooltip>
+						</span>
+					}
+					type="text"
+					placeholder="This will show up in citations and BibTex as your journal name"
+					value={citeAs}
+					onChange={(evt) => {
+						setCiteAs(evt.target.value);
+					}}
```
3mcd, line 259: The `<InputField>` component currently doesn't account for non-string labels. It renders our label w/ tooltip fine (especially with the CSS changes we made), but the `for` and `id` attributes render like `id="input-[object Object]"` because JS automatically casts the JSX we pass to the `label` prop to the string "[object Object]" when  concatenated with the `"label-"` prefix [here](https://github.com/pubpub/pubpub/blob/master/client/components/InputField/InputField.tsx#L56) and [here](https://github.com/pubpub/pubpub/blob/master/client/components/InputField/InputField.tsx#L83). This introduces an accessibility problem where clicking the label of the "Publisher" input actually focuses the "Journal Citation" input, since the `for` attribute links the label to an `<input>` element.

We'll need to modify the `<InputField>` component to fix this. Taking a look at the `<ImageUpload>` component, it looks like an `htmlFor` attribute was exposed which is used in both the final `for` and `id` props passed to the underlying HTML `<input>` element.

Sorry, I know this is an information-dense comment. Happy to hop on a quick call to discuss if I did a bad job with my explanation.

server/utils/ssr.tsx
gabestein: yes!

Also, sorry for adding a new thing, but I just realized the `dc.publisher` should also change to be whatever's in the Publisher field. That's properly where Zotero, etc. have been grabbing the publisher from.

server/utils/citations/generateCitationHtml.ts
gabestein: Yes, this only applies to Journal Issue collections. In Books and Conferences, there is no `container-title`, it's a book or conference title, and we use the Collection's title instead of the Community title. This is because journals have an extra layer to their metadata (Journal Title (Community or Cite As) -> Issue Title (Collection Title) -> Article Title (Pub Title)), whereas books and conference only have two (Book/Conference Title (Collection Title) -> Chapter/Paper Title (Pub Title))

workers/tasks/export/html.tsx
```diff
@@ -194,15 +198,19 @@ const renderFrontMatterForHtml = ({
...
+	publisher,
 }) => {
 	const affiliations = [
 		...new Set(attributions.map((attr) => attr.affiliation).filter((x) => x)),
 	];
 	const communityAndCollectionString =
-		communityTitle + (primaryCollectionTitle ? bullet + primaryCollectionTitle : '');
+		(citeAs || communityTitle) +
+		(primaryCollectionTitle ? bullet + primaryCollectionTitle : '') +
+		(publisher ? bullet + publisher : '');
```
3mcd, line 213: Could we hoist the `citeAs` prop out of the `renderFrontMatter...` functions and compute the title (using the same `citeAs || pubData.community.title` logic) in `getPubMetadata`? That way we only have to do it in one spot, instead of both frontmatter render functions.

workers/tasks/export/metadata.ts
```diff
@@ -68,6 +70,10 @@ export const getPubMetadata = async (pubId) => {
 		citationStyle: pubData.citationStyle,
 		citationInlineStyle: pubData.citationInlineStyle,
 		nodeLabels: pubData.nodeLabels,
+		...(primaryCollection?.kind === 'issue' && { citeAs: pubData.community.citeAs }),
```
3mcd, line 73: We might not need this line anymore If we implement my suggested changes to the frontmatter render functions.

3mcd (changes requested):
Really great work @qweliant. I left just a few comments, but it looks great on the whole!

client/containers/DashboardSettings/CommunitySettings.tsx
```diff
@@ -222,6 +228,65 @@ const CommunitySettings = () => {
...
+								content={
+									<span>
+										When filled out, this field will be used as the journal
+										title for
+										<br />
+										issue collections in citations, PDF exports, and Crossref
+										deposits.
+										<br />
+										By default, the Community Title is used.
+									</span>
```
gabestein, line 245: Sorry for being sloppy when I wrote the copy! Just a few things to capitalize in the text that I should have seen the first time. Also, I'll defer to @3mcd on this, but my sense is we shouldn't be using line breaks, as they're unpredictable at different monitor sizes. Instead, we should make sure the tooltips have a good width set, or something? But I could be totally wrong.

```When filled out, this field will be used as the Journal Title for Issue Collections in citations, PDF exports, and Crossref deposits.```

File: client/containers/DashboardSettings/CommunitySettings.tsx [hunk 10]
```diff
@@ -222,6 +228,65 @@ const CommunitySettings = () => {
...
+								}
+								// @ts-expect-error ts-migrate(2322) FIXME: Type '{ children: Element; content: Element; toolt... Remove this comment to see the full error message
+								tooltipClassName="bp3-dark"
+							>
+								<Icon icon="info-sign" />
+							</Tooltip>
+						</span>
+					}
+					type="text"
+					placeholder="This will show up in citations and BibTex as your publisher name"
```
gabestein, line 284: I'm tempted to say we should remove the placeholder text, since it isn't exactly accurate, and is more fully described by the tooltip.

client/containers/DashboardSettings/CommunitySettings.tsx
```diff
@@ -222,6 +228,65 @@ const CommunitySettings = () => {
...
+								}
+								// @ts-expect-error ts-migrate(2322) FIXME: Type '{ children: Element; content: Element; toolt... Remove this comment to see the full error message
+								tooltipClassName="bp3-dark"
+							>
+								<Icon icon="info-sign" />
+							</Tooltip>
+						</span>
+					}
+					type="text"
+					placeholder="This will show up in citations and BibTex as your journal name"
```
gabestein, line 255: I'm tempted to say we should remove the placeholder text, since it isn't exactly accurate, and is more fully described by the tooltip.

client/containers/DashboardSettings/CommunitySettings.tsx
```diff
@@ -222,6 +228,65 @@ const CommunitySettings = () => {
...
+							<Tooltip
+								content={
+									<span>
+										When filled out, this field will be used as the publisher
+										for book and
+										<br />
+										conference proceedings collections in
+										<br />
+										citations, PDF exports, and Crossref deposits.
+									</span>
```
gabestein, line 274: Same thing here, sorry. Sloppy copy by me. And same question for @3mcd about the use of line breaks.

```When filled out, this field will be used as the Publisher for Book and Conference Proceedings Collections in citations, PDF exports, and Crossref deposits.```

**Review by gabestein — CHANGES_REQUESTED** [review]
Don't look until Monday!!




Very nice, we're getting super close. In addition to Eric's changes, a few functional things:

## UX
Added some tiny copy tweaks to the tooltips inline. Otherwise, looking good!

## Publisher
- Citation looks good!
- Crossref preview/XML looks good for books!
- Crossref preview/XML needs to be updated for conference proceedings to use the publisher value.
- As you noted in https://github.com/pubpub/pubpub/pull/1502/files#r675123929, all 3 should types of collections should have publisher HTML metadata if filled in, and we should also have the same behavior for the `dc.publisher` metadata field.
- Let's chat next week about export design...I think we'll just want to add the publisher to the cover page (for both PDF and non-PDF), rather than in the header.

## Cite As
- Citation looks good!
- Metadata looks good!
- Crossref preview/XML looks good!
- Exports still seems to be using the site title, rather than the Cite As field.

File: server/utils/citations/generateCitationHtml.ts (hunk 3 above)
Comment (qweliant) [reply server/utils/citations/generateCitationHtml.ts:69]: Quick question. @3mcd pointed out that the function `getCollectionLevelData` seems to be setting the `container-title` for all collections that are not an issue. This seems to conflict with `container-title` being set only for the issue. Can you clarify is we are interpreting this incorrectly or if this is an issue we should address. 

the function in question
```
const getCollectionLevelData = (collection) => {
	if (!collection) {
		return { type: 'article-journal' };
	}
	const { kind, title } = collection;
	const metadata = collection.metadata || {};
	const useCollectionTitle = collection.kind !== 'issue';
	return {
		type: collectionKindToCitationJSPart(kind),
		...(useCollectionTitle && { 'container-title': title }),
		containerDoi: getCollectionDoi(collection),
		ISBN: metadata.isbn,
		ISSN: metadata.issn || metadata.printIssn || metadata.electronicIssn,
		edition: metadata.edition,
		volume: metadata.volume,
		issue: metadata.issue,
	};
};
```

how it is used
```
	const commonData = {
		// @ts-expect-error ts-migrate(2783) FIXME: 'type' is specified more than once, so this usage ... Remove this comment to see the full error message
		type: 'article-journal',
		title: pubData.title,
		...authorsEntry,
		'container-title': communityData.citeAs || communityData.title,
		...getCollectionLevelData(primaryCollection),
		publisher: communityData.publishAs || '',
	};
```

**Comment by qweliant** [comment]
will also add migrations

File: server/utils/ssr.tsx [hunk 13]
```diff
@@ -121,6 +130,11 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 				<meta key="c5" name="citation_inbook_title" content={collection.title} />,
 				<meta key="c6" name="citation_book_title" content={collection.title} />,
 				<meta key="c7" name="citation_isbn" content={collection.metadata?.isbn} />,
+				<meta
```
gabestein, line 133: sorry, I think I got this wrong. This should appear for all types of collections, not just books.

gabestein (changes requested):
Okay, we are very, very close! Just two small things that I think should be pretty quick.

## Publisher
- In Crossref previews/XML, the publisher is still appearing as "PubPub" for conference proceedings. I think you'll just need to update `utils/crossref/schema/conference.js` similar to how you updated `book.js`.
- I probably messed this up somewhere, but the `citation_publisher` meta tag should be set for conferences, books, and journals. Right now it's only being set for books. (Context: meta tags are for Google Scholar and Zotero, which care about that field for all types, vs. crossref, which only wants it for books and conferences).

## Cite As
- All good here!

server/utils/ssr.tsx
```diff
@@ -121,6 +130,11 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
...
+					key="c8"
+					name="citation_publisher"
+					content={communityPublisher || 'PubPub'}
+				/>,
```
gabestein, line 137: sorry, I think I must have gotten this wrong. This should apply to all types of collections.

qweliant: the meta tag for publisher should be set on every collection, but publisher should only appear in exports for books and conferences in the exports and crossrefs?

gabestein: I know how crazy that sounds. But that is correct. The HTML metadata gets read by Google Scholar and Zotero and a few others, who do care about publisher for journals.

But crossref's schema doesn't have a publisher for journals, apparently, so here we are.

workers/tasks/export/html.tsx
```diff
@@ -166,12 +166,14 @@ const renderFrontMatterForPandoc = (
...
 		licenseSlug,
+		publisher,
 	},
 	targetPandoc,
 ) => {
 	const pandocFormatsWithoutTemplate = ['docx', 'plain', 'odt'];
+	// do not put community title if this is a book
 	const communityAndCollectionString =
-		communityTitle + (primaryCollectionTitle ? bullet + primaryCollectionTitle : '');
+		(!publisher ? communityTitle + bullet : '') + (primaryCollectionTitle || '');
```
3mcd, line 176: Super small thing, but it's a good practice to try and keep conditionals/ternaries positive, e.g. instead of saying "if there is no publisher, do Y, otherwise X", the code should read "if there is a publisher, do X, otherwise Y". Don't let this block your PR from moving forward though!

3mcd (approved):

workers/tasks/export/html.tsx
qweliant: i just changed it, trying to be different for no reason 😅

gabestein (approved):
Looking good to me! Thanks for persevering, and sorry this ended up being such a complex one!

PR #1506, opened 2021-07-19
Pub tools quick fix

gabestein:
- Changes `depositCollectionPubs` to only deposit released pubs in a collection (I think trying to deposit unreleased pubs would fail at crossref processing time given no public URI). There's probably a better way to do this selection at query-time, but after battling Sequelize for 3 hours I gave up. Pointers welcome.
- Fixes the pubCrawl tool to spit out a real subdomain so you can click from the terminal when running it

_Test plan_
- Run `depositCollecitonPubs` against a collection with both released and unreleased Pubs (e.g. "aaaaaa" on demo). Make sure it only processes and deposits released pubs (the number spit out in the terminal should correspond to the number of released pubs).
- Run `pubCrawl` against a community _without_ a custom domain. Make sure the full subdomain (e.g. http://demo.pubpub.org...) shows up in any undeposited Pubs (depositing a Pub on a dev server community should create this scenario, so I think using demo should work).

### DIFF
--- a/tools/depositCollectionPubs.js
+++ b/tools/depositCollectionPubs.js
@@ -1,7 +1,7 @@
 /* eslint-disable no-console */
 import Bluebird from 'bluebird';
-
-import { Community, Collection, CollectionPub } from 'server/models';
+import { Sequelize } from 'sequelize';
+import { Community, Collection, CollectionPub, Pub, Release } from 'server/models';
 import { setDoiData } from 'server/doi/queries';
 
 const {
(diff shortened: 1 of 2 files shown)


tools/depositCollectionPubs.js
```diff
@@ -19,14 +18,36 @@ const main = async () => {
...
+				],
+			},
+		],
+	});
+	const releasedPubs = collectionPubs.reduce((acc, cp) => {
+		if (cp.pub.releases.length > 0) {
+			acc.push(cp);
+		}
+		return acc;
+	}, []);
```
3mcd, line 43: A matter of preference but this could be simplified a little with `filter` e.g.:
```ts
const releasedPubs = collectionPubs.filter(cp => cp.pub.releases.length > 0);
```

tools/pubCrawl.js
```diff
@@ -19,7 +19,7 @@ const crawl = async (pub) => {
 		if (response.status === 404) {
 			console.log(
 				`DOI did not resolve: ${pub.title} • https://${pub.community.domain ||
-					pub.community.subdomain}/pub/${pub.slug}`,
+					pub.community.subdomain + `.pubpub.org`}/pub/${pub.slug}`,
```
idreyn, line 22: FYI there is a function called `communityUrl()` in `utilitys/canonicalUrls` that will handle this for you!

tools/depositCollectionPubs.js
```diff
@@ -19,14 +18,36 @@ const main = async () => {
...
+					{
+						model: Release,
+						as: 'releases',
+						attributes: ['id'],
+					},
+				],
+			},
+		],
+	});
+	const releasedPubs = collectionPubs.reduce((acc, cp) => {
```
idreyn, line 38: Some naming-related pedantry — `releasedPubs` is still a list of `CollectionPub` objects, so perhaps:
- The name should reflect that, e.g. `collectionPubsWithReleases`, or
- The `reduce` should collect the Pubs: `acc.push(cp.pub)`

idreyn (approved):
Looks good to me! (modulo some nitpicking in the comments)

tools/depositCollectionPubs.js
gabestein: Thanks! I always forget about filter. I always search map and reduce and forget filter.

PR #1503, opened 2021-07-09
fix: more robust community check to avoid impersonation, phishing, and other hacks

qweliant:
addresses issue #1429 

This fixes an issue where bad actors can impersonate members of communities they are not apart of. We now verify that a user has contributed to a community before their user page can be shown. If they are not, they will be redirected to the main pubpub page. 

Test:

1. Unit test for user redirect
2. Go to a pubpub community(https://www.pubpub.org/explore) you are not part of. At the end of the community homepage url type: /user/<firstname-lastname>. If you are not a member, this should redirect you back to the main pubpub site

PR #1516, opened 2021-07-26
fix: more robust community check to avoid impersonation, phishing, an…

qweliant:
…d other hacks (#1503)

* add check file

* added check to routes

* format with prettier

* all work and no play makes jack a dull boy

* added test, moved function, +14 XP

* make it purrty

* this should not push

* this should push becaus e its prettier

* format names

* communityId/userId gives

* manual fix like a pleeb

* add requested changes

* added chanfes frfr

### DIFF
--- a/server/routes/user.tsx
+++ b/server/routes/user.tsx
@@ -6,12 +6,22 @@ import { getUser } from 'server/utils/queryHelpers';
 import { handleErrors } from 'server/utils/errors';
 import { getInitialData } from 'server/utils/initData';
 import { generateMetaComponents, renderToNodeStream } from 'server/utils/ssr';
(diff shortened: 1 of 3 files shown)


server/routes/user.tsx
```diff
@@ -6,12 +6,22 @@ import { getUser } from 'server/utils/queryHelpers';
 import { handleErrors } from 'server/utils/errors';
 import { getInitialData } from 'server/utils/initData';
 import { generateMetaComponents, renderToNodeStream } from 'server/utils/ssr';
+import { isUserAffiliatedWithCommunity } from 'server/community/queries';
+// check if member has any pubs or collections for user
```
idreyn, line 10: Please remove this comment since it's no longer attached to the function it describes!

idreyn (approved):

PR #1518, opened 2021-07-27
update: Automatically create ActivityItems when things happen on PubPub

### LINKED ISSUE #1437: Add callsites for ActivityItem creation helpers
With #1436 under our belt it will be time to actually populate the backend with calls to our `ActivityItem` creation helpers so that they will be created by user activity. We hope to make heavy use of Sequelize lifecycle hooks here.

idreyn:
Resolves #1437

This PR adds code which automatically creates `ActivityItems` at the appropriate time. This is mostly accomplished using the following pattern:

- We create Sequelize hooks for model creation/updating/destruction (using a `createActivityHooks` utility). These call some of the activity item creation helpers (e.g. `createPubReviewUpdatedActivityItem`) which mostly have the same function signature.
- When calling query methods on capital-M Sequelize Model objects (e.g. `Pub.update()`) we need to pass a new piece of context into the options object: the `actorId` of the user causing the change.
- We may also need to pass the `individualHooks` option so that Sequelize runs the `update`/`destroy` hooks when these methods are called. It turns those off by default because in theory those operations could affect many rows.
- For `ActivityItems` created on Sequelize `update` hooks, we use the very questionable looking `model._previousDataValues` object to get the model's values before the update occurred, for diffing purposes. This, uh, might be a private field? But [it's repeatedly referenced in the Sequelize documentation](https://sequelize.org/master/class/lib/model.js~Model.html#instance-method-previous) in a way that suggests its exclusion from the public API is something of an oversight. If it ever disappears, we can trivially shim it using the keys from `model.dataValues` and the values from `model.previous()`, so I feel at least somewhat justified using it here.

I don't want it to get lost in the weeds here that using Sequelize hooks is a potentially controversial design decision. We could instead choose to call the activity item creation helpers directly from query functions like `updatePub`. This would add what I consider to be a lot of noisy boilerplate code to these functions — particularly where `defer()` is involved — but it would also be clearer to readers how these items are created, and it would mean we would not have to pass `actorId` and `individualHooks` into Sequelize queries to cause some spooky action at a distance. In practice I'm worried that future refactors may drop some of these, though I'm confident the changes I've made to the unit tests would catch that.

_Test plan:_
Just run the whole dang thing:
```
npm run test-dev server/
```
The `Review` model doesn't have a unit test suite. You can manually verify that the act of creating, commenting on, or releasing/closing a Review creates activity items.


server/member/hooks.ts
```diff
@@ -0,0 +1,14 @@
...
+	createMemberCreatedActivityItem,
+	createMemberUpdatedActivityItem,
+	createMemberRemovedActivityItem,
+} from 'server/activityItem/queries';
+
+createActivityHooks({
+	Model: Member,
+	onModelCreated: createMemberCreatedActivityItem,
+	onModelUpdated: createMemberUpdatedActivityItem,
+	onModelDestroyed: createMemberRemovedActivityItem,
```
3mcd, line 13: Super minor nit (feel free to ignore) – having a combination of generic `createXActivityItem` and more specific `createXUpdatedActivityItem` functions introduces a decision point to either write logic in the hooks files vs. the queries files which might result in split logic in the future e.g. someone might add logic to the anonymous functions in the hooks.ts file instead of keeping it all in queries.ts.

server/server.ts
```diff
@@ -12,12 +12,13 @@ import path from 'path';
 import { setEnvironment, setAppCommit, isProd, getAppCommit } from 'utils/environment';
 import { HTTPStatusError, errorMiddleware } from 'server/utils/errors';
 
+import 'server/utils/serverModuleOverwrite';
```
3mcd, line 15: :ok_hand:

server/utils/activityHooks.ts
```diff
@@ -0,0 +1,34 @@
...
+	onModelUpdated?: (
+		actorId: null | string,
+		modelId: string,
+		previousModel: Model,
+	) => Promise<void>;
+	onModelDestroyed?: (actorId: null | string, modelId: string) => Promise<void>;
+};
+
+export const createActivityHooks = <Model>(options: CreateActivityHooksOptions<Model>) => {
+	const { Model, onModelCreated, onModelUpdated, onModelDestroyed } = options;
```
3mcd, line 15: Super minor and feel free to ignore, but the generic parameter `Model` and constant of the same name `Model` reduces the readability of this function just a bit. Could potentially make the option lowercase?

server/utils/activityHooks.ts
```diff
@@ -0,0 +1,34 @@
+import { defer } from 'server/utils/deferred';
+
+type CreateActivityHooksOptions<Model> = {
+	Model: any;
```
3mcd, line 4: Could this be like

```ts
import { Model as SequelizeModel } from 'sequelize'

type CreateActivityHooksOptions<Model> = {
  Model: typeof SequelizeModel
}
```

instead of `any`?

server/utils/deferred.ts
```diff
@@ -0,0 +1,40 @@
...
+		}
+	};
+
+	const finishDeferredTasks = (): Promise<void> => {
+		return new Promise((resolve) => onDeferredTasksComplete(resolve));
+	};
+
+	const defer = (task: DeferredTask) => {
+		incrementCount();
+		task().then(decrementCount);
```
3mcd, line 34: I guess `defer` will be in the call stack so it should be clear that an uncaught error was the result of a deferred task. But we could add a catch here to be more deliberate about it? idk

3mcd (approved):
Looks really good! All the feedback I left is minor so feel free to ignore.

server/utils/activityHooks.ts
idreyn: I went back and forth on casing here but I didn't even realize that I was using the same name twice here, so yeah, it should totally be lowercase.

server/utils/activityHooks.ts
idreyn: This would be a helpful clue to the reader, but I don't know if it would provide any type safety, since TS treats all the model singletons as `any`. Maybe there's a `@types/sequelize` that could help us here?

server/utils/deferred.ts
idreyn: I'm into the idea of doing something here, though I don't know what would be most appropriate. Catch and print the stack trace?

3mcd: Yeah I'm not exactly sure, maybe something like:

```ts
.catch(err => {
  console.error("An error occurred in a deferred task:")
  console.error(err)
})
```

Would at least give us a string to grep for if we needed it?

server/utils/activityHooks.ts
3mcd: This could be totally wrong, but I was thinking that if we do `typeof SequelizeModel`, it would capture the type of the complete model class. I tried something that seems to kinda work below... but it's funky. We'd at least get type checking for stuff like `afterCreate` etc. I'm sure there's a better way to do this.

Screen Shot 2021-07-27 at 3 41 34 PM

idreyn: Ah, that's extremely useful to know about! In some other branch I have been building up my own `SequelizeModel<T>` that augments `T` with ORM methods like `.toJSON()` It seems I might be able to use this instead. Two problems remain:

1. The model singleton has the `afterCreate` etc. methods (e.g. `Pub` instead of a `pub`). I don't know if Sequelize has a type for that.
2. `Model` is apparently actually generic: `Model<T, T2>`, but neither type argument seems to provide the instance-specific properties when I try it (i.e. `Model<{id: string}>` doesn't have an `id` property). I think that's a bug.

I did end up adding some better types to this to document our expectations. Feel free to weigh in on these; I'm worried it came out a little overwrought.

PR #1529, opened 2021-08-10
fix: Prevent layout editor on Pages dashboard from rendering when it doesn't need to

idreyn:


howebenhowe (approved):
Reads clean. Combing through, I can see how your choices to wrap in `memo` and render-preventing `useCallback` will prevent unnecessary labor, but I have no idea how you identified those as the places to take the extra measure of efficiency. I shall trouble you for your secrets elsewhere.

PR #1533, opened 2021-08-10
fix: add line height to footnote

### LINKED ISSUE #1491: Footnote reference breaks general line-height.
Adding `line-height: 1;` would fix it.

https://github.com/pubpub/pubpub/blob/f913b01c1cdef9a35189e6ba41446bbd9faa9ad2/client/components/Editor/styles/main.scss#L36

gabestein:
Sets the line height for footnotes, so that they don't add height to interior lines of paragraphs. Fixes #1491.

Before:
Screen Shot 2021-08-10 at 17 48 11

After:
Screen Shot 2021-08-10 at 17 48 18

_Test Plan_
1. Visit any Pub with a footnote on an interior (e.g. not the first) line of a paragraph. Make sure that line is not bigger vertically than surrounding lines.
2. Export the Pub to a PDF and make sure the footnote also looks good in the PDF export.

### DIFF
--- a/client/components/Editor/styles/main.scss
+++ b/client/components/Editor/styles/main.scss
@@ -36,6 +36,7 @@ h6 {
 .footnote {
 	vertical-align: super;
 	font-size: 0.85em;
+	line-height: 1;
 }
 
 span.citation {


client/components/Editor/styles/main.scss
```diff
@@ -36,6 +36,7 @@ h6 {
 .footnote {
 	vertical-align: super;
 	font-size: 0.85em;
+	line-height: 1;
```
idreyn, line 39: I didn't know you could do, this — I've been in the unfortunate habit of repeating the same constant between `font-size` and `line-height` for this purpose.

idreyn (approved):
👍

client/components/Editor/styles/main.scss
gabestein: neither did I! A user suggested it!

Issue #1473, opened 2021-06-14
Activity digest email body rendering


idreyn:
Separate from the questions of what goes into Community activity digests or their precise appearance, we need to determine what tools we'll use to render them to HTML+CSS. The end products of this would be:

- [ ] A decision on whether to use React to render our emails, or some other templating system
- [ ] A decision about what email clients we want to support and thus which CSS features are available to us — we should find out what industry best practices dictate here and also get @gabestein to sign off on this decision
- [ ] Based on the above, a decision about how CSS will be added to emails — do the email clients we want to support allow us a `<style>` tag in the email body? Or if not, could we use a library like [`inline-css`](https://www.npmjs.com/package/inline-css) to automatically inline CSS (i.e. in element `style` attributes)? Or if not, do we have to do this by hand? Keep in mind we will need to inject Community-specific colors into the emails.
- [ ] Taking all of this into account, a decision on whether we can/should use existing components in emails, e.g. `PubPreview` to represent a Pub and `Avatar` to represent a User. Alternatives might be copying the styles for these components into an email-related part of the codebase, or using a completely different set of styles in emails. If we reuse existing components, we'll need a way to generate CSS from the import tree of those components, for instance by injecting a custom `require` into scope that collects a list of all referenced `.scss` files (@idreyn can elaborate). This sounds fun but is possibly not worth the time.

With all of this in hand we then want:

- [ ] A server-side email rendering utility that (maybe) uses `react-dom` to render a React component for an email body to an HTML string, or maybe uses some other templating system we deem better
- [ ] A proof-of-concept for generating an emailable HTML+CSS bundle from a small subset of `ActivityItems`

PR #1536, opened 2021-08-12
Email components with inline styling

howebenhowe:
Addresses #1473

+ Renders email as static HTML with inline CSS.
+ Aims at visual parity with [these mocks](https://www.figma.com/file/RUmAt42roXLhNobGwd9yfa/email-digest-draft?node-id=25%3A639)
+ Sets framework for further email templating
+ preview of email for a community can be viewed at `:communitybaseurl/email/:emailTemplateSlug`, for example: `localhost:9876/email/digest`


client/components/Email/render.tsx
```diff
@@ -0,0 +1,24 @@
...
+import beautify from 'js-beautify';
+
+import { reset, globals } from './styles';
+
+const inlineStylesWithMarkup = (EmailMarkup: React.ReactNode, extraStyles: string) => {
+	const stylesheet = new ServerStyleSheet();
+	const renderedStringFromEmailMarkup = ReactDOMServer.renderToStaticMarkup(
+		<StyleSheetManager sheet={stylesheet.instance}>{EmailMarkup}</StyleSheetManager>,
+	);
+	const basicMarkup = beautify.html(renderedStringFromEmailMarkup);
```
idreyn, line 14: My sense is that we'd want to minify rather than beautify this HTML before sending it to people. I can see why this is useful for testing, though.

client/components/Email/render.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import ReactDOMServer from 'react-dom/server';
+import { ServerStyleSheet, StyleSheetManager } from 'styled-components';
+import juice from 'juice';
+import beautify from 'js-beautify';
+
+import { reset, globals } from './styles';
+
+const inlineStylesWithMarkup = (EmailMarkup: React.ReactNode, extraStyles: string) => {
```
idreyn, line 9: Any reason this is uppercase? In my opinion it should not be unless it is to indicate that the argument is, e.g. a JavaScript class, React component definition, Sequelize model, etc. (as opposed to an instance of these things)

idreyn (approved):
Not looking too closely at the actual styles here since they'll probably change when designs come through, but this seems like a great approach. You raise some interesting questions:

> How to test the output as we code

Back when I worked at Khan Academy we had an internal kitchen-sink utility dashboard with a page that let you preview rendered emails without sending them. It would be cool to have something like that someday; for now, if this need presents itself, we could just have a standalone quasi-internal route that takes some GET parameters for the scope and renders the email the logged-in user would see if it were to be sent to them immediately?

> How to pull in blueprintjs css for injection alongside email-specific css?

This is an interesting question, but I think we may be able to get around it by not including any Blueprint styles. I think @djagdish is amenable to this, as I sense him nudging us towards a house style for buttons and things. The existing Pub preview components don't use any Blueprint design elements either. Icons specifically may be a problem: those are rendered as inline SVGs, but there is a little bit of Blueprint CSS that makes them look right.

client/components/Email/render.tsx
howebenhowe: When scoping this project out, I encountered someone saying that email clients don't like minified code, and that explicit beautification helps them get along. I shall probe further on the beautify/minify question.

idreyn:
Does this need to be merged into the activity emails branch, or is has all of this code made it in with your more recent PR?

Issue #1357, opened 2021-04-20
Don't allow deletion of Home Page.


isTravis:
Right now users can delete the default "Home" Page (the page that resolves at `community.pubpub.org/`). 

When they do this, it's impossible to create a new Page that resolves on the landing page (i.e. the `/` route). 

I think the simplest fix is to simply prohibit the Home Page from being deleted. 

An example of this can be seen in [this ticket](https://knowledgefutures.freshdesk.com/a/tickets/3312)

PR #1538, opened 2021-08-12
disallow homepage deletion

howebenhowe:
Addresses #1357.

Still possible at the database level to delete a homepage. This PR changes the available user interface on the page dashboard, replacing the deletion dialogue if the page in question is a home page for its community.

Depending on how bullet proof we need to be, we could also reject deletion with a beforeDestroy hook, but I liked just denying the user the opportunity to delete and arming them with a heads-up on why.

To test: open the  demo community's homepage dashboard (titled HOME!!!!!!) any one other of its pages' dashboards, compare the deletion control at the bottom of of the dash:
Screenshot from 2021-08-12 19-49-00


client/containers/DashboardPage/PageDeleteForbidden.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const PageDeleteForbidden = (props: Props) => {
+	const {
+		pageData: { title },
+	} = props;
+	return (
+		<div className="bp3-callout bp3-intent-danger">
+			<h5>Delete Page from Community</h5>
+			<p>
+				<b>{title}</b> cannot be deleted because it is the home page for this community.
```
idreyn, line 17: "community" should be capitalized in user-facing copy.

client/containers/DashboardPage/PageDeleteForbidden.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+type Props = {
+	pageData: {
+		title?: string;
+	};
+};
+
+const PageDeleteForbidden = (props: Props) => {
```
idreyn, line 9: Personally I would consider merging this functionality with `PageDelete` by adding an `isForbidden` prop to that component and either rendering the existing contents or this new notice based on its value. Otherwise any stylistic changes to that component have to be duplicated into this one.

idreyn (approved):
LGTM

client/containers/DashboardPage/PageDeleteForbidden.tsx
howebenhowe: So sensible.
I shall act accordingly.

PR #1540, opened 2021-08-16
Bh typescript conversion

howebenhowe:
Move all .js and .jsx files in email components dir to .ts and .tsx
Install types for styled-components, js-beautify, react-dom.
Appease TS errors for potential null values in react-dom function calls.


idreyn (approved):
LGTM. I know I still have to go back and review the base PR as well.

Issue #1537, opened 2021-08-12
Journal citations should only apply to journal issue collections


gabestein:
### What went wrong, step-by-step?
1. Add a journal citation
2. It applies to the citation, metadata, deposit, and export of every non-book, non-conference collection in the community.

### What did you expect to happen?
It applies only to pubs in journal issue collections, so that e.g. blog posts in a random tag collection aren't cited/deposited as part of the journal.

### What browser/operating system were you using?

===

PR #1549, opened 2021-08-26
fix: Citation data rendering

qweliant:
this addresses issue #1537 

When rendering information for crossref, metadata, exports, and citations the journal name is being created for all collections. The set journal name should only apply to issues. I have written changes so that the expected behavior happens for issues journal title/citation-journal-title/container-title/title.  

Test plan(the only thing that changed):

With journal citation set and in a primary issue collection, create a crossref deposit for a pub. Be sure the journal metadata title is set to the journal citation name. Export the pub, be sure the header includes the journal name. Look at the citations in pub overview or details and be sure the Journal Name is listed. Use the inspector to make sure the the citation-journal-title is set to the journal name.

With Journal not set and in a primary issue collection, create a crossref deposit for a pub. Be sure the journal metadata title is set to the community name. Export the pub, be sure the header includes the community name. Look at the citations in pub overview or details and be sure the community name is listed. Use the inspector to make sure the the citation-journal-title is set to the community name.

This should be the same for any pub not in a book, issue, or, conference 

While in a book or conference, and with the publisher name set, create a crossref deposit for a pub. Be sure the book metadata title is set to the collection name. Export the pub, be sure the subheader includes the publisher name. Look at the citations in pub overview or details and be sure the publisher name is listed. Use the inspector to make sure the the citation-book-title is set to the collection name. 

When publisher name is not set these values should be the same, with 'PubPub' replacing publisher name

### DIFF
--- a/server/utils/citations/generateCitationHtml.ts
+++ b/server/utils/citations/generateCitationHtml.ts
@@ -73,7 +73,6 @@ export const generateCitationHtml = async (pubData, communityData) => {
 	};
 	const pubCiteObject = await Cite.async({
 		...commonData,
-		// @ts-expect-error ts-migrate(2339) FIXME: Property 'containerDoi' does not exist on type '{ ... Remove this comment to see the full error message
 		DOI: pubData.doi || commonData.containerDoi,
 		ISSN: pubData.doi ? communityData.issn : null,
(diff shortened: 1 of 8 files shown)


3mcd (approved):
LGTM!

PR #1555, opened 2021-09-07
fix: allow about page to scope to community

qweliant:
about page was in list of reserved URLs, now it's not.
This should allow communities to create default about pages.


gabestein:
Thank you!

I think we can also remove:

- explore
- pricing
- landing

gabestein (approved):
LGTM!

Issue #1546, opened 2021-08-23
Enhancement: Give all communities custom CSS access; limit JavaScript


gabestein:
### Summary
Right now, communities with IDs hardcoded in `server/customScript/permissions.ts` can access custom CSS. Instead, we would like all communities to be able to access custom CSS. However, currently the feature is built to enable both custom CSS and custom JS. For security and legal reasons, we should only expose custom CSS, leaving custom JS only for the communities hardcoded.

### Requirements
- All communities can access and edit custom CSS
- When a community does not have ability to edit JS, no "CSS/JS" tabs are shown at the top of the scripts dashboard
- Only communities specified in code can add custom JS
- As before, only Community Admins can edit either CSS or JS
- Community settings has a link, in the details section, to customize CSS, but with a warning:

![custom_css_settings](https://user-images.githubusercontent.com/639110/130470033-63ece25d-19e1-4a14-86db-15908225cf35.png)

- The scripts dashboard page also provides a warning

![script_dash_warning](https://user-images.githubusercontent.com/639110/130470051-5392a65a-d6b5-46ef-b6ad-57b005a08135.png)

PR #1557, opened 2021-09-08
enhancement: Custom css for communities

qweliant:
addresses issue #1546 

Adds custom css for all communities. Community admins can now access a custom scripts editor via the settings dashboard. 

image

image

Test plan
go to community settings as an admin
scroll to custom css link, hover over ⚠️ symbol and read the tooltip disclaimer 
click on custom css link
most community admins should only see the css tab. if you are hardcoded to have js access you will see both

run test suite locally for custom scripts

### DIFF
--- a/client/containers/DashboardCustomScripts/DashboardCustomScripts.tsx
+++ b/client/containers/DashboardCustomScripts/DashboardCustomScripts.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { Spinner, Tabs, Tab } from '@blueprintjs/core';
+import { Callout, Spinner, Tabs, Tab } from '@blueprintjs/core';
 
 import { communityCanUseCustomScripts } from 'utils/customScripts';
 import { DashboardFrame } from 'components';
@@ -20,7 +20,6 @@ const DashboardCustomScripts = (props: Props) => {
 	const { js, css } = customScripts;
 	const { id } = communityId;
 	const [Editor, setEditor] = useState<null | EditorComponentType>(null);
-
 	useEffect(() => {
 		import('@monaco-editor/react').then(({ default: EditorComponent }) =>
 			setEditor(EditorComponent),
@@ -72,6 +71,15 @@ const DashboardCustomScripts = (props: Props) => {
 
 	return (
 		<DashboardFrame className="dashboard-custom-scripts-container" title="Custom Scripts">
+			<div className="callout-text">
+				<Callout className="error-callout" intent="warning" title="Warning: Experimental">
+					We do not guarantee that future product updates will preserve existing page
+					structures of class names. Adding custom CSS can cause your community to be
+					unusable. We do not provide support for issues caused by CSS customization. Use
+					at your own risk.
+				</Callout>
+			</div>
+
 			{!Editor && renderLoading()}
 			{Editor && renderTabs()}
 		</DashboardFrame>
--- a/client/containers/DashboardCustomScripts/DashboardCustomScripts.tsx
+++ b/client/containers/DashboardCustomScripts/DashboardCustomScripts.tsx
@@ -1,8 +1,9 @@
 import React, { useEffect, useState } from 'react';
 import { Spinner, Tabs, Tab } from '@blueprintjs/core';
 
+import { communityCanUseCustomScripts } from 'utils/customScripts';
 import { DashboardFrame } from 'components';
-import { CustomScripts } from 'types';
+import { CustomScripts, Community } from 'types';
 
 import { EditorComponentType } from './types';
 import CustomScriptPanel from './CustomScriptPanel';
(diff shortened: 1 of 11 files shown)


client/containers/DashboardCustomScripts/DashboardCustomScripts.tsx
```diff
@@ -11,13 +12,14 @@ require('./dashboardCustomScripts.scss');
 
 type Props = {
 	customScripts: CustomScripts;
+	communityId: Community;
 };
 
 const DashboardCustomScripts = (props: Props) => {
-	const { customScripts } = props;
+	const { customScripts, communityId } = props;
```
idreyn, line 19: We have a handy `usePageContext()` hook that aggregates a lot of useful information about the page the user is currently visiting, including the Community. So you'll never have to pass that info down from the server by hand — instead you can grab the ID like this:

```ts
const { communityData: { id: communityId }} = usePageContext();
```

We should really convert `utils/hooks` to TypeScript and add a `PageContext` type to help clarify what this contains, but I encourage you to take a look around at callsites for this hook to get a sense for its various usess.

client/containers/DashboardCustomScripts/dashboardCustomScripts.scss
```diff
@@ -9,4 +9,8 @@
     .bp3-tab-panel {
         height: calc(100vh - 300px);
     }
+    .callout-text {
+        max-width: 500px;
+        line-height: 15px;
```
gabestein, line 14: let's add a little bottom padding here (20px?). It feels a little cramped.

Also, I don't think we need the max width. I did the mockup that way only because I was lazily copying and pasting UI elements in screenshots, but it looks better and more warning-y, I think, all the way across the screen.

gabestein (changes requested):
Just a copy update and a couple of small ux things from me.

client/containers/DashboardSettings/CommunitySettings.tsx
```diff
@@ -342,6 +342,24 @@ const CommunitySettings = () => {
 						/>
 					</InputField>
 				</div>
+				<div className="row-wrapper">
+					<a href="/dash/scripts">Customize CSS</a>
```
idreyn, line 346: I think we should replace this with `getDashUrl({ mode: 'scripts' })` so this doesn't break if we change the dashboard URL scheme.

server/routes/dashboardCustomScripts.tsx
```diff
@@ -21,13 +21,13 @@ app.get('/dash/scripts', async (req, res, next) => {
 		}
 
 		const customScripts = await getCustomScriptsForCommunity(communityData.id);
-
+		const communityId = { id: communityData.id };
```
idreyn, line 24: As mentioned above the `usePageContext()` hook (which, n.b., is reading from the `initialData` we pass in here) makes it unnecessary to pass this down here.

idreyn (changes requested):
This largely looks great to me! I'm requesting changes primarily to use the `usePageContext()` hook instead of passing down `communityId`, but I'd also appreciate if you took a pass at de-inlining some of the added CSS styles.

client/containers/DashboardSettings/CommunitySettings.tsx
```diff
@@ -342,6 +342,24 @@ const CommunitySettings = () => {
...
+					<div style={{ paddingLeft: '5px' }}>
```
idreyn, line 347: Let's avoid using inline CSS here, and use the class name props that `Tooltip` provides instead:

```tsx
<Tooltip
    targetClassName="custom-css-tooltip-target"
    popoverClassName="dashboard-settings-container_custom-css-popover"
    content={
        <>
            Warning: Experimental...
        </>
    }
>
```

Then you can add a `.custom-css-tooltip-target` selector inside of `.dashboard-settings-container` to add that 5px of left margin. You'll then need to add a _top-level_ selector `.dashboard-settings-container_custom-css-popover` to hold the 400px max-width property. This gets rendered in a React portal, so it's annoyingly not a child of `.dashboard-settings-container`. Its long class name with the `_` separating the two parts is a loose convention of ours for naming these portalled children-but-not-really.

In general when adding styles that are statically known, we should hew towards using SCSS rather than inline styles. I think we can bend this rule when it's exceedingly inconvenient; this is an example of where it's kinda inconvenient, but there are patterns in place to make it work :)

idreyn (approved):
Looks good to me! People are going to love this.

gabestein (approved):
LGTM. Thank you for this! Please feel free to post about it on the Discussion forums when it's live. Lots of folks will want to know.

qweliant:
> LGTM. Thank you for this! Please feel free to post about it on the Discussion forums when it's live. Lots of folks will want to know.

Wait...we have discussion forums? @gabestein

gabestein:
https://github.com/pubpub/pubpub/discussions :)

PR #1561, opened 2021-09-13
bug: fix potential phishing attacks

qweliant:
fixes issue #1543

An attacker can add a redirect link to the login redirect URL.

Test plan:
On the login page, add any number of forward slashes to the end of the login redirect url with some site after it.

ex:
http://localhost:9876/login?redirect=/////////////////google.com

http://localhost:9876/login?redirect=/pub/testing-this-out-man/release/1//spam.com

enter your login info.

you should see this screen

image

### DIFF
--- a/client/containers/Login/Login.tsx
+++ b/client/containers/Login/Login.tsx
@@ -43,8 +43,11 @@ const Login = () => {
 			}),
 		})
 			.then(() => {
-				window.location.href =
-					`/${trimStart(`${locationData.query.redirect}`, '/')}` || '/';
+				window.location.href = locationData.query.redirect
+					? `/${trimStart(locationData.query.redirect, '/')}`
+					: '/';
+				// window.location.href =
+				// 	`/${trimStart(`${locationData.query.redirect}`, '/')}` || '/';
 			})
 			.catch(() => {
 				setLoginLoading(false);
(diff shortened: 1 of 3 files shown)


client/containers/Login/Login.tsx
```diff
@@ -43,7 +43,11 @@ const Login = () => {
 			}),
 		})
 			.then(() => {
-				window.location.href = locationData.query.redirect || '/';
+				window.location.href = locationData.query.redirect
+					? `/${trimStart(locationData.query.redirect, '/')}`
```
isTravis, line 47: Nice - this works! The logic is right.

It's probably negligible, but looking at the [lodash.trimstart package](https://github.com/lodash/lodash/blob/4.5.1-npm-packages/lodash.trimstart/index.js), it seems like a lot of code we don't need for this simple, specific case (e.g. it has code to handle special unicode characters). We might save a couple kb on the js bundle we send to clients by using an inline regex. Regex has tricked me plenty of times before, but I'm pretty sure this works:
```
window.location.href = locationData.query.redirect 
    ? `/${locationData.query.redirect.replace(/^\//,"")}`
    : '/';
```

It would force us to use the if else rather than `||` because we don't want to call `.replace()` on undefined if there is no query string.

isTravis: Ha! You beat me too it on the if:else by a few seconds it seems! Good catch.

isTravis: All that said - if you or others think the lodash approach is more clear and maintainable for future developers, totally fine!

It is somewhat cumbersome that regex lines often need their own comment block just to explain their intent and make sure it's not misinterpreted.

qweliant: no i really did that trying to avoid writing it myself or getting tricked by regex but i'm glad you called it out. i know the bundle size is also approaching it's limits so i will try the regex way and if it works it works!

isTravis (approved):
🎉

3mcd (approved):

Issue #1560, opened 2021-09-13
Custom CSS does not apply to Pubs


ThomasFKJorna:
_Before filing your bug report, please search existing issues to make sure a similar one doesn't already exist. If one does, please leave a comment or reaction in support of the issue rather than creating a new one._

_To fill out the report, please leave the below headers (**) in tact, and fill out each section. If a section isn't applicable, leave it blank or write n/a under it. Please do not delete any template sections._

### What went wrong, step-by-step?
1. Use new shiny CSS feature to make all text a nice ugly shade of red, by adding to custom CSS
``` css
body { 
   color: red;
}
```
2. Open random Page, see glorious red text
3. Open random Pub, see boring black text

Actually what I tried to do was to shorten the DOI (client for some reason picked really long DOIs, causing the collection button not to display), but this does the same.

When loading a page, an inline stylesheet is loaded (usually called # 2), but when loading a pub this doesn't happen.

### What did you expect to happen?
I expected the CSS to also be loaded in Pubs. This is probably intended behavior, but since it's intended behavior I don't like I made it a bug-report :smile: . 

On a more serious note, I didn't see anything in the descriptions of the new feature which indicated that it was limited to pages only. Furthermore, I find it a bit strange to only limit this to pages: someone using custom CSS would probably make quite some changes to the default style, which would make the default Pub style look somewhat out of place. 
If it is in place as a "safety" feature I also find that somewhat strange, as the Custom CSS already has a ton of warnings saying "This could break your community" etc. Please let me break my community.

### What URL can we use to see the issue?
For the lengthy ass DOI, see https://openpresstiu.pubpub.org/pub/powerpolitical/release/5

### What browser/operating system were you using?
Firefox (latest), Arch Linux, but I don't think that really matters

===

### Have any screenshots? (Attach them by dragging the image here)
:D 
![image](https://user-images.githubusercontent.com/21983833/133117697-159f89ad-46ce-4fb3-9f01-eca6f84ec65d.png)

:( 
![image](https://user-images.githubusercontent.com/21983833/133117468-365f2eaa-d578-4827-aabd-1ffb595f6a8b.png)

gabestein:
Hi Thomas — thanks for pointing this out. I am inclined to agree, and we'll add styling to Pubs as well in the coming days.

PR #1562, opened 2021-09-13
dev: Apply css to pubs and add global className based on active element type

gabestein:
Okay, third time's the charm!

This does two simple things, resolving #1560:

- Adds custom scripts to Pubs
- Adds two classnames to the body of Pubs, Collections, and Pages (and no other objects):

1. `x-body-wrapper`. This is passed from the <Html> component in a new prop and is meant to describe the type of content on the page, primarily for allowing people to target their page layouts, generated app pages, and content (e.g. pubs) separately.

2. `active-[object]-[slug]`, where the object is `pub | collection | page` and the slug is the item's slug. This allows people to target specific Pubs, Collections, and Pages with CSS.

_Test Plan_
1. Make sure custom CSS is injected into Pubs
2. Check pages and make sure the body has the classes: 'layout-body-wrapper active-page-[slug]`
3. Check collections and make sure the body has the classes: 'layout-body-wrapper active-collection-[slug]`
4. Check pubs and make sure the body has the classes: 'pub-body-wrapper active-pub-[slug]`
5. Visit dashboard, login/logout, legal, and www pages and make sure they load and have no injected css or body class names.

_Post-Release_
Gabe to replace Commonplace and Contours CSS with a new version he made (with compiled scss!) that applies the styles only to pages so nothing changes.

I can see a future where we might want to add "app" body prefixes and inject css to community-owned/pubpub-controlled pages like Login/Logout/Signup/User/Legal. But for now, this feels like a good-enough extension.

### DIFF
--- a/server/Html.tsx
+++ b/server/Html.tsx
@@ -39,7 +39,7 @@ const polyfills = [
 
 type Props = OwnProps & typeof defaultProps;
 
-const activeSlug = (initialData: InitialData, viewData?: any) => {
+const getActiveSlugClassName = (initialData: InitialData, viewData?: any) => {
 	if (viewData?.pageData?.slug) {
 		return `active-page-${viewData.pageData.slug}`;
 	}
@@ -87,7 +87,7 @@ const Html = (props: Props) => {
 			<body
 				className={classNames(
 					props.bodyClassPrefix && `${props.bodyClassPrefix}-body-wrapper`,
-					activeSlug(props.initialData, props.viewData),
+					getActiveSlugClassName(props.initialData, props.viewData),
 				)}
 			>
 				{/* This script tag is here to prevent FOUC in Firefox: https://stackoverflow.com/questions/21147149/flash-of-unstyled-content-fouc-in-firefox-only-is-ff-slow-renderer */}
--- a/server/Html.tsx
+++ b/server/Html.tsx
@@ -43,8 +43,6 @@ const Html = (props: Props) => {
 	const getPath = (chunkName, extension) => {
 		return `${manifest[`${chunkName}.${extension}`]}`;
 	};
-	const activeTargetType = props.initialData.scopeData.elements.activeTargetType ?? 'community';
-
 	return (
 		<html lang="en">
 			<head>
@@ -76,7 +74,7 @@ const Html = (props: Props) => {
 			</head>
 			<body
 				className={`${props.bodyClassPrefix}-body-wrapper active-target-${props.initialData
-					.scopeData.elements?.activeTargetType || 'community'}`}
+					.scopeData.elements?.activeTargetType ?? 'community'}`}
 			>
 				{/* This script tag is here to prevent FOUC in Firefox: https://stackoverflow.com/questions/21147149/flash-of-unstyled-content-fouc-in-firefox-only-is-ff-slow-renderer */}
 				<script>0</script>
--- a/server/Html.tsx
+++ b/server/Html.tsx
@@ -43,6 +43,13 @@ const Html = (props: Props) => {
 	const getPath = (chunkName, extension) => {
 		return `${manifest[`${chunkName}.${extension}`]}`;
 	};
+
+	const activeTargetSlug =
+		props.initialData.scopeData.elements?.activeTarget.slug ||
+		(props.initialData.locationData.path.split('/')[1]
+			? props.initialData.locationData.path.substring(1).replace(/\//gi, '-')
+			: props.initialData.scopeData.elements?.activeTarget.subdomain || 'www');
+
 	return (
 		<html lang="en">
 			<head>
@@ -75,10 +82,7 @@ const Html = (props: Props) => {
 			<body
 				className={`${props.bodyClassPrefix}-body-wrapper active-target-type-${props
 					.initialData.scopeData.elements?.activeTargetType ??
-					'community'} active-target-slug-${props.initialData.scopeData.elements
-					?.activeTarget.slug ||
-					props.initialData.scopeData.elements?.activeTarget.subdomain ||
-					'www'}`}
+					'community'} active-target-slug-${activeTargetSlug}`}
 			>
 				{/* This script tag is here to prevent FOUC in Firefox: https://stackoverflow.com/questions/21147149/flash-of-unstyled-content-fouc-in-firefox-only-is-ff-slow-renderer */}
 				<script>0</script>
--- a/server/Html.tsx
+++ b/server/Html.tsx
@@ -73,8 +73,12 @@ const Html = (props: Props) => {
 				/>
 			</head>
 			<body
-				className={`${props.bodyClassPrefix}-body-wrapper active-target-${props.initialData
-					.scopeData.elements?.activeTargetType ?? 'community'}`}
+				className={`${props.bodyClassPrefix}-body-wrapper active-target-type-${props
+					.initialData.scopeData.elements?.activeTargetType ??
+					'community'} active-target-slug-${props.initialData.scopeData.elements
+					?.activeTarget.slug ||
+					props.initialData.scopeData.elements?.activeTarget.subdomain ||
+					'www'}`}
 			>
 				{/* This script tag is here to prevent FOUC in Firefox: https://stackoverflow.com/questions/21147149/flash-of-unstyled-content-fouc-in-firefox-only-is-ff-slow-renderer */}
 				<script>0</script>
--- a/server/Html.tsx
+++ b/server/Html.tsx
@@ -43,6 +43,7 @@ const Html = (props: Props) => {
 	const getPath = (chunkName, extension) => {
 		return `${manifest[`${chunkName}.${extension}`]}`;
 	};
+	const activeTargetType = props.initialData.scopeData.elements.activeTargetType ?? 'community';
 
 	return (
 		<html lang="en">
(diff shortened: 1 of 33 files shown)


server/Html.tsx
```diff
@@ -71,7 +71,7 @@ const Html = (props: Props) => {
 					href="/opensearch.xml"
 				/>
 			</head>
-			<body>
+			<body className={props.initialData.scopeData.elements.activeTargetType}>
```
idreyn, line 74: To summarize what I was saying on the dev call — `activeTargetType` will be `community` on Pages and `collection` on Collections, but most users will want to apply styles uniformly to both types of layout. Rather than making them type:
```css
.community .thing, .collection .thing {
    /* and stuff */ 
}
```
I propose a new prop on `<Html>` called `bodyClassPrefix` which if set to e.g. `layout` would yield a `body.layout-body-wrapper` item — wordy, but unambiguously not any of the other classes we have floating around relating to layouts. This should be set to the same value on Pages and Collection layouts, and then something else on Pubs.

I'm all for including `activeTargetType` as e.g. `active-target-pub` as well if that seems like it will be helpful.

idreyn (changes requested):
Thanks for putting this together. I think we should use something other than `activeTargetType` (details inline) to disambiguate pages, but overall I like the approach.

server/Html.tsx
gabestein: Thanks, good suggestion. Take a look at the rewritten PR.

server/Html.tsx
```diff
@@ -42,6 +44,12 @@ const Html = (props: Props) => {
 		return `${manifest[`${chunkName}.${extension}`]}`;
 	};
 
+	const activeTargetSlug =
```
idreyn, line 47: This is too much logic for a single expression. I rewrote this as a helper function to get a handle on what is does:

```ts
const getActiveTargetSlug = (iniitalData: InitialData) => {
	const { scopeData, locationData } = iniitalData;
	const { activeTargetType, activePub, activeCollection, activeCommunity } = scopeData.elements;
	if (activeTargetType === 'community') {
		if (locationData.path === '/') {
			return locationData.isBasePubPub ? 'www' : activeCommunity.subdomain;
		}
		return locationData.path.split('/')[1];
	}
	if (activeTargetType === 'collection') {
		return activeCollection!.slug;
	}
	return activePub!.slug;
};
```

This more clearly breaks down into the different scope cases we need to address, but it also raises a concern for me. If `activeTargetType` is `community`, I would expect that this slug would be always be the Community's subdomain (a bit of a name kludge, but whatever) but in fact it will be a Page slug everywhere except the homepage.

More likely we want class names like `active-page-<slug>` or `active-collection-<slug>` that explicitly differentiate between these entities, and prevent quiet accidents when there's a slug collision across Pubs/Collections/Pages. It would also probably be more proper to provide HTML attributes like `data-active-page` that can be targeted with CSS, but I understand that this may be less ergonomic.

server/Html.tsx
```diff
@@ -42,6 +44,12 @@ const Html = (props: Props) => {
...
+		props.initialData.scopeData.elements?.activeTarget.slug ||
+		(props.initialData.locationData.path.split('/')[1]
+			? props.initialData.locationData.path.substring(1).replace(/\//gi, '-')
```
idreyn, line 50: Can you provide an example of where this replacement might take effect? I see that it would turn `/foo/bar` into `foo-bar`, but I don't know where we'd see that happen in our URL scheme, and if it did I worry that it would quietly collide with something that actually has the slug `foo-bar`.

server/Html.tsx
```diff
@@ -42,6 +44,12 @@ const Html = (props: Props) => {
...
+			: props.initialData.scopeData.elements?.activeTarget.subdomain || 'www');
```
idreyn, line 51: Depending on how you look at it, this `www` is a bit of a red herring since we'll never load custom CSS on pubpub.org.

server/Html.tsx
```diff
@@ -71,7 +79,11 @@ const Html = (props: Props) => {
 					href="/opensearch.xml"
 				/>
 			</head>
-			<body>
+			<body
+				className={`${props.bodyClassPrefix}-body-wrapper active-target-type-${props
```
idreyn, line 83: Prefer the `classNames()` utility to string interpolation for long expressions like this.

server/routes/about.tsx
```diff
@@ -23,6 +23,7 @@ app.get('/about', (req, res, next) => {
 						initialData,
 						title: 'About PubPub',
 					})}
+					bodyClassPrefix="www"
```
idreyn, line 26: Do we need this if it's never going to be referenced in custom CSS?

server/routes/dashboardActivity.tsx
```diff
@@ -41,6 +41,7 @@ app.get(
 						title: `Activity · ${initialData.scopeData.elements.activeTarget.title}`,
 						unlisted: true,
 					})}
+					bodyClassPrefix="dashboard"
```
idreyn, line 44: Ditto here — the dashboard won't ever see custom CSS so we might as well omit the prop in that case.

server/routes/pubDocument.tsx
```diff
@@ -50,6 +52,7 @@ const renderPubDocument = (res, pubData, initialData) => {
 				title: pubData.title,
 				unlisted: !pubData.isRelease,
 			})}
+			bodyClassPrefix="content"
```
idreyn, line 55: Curious about the choice of this versus `pub`.

server/routes/signup.tsx
```diff
@@ -19,6 +19,7 @@ app.get('/signup', (req, res, next) => {
 						title: `Signup · ${initialData.communityData.title}`,
 						description: initialData.communityData.description,
 					})}
+					bodyClassPrefix="app"
```
idreyn, line 22: Here too I am unsure whether we need/want to provide a class that will never be targeted by custom CSS.

idreyn (changes requested):
This is pretty neat but I have a few concerns:

1. The logic that derives an `active-target-slug-<x>` class is semantically muddled — the `x` is not always a slug, and does not always refer to the object we call the `activeTarget`. More to the point of usability, it opens users up to the possibility of silent collisions between e.g. Pubs and Pages with the same slug that could have strange stylistic consequences.
2. I'm not sure that most of the `bodyClassPrefix` additions here (`www`, `dashboard`, `app`) should be provided if users are not able to target them with custom CSS. I also question the choice of `content` instead of `pub` on the Pub route.

Let's talk about this more at DOO, but my sense is that we can get a quick win here by dropping the ambitious `active-target-slug` stuff, removing the `bodyClassPrefix` props that can never be targeted by custom CSS, and using the `classNames` utility to add some structure, e.g.

```tsx
className={classNames(bodyClassPrefix && `${bodyClassPrefix}-body-wrapper`, `active-target-type-${activeTargetType}`)}
```

gabestein:
Thanks @idreyn! I agree with your two suggestions for making this simpler, with one exception: I'd like to keep a css class with slugs in place for objects that have them. Our test users are already doing this (using javascript based on the route). So, why don't I give that a go for just pages, pubs, and collections, based on your suggestion?

idreyn:
Totally, go ahead!

server/routes/pubDocument.tsx
gabestein: yeah, I'm being too clever.

server/routes/signup.tsx
gabestein: I agree for www and dashboard. But I can see app pages being targeted by custom CSS in the future (e.g. you want to make footer mods across your site). We don't have to do it with this project, though.

server/Html.tsx
gabestein: q for you: where should this helper function go in practice? Is it okay to leave in Html.tsx for now, or should I put it in utils? I can't really see it being that useful outside of Html, but you never know.

idreyn: Definitely in `Html.tsx`...by convention we put these after the imports and type definitions but before the component itself.

gabestein:
ok, third time's the charm, perhaps?

server/Html.tsx
```diff
@@ -36,6 +39,16 @@ const polyfills = [
 
 type Props = OwnProps & typeof defaultProps;
 
+const activeSlug = (initialData: InitialData, viewData?: any) => {
+	if (viewData?.pageData?.slug) {
```
idreyn, line 43: This is Probably Fine™ but it will break silently if we ever change the `pageData` prop to something else. Still definitely cleaner than doing URL parsing.

server/Html.tsx
```diff
@@ -36,6 +39,16 @@ const polyfills = [
 
 type Props = OwnProps & typeof defaultProps;
 
+const activeSlug = (initialData: InitialData, viewData?: any) => {
```
idreyn, line 42: Maybe `getActiveSlug` (or more accurately, `getActiveSlugClassName`)? I don't have an automatic preference for prefixing helper functions with "get" but I do often find myself in this kind of trouble:

```ts
const activeSlug = activeSlug(...); // oh wait
```

so I tend to start off with "get" more often these days.

idreyn (approved):
Looks fantastic, thanks for iterating on this!

server/Html.tsx
gabestein: Yeah...ideally, we add pages to scopeData, right? For now, don't know a better way to do it.

3mcd (approved):

PR #1563, opened 2021-09-15
Edit Existing Pub Edges

3mcd:
This PR adds the ability to edit existing Pub Connections.

### UI changes

Connections are edited using a new 'Edit' button in the top right corner of a connection card in the connections dashboard page:

Screen Shot 2021-10-04 at 10 25 09 AM

Clicking this button will swap the read-only view of the connection to an editable view which functions identically to the NewEdgeEditor, with a couple of small changes (e.g. the 'Add connection' button is renamed to 'Save connection'):

Screen Shot 2021-10-04 at 10 27 37 AM

### Metadata changes

Crossref's documentation states that in order to update metadata (e.g. relationships) you must first deposit content with empty metadata, then follow-up with a deposit containing the updated metadata. This PR also implements [this behavior](https://github.com/pubpub/pubpub/pull/1563/files#diff-3a45b33c886df538ea597c6daab024b1d2b5d6e0cfee2dce334669e6ec93e179R153).

In addition, for Pubs that already have been deposited, users are notified that updating a Pub's connections will also require a re-deposit for the connections to update in Crossref.

## Test plan

- Add a connection to a Pub. The Pub should have at least one release and a DOI.
- Go to the Pub's connection settings. You should see a notification that modifying the Pub's connections will require a manual re-deposit.
- Edit the connection using the new 'Edit' button. The connection should be updated and persisted.
- Go to the DOI section in the Pub's settings. The deposit preview should reflect the changes made to the Pub, i.e., the relationships should be updated.
- Deposit the Pub. Log in to https://test.crossref.org. You should see two deposit records in the queue: one with no relationship metadata, and one with the updated relationships.

-- 
~~WIP, just putting this here for visibility/early feedback.~~

~~My approach so far has been to refactor the `NewEdgeEditor` into a presentational component, while hoisting the fetch/async logic higher up into the dashboard. I then use the `NewEdgeEditor` within the `DasboardEdgesListing` component to swap an outbound edge component with the editor when an "edit" button is pressed.~~

~~Things left to do:~~
- [x] ~~Solicit feedback on approach/design~~
- [x] ~~Inform user that the updated Pub and connected Pubs will need to be redeposited~~
- [x] ~~Update deposit pipeline to first deposit with empty connection metadata, then follow up with a deposit with the newly updated metadata (according to guidelines in the [Crossref docs](https://www.crossref.org/documentation/metadata-stewardship/maintaining-your-metadata/updating-your-metadata/))~~


client/containers/DashboardEdges/DashboardEdgesListing.tsx
```diff
@@ -1,12 +1,18 @@
...
 import classNames from 'classnames';
+import React, { useState } from 'react';
+import { DragDropContext } from 'react-beautiful-dnd';
+import { PubEdge } from 'types';
+
+import { usePendingChanges } from 'utils/hooks';
+import { apiFetch } from 'client/utils/apiFetch';
 
 import { ConfirmDialog, DragDropListing, PubEdgeListingCard } from 'components';
+import NewEdgeEditor from 'containers/DashboardEdges/NewEdgeEditor';
```
idreyn, line 11: This should just be `./NewEdgeEditor` (and fwiw, the `components` import above should be in the previous section)

utils/crossref/createDeposit.js
```diff
@@ -185,6 +185,11 @@ export default (context, doiTarget, dateForTimestamp) => {
 			RelationType.Review,
 			RelationType.Rejoinder,
 		]);
+
+		if (!includeRelationships) {
```
idreyn, line 189: What if we

```ts
let pub = context.pub && {...context.pub};
```
above so we're not mutating an argument?

utils/assert.ts
```diff
@@ -0,0 +1,9 @@
+class AssertionError extends Error {
+	name = 'AssertionError';
+}
+
+export function assert(condition: boolean, message = ''): asserts condition {
```
idreyn, line 5: Very cool!

utils/canonicalUrls.js
```diff
@@ -71,6 +71,12 @@ export const bestPubUrl = ({ pubData, communityData }, options = {}) => {
 	return pubShortUrl(pubData);
 };
 
+export const pubSettingsUrl = (community, pub, section) => {
```
idreyn, line 74: We already have a `getDashUrl` that'll do this for ya. The invocation you want is:

```ts
getDashUrl({ pubSlug: pub.slug, mode: 'settings' })
```

client/containers/DashboardEdges/newEdgeEditor.scss
```diff
@@ -48,4 +49,6 @@ $margin-size: 15px;
 		margin-top: $margin-size;
 		margin-bottom: $margin-size;
 	}
+
+	margin-bottom: 30px;
```
idreyn, line 53: This is the same `30px` that's present in `PubEdgeListingCard`, right? I'm tempted to say it would be cleaner to remove both this one and that one (leaving no margin between cards on either the dashboard or the Pub-side listing) and then adding
```scss
container > * { 
    margin-bottom: 30px;
}
```
for some `container` in each of `DashboardEdgesListing` and `PubEdgeListing`. I say "tempted" because now that I've written it out I'm not sure it's actually cleaner, except that I like the idea of the parent owning the child spacing.

Food for thought, basically, feel free to ignore entirely.

idreyn (changes requested):
Looks really good! Specifically requesting changes to use `getDashUrl` as described inline.

client/containers/DashboardEdges/newEdgeEditor.scss
3mcd: I like your idea, and I actually tried to do the same thing before and it just got a bit hairy. Definitely happy to do in a follow-up PR.

idreyn (approved):
LGTM!

Issue #1509, opened 2021-07-22
Chinese characters (and other CJK?) not appearing in PDF exports


idreyn:
Via @Dawit-Tegbaru. Probably down to the font we're using — we may want to list additional fallbacks. Or it could be an issue requiring a Paged.js upgrade (see #1356)

gabestein:
As discussed this AM, we believe the route is simply to add Han to our Adobe bundle (@djagdish). We may potentially need to include it in our `printStyles.css` body font-family as well, but we're not sure if that's even needed.

idreyn:
Okay so here's my latest on this. I checked in on this line in `printDocument.scss`:

```scss
@import url('https://use.typekit.net/kmi0tdo.css');
```

This URL tuns out to be specific to our `pubpub-v6-prod` "Web Project" on Adobe Fonts which is visible [here](https://fonts.adobe.com/my_fonts?manage_fonts_filter=all&max_styles=26&min_styles=1#web_projects-section). In theory this means we should be able to add more fonts to this project and they'll automatically get added to this bundle, without making any changes to our source code.

However, the Adobe CJK webfonts come with a complication. Because their character sets are so big, Adobe uses something called [dynamic subsetting](https://helpx.adobe.com/fonts/using/dynamic-subsetting.html) to load only the glyphs required to display the actual content of a page. This requires embedding a snippet of JavaScript to read page content and request more glyphs from Adobe as appropriate.

It seems like we cannot retrofit the `pubpub-v6-prod` project to use dynamic subsetting. We would have to create another project and add [Source Han Serif](https://fonts.adobe.com/fonts/source-han-serif-simplified-chinese) in the right flavors, along with our other fonts. 

## Option 1: New web project with dynamic subsetting

This may or may not work. There are two questions to answer:

1. Does our export pipeline support dynamic subsetting? To find out, we can add the JavaScript snippet from the `cjk-testing` Web Project I created to this file: https://github.com/pubpub/pubpub/blob/454965c90718180dcbb976e1a423a4f396f822b0/workers/tasks/export/html.tsx#L299 using the same `dangerouslySetInnerHtml` technique. This script _should_ get run by the headless Chrome instance that Paged.js uses to produce our PDF exports, but it's an open question whether it will load the fonts to completion before rendering the PDF. Some of the [font display](https://helpx.adobe.com/fonts/using/font-display-settings.html) options may help us here.
[...]

PR #1575, opened 2021-09-29
bug: CJK fonts render for export

qweliant:
addresses issue #1509

CJK fonts cannot be loaded via bundle on Heroku because they are huge. Having dynamic subsetting allows fonts to be loaded during export as opposed to resting on the Heroku bundle. 

before
image

after
image

Test:
Add Korean, Japanese, and Chinese characters in header of pub doc
Add Korean, Japanese, and Chinese characters in body of pub doc

verify that all characters a re rendered

heres a pub with all three
https://demo.duqduq.org/pub/t0tkgi74/draft

### DIFF
--- a/workers/tasks/export/html.tsx
+++ b/workers/tasks/export/html.tsx
@@ -14,7 +14,9 @@ const nonExportableNodeTypes = ['discussion'];
 const katexCdnPrefix = 'https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.11.1/';
 const bullet = ' • ';
 
-const createCJKFontScript = `
+// This script is provided by the "cjk-fonts" Web Fonts project that we manage from here:
+// https://fonts.adobe.com/my_fonts#web_projects-section
+const loadCjkFontsScript = `
 (function(d) {
   var config = {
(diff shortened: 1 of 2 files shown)


workers/tasks/export/html.tsx
```diff
@@ -14,6 +14,17 @@ const nonExportableNodeTypes = ['discussion'];
 const katexCdnPrefix = 'https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.11.1/';
 const bullet = ' • ';
 
+const createCJKFontScript = `
```
idreyn, line 17: Could you add a comment to the effect of:

```
This script is provided by the "cjk-fonts" Web Fonts project that we manage from here:
https://fonts.adobe.com/my_fonts#web_projects-section
```
so that future readers know where it came from?

I might also call the variable `loadCjkFontsScript` since "load" feels like a more accurate verb than "create".

idreyn (approved):
LGTM, thanks for powering through this!

Dawit-Tegbaru:
Thanks guys! 🙏🏽

Issue #1551, opened 2021-08-30
Upgrade katex to 0.13.x


gabestein:
There have been a number of improvements to KaTeX since we last upgraded, but also a few breaking changes. I think they will mostly be positive, but we'd want to do a bit of testing on HDSR in particular.

Summary of breaking changes: https://github.com/KaTeX/KaTeX/issues/2490

Other cool stuff:

Error messages in 0.12.0
Pure MathML display mode
Matrix support
Align(at/at*/edat/edat*) support
Equation support!
More LaTeX-y color support

PR #1576, opened 2021-09-29
dev: bump katex to 0.13.18

gabestein:
Bumps KaTeX to 0.13.18, resolving #1551 and making our LaTeX support much better.

I did basic QA on a number of HDSR articles with lots of LaTeX, particularly: http://localhost:9876/pub/fxz7kr65/release/6 and https://localhost:9876/pub/y9vc2u36/release/7.

Luckily, it doesn't look like this will break our implementation of KaTeX in any way. In the future, we may want to take advantage of some new features like the improved error messages in particular, and we will want to make the fix described in #1577 to take full advantage of the new features. But that's beyond the scope of this PR.

There were a few breaking changes in KaTeX's support of LaTeX, however, and I searched the Docs table in the db for Pubs that may have these issues. Once we release, we'll want to fix these pubs

### Macro token expansion
Very hard to search for, but I didn't find anyone using `\newcommand` or `\def`, so I think our exposure to this kind of token expansion will be very limited, if at all.

### `\newline` and `\cr` --> \\
The following pubs use `\cr` and will need to be manually converted to `\\`:

- cerje44v
- arrih5vx
- arrih5vx

### `\cfrac`, `\color`, `\textcolor`, `\colorbox`, `\fcolorbox`
There was one use of \color, but it didn't use it as an argument, so there were no mods to make. I tested them on localhost with the bumped KaTeX to be sure.

_Test Plan_
1. Create a new Pub and add some LaTex, making sure it still works as expected.
2. Browse some LaTeX-heavy Pubs (HDSR, mostly) and make sure no errors appear (red).
3. Export a fresh Pub with tex as PDF and HTML and make sure it appears correctly

_Post release_
Modify those three pubs and re-release

### DIFF
--- a/workers/tasks/export/html.tsx
+++ b/workers/tasks/export/html.tsx
@@ -11,7 +11,7 @@ import SimpleNotesList from './SimpleNotesList';
 import { digestCitation } from './util';
 
 const nonExportableNodeTypes = ['discussion'];
-const katexCdnPrefix = 'https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.11.1/';
+const katexCdnPrefix = 'https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.13.18/';
 const bullet = ' • ';
 
 const createCss = () => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -29454,17 +29454,17 @@
 			"integrity": "sha512-40aUybvhH9t2h71ncA1/1SbtTNCVZHgsTsTgqPUxGWDmUDrXyDf2wMNQKEbdBjbf4AI+fQhbECNTV6lWxQKUzg=="
 		},
 		"katex": {
-			"version": "0.11.1",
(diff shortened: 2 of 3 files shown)


workers/tasks/export/html.tsx
```diff
@@ -11,7 +11,7 @@ import SimpleNotesList from './SimpleNotesList';
 import { digestCitation } from './util';
 
 const nonExportableNodeTypes = ['discussion'];
-const katexCdnPrefix = 'https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.11.1/';
+const katexCdnPrefix = 'https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.13.18/';
```
idreyn, line 14: Great catch. I wish we had a better way of packaging this.

idreyn (approved):
Hard to argue with a thorough test plan and a `package.json` update :) Excited to see this in users' hands!

PR #1584, opened 2021-10-04
Update exporter to use prosemirror-pandoc and produce better JATS

idreyn:
PubPub exports Pubs to the [JATS](https://en.wikipedia.org/wiki/Journal_Article_Tag_Suite) format for consumption by archiving services and the like. We have long wanted to improve the fidelity of our JATS output, but have been limited by the fact that we export to JATS by rendering our Prosemirror document to HTML and then using Pandoc to convert that to JATS, losing some semantic information (particularly around citations) along the way.

This PR is a companion to our work on [`prosemirror-pandoc@1.0.0`](https://github.com/pubpub/prosemirror-pandoc/releases/tag/1.0.0) that finally allows us to convert directly from Prosemirror _to_ Pandoc (we have had the reverse pathway in production for a few years). By circumventing the HTML step we can now render to JATS (and Markdown, LaTeX, and Word, though these formats are of secondary concern to us) with much higher fidelity. This includes elements like:

- Math
- Abstracts
- Citations and bibliographies

Along the way we've eliminated Prosemirror -> HTML -> Pandoc export codepaths, which simplifies a number of things. Quite a bit of the remainder of this changeset is dedicated to codifying and passing around citation data. We use a library called Citation.js to render citations from a "structured" input like a DOI to a number of formats such as APA, MLA, etc. One such output format is a serializable CSL-JSON, which we ultimately pass into Pandoc as a separate bibliography JSON file. Mapping citation elements in the document itself to entries in this bibliography takes some kid gloves and ultimately requires us to hash citations to ensure each one has a quasi-unique ID.

I am tracking loose ends related to this, uh, project on [this board](https://github.com/pubpub/pubpub/projects/6). Test plan TK. Gabe and I are going to put this thing through the ringer with a number of sample files to make sure it's ready to go.


workers/tasks/import/util.ts
```diff
@@ -1,79 +1,6 @@
 import path from 'path';
 import fs from 'fs-extra';
 import { spawn as spawnChildProcess } from 'child_process';
-import {
```
idreyn, line 4: There are versions of these exported from `prosemirror-pandoc` that we can use instead.

workers/tasks/import/rules.ts
```diff
@@ -1,49 +1,50 @@
-import { buildRuleset, transformers, transformUtil } from '@pubpub/prosemirror-pandoc';
-import { defaultNodes, defaultMarks } from 'components/Editor/schemas';
-
 import * as katex from 'katex';
```
idreyn, line 1: This file is actually an important part of the changeset despite being hidden by default from Github.

Issue #1574, opened 2021-09-24
Better collection metadata blocks


gabestein:
### What's your enhancement request?
Display better/more full collection metadata (particularly for books, conferences, and journal issues) in a standard way, including google scholar/etc. metadata.

### What would this enhancement help you accomplish?
- Make it easier to layout books, issues, and conferences.
- Allow aggregators/indexers to properly index books in particular

---

### Who requested this? (If not you)
Thomas Journa/Jeff Pooley

johav:
this will be indeed useful as we (africarxiv) need the stats of submissions per repository per collection per country. thanks :)

PR #1595, opened 2021-10-06
enhancement: Add metadata block for collections layout editor

qweliant:
addresses issue #1574 

Admins can now add metadata for collections without manually writing them out

image

image

Test plan:
Create a collection (tag collections do not have metadata)

Add metadata to the collection

Navigate to the layout tab

Select the Header fields dropdown to make sure the Byline, Contributors, Collection Kind, Creation date and DOI if present are toggling on and off appropriately. Save the changes as you make them and view them on the collection homepage.

Select the Metadata fields dropdown to make sure the fields present are toggling on and off appropriately. Save the changes as you make them and view them on the collection homepage.


gabestein (changes requested):
Thanks Qwelian, looking pretty close. A few things from me on the cosmetic side, although one thing I'm realizing is that we should probably throw a fair amount of this into a citation engine eventually. But we'll get there. For now:

- DOI should be in the metadata fields dropdown, not the details fields dropdown
- Order of the metadata dropdowns and how the elements are displayed should match the order of the collection metadata fields in settings. (e.g. for journals it should be: DOI, Print ISSN, Electronic ISSN, Volume, Issue, Print Publication Date, Publication Date). This is particularly important for Volume and Issue, which almost always go together, but right now are separated.
- For print publication date, let's call it "Printed" and, per Catherine's suggestion, use the same more friendly date format that we use for the Created date. So in the example it should read: "Printed Feb 02, 2020"
- For the publication date, it should use the same format, so should be "Published Dec 20, 2020"
- Volume and Issue should always be in order next to each other, Volume first and then Issue, both in the dropdown and in the display. They should also be prepended with the names, so it should say "Volume [X] • Issue [Y]"
- Print and Electronic ISSNs should be prepended, e.g. "ISSN: [TEXT]" or "e-ISSN: [TEXT]"
- Conference date should not be prepended with anything, and should match the more friendly display date format above, e.g. "Dec 1, 2020"
- ISBN should be prepended, e.g. "ISBN: [TEXT]"
- Copyright year should be prepended, e.g. "Copyright © 2015"
- Publication date should be prepended and displayed in the friendly format, e.g. "Published Dec 1, 2020"
- Edition number should be appended, e.g. "23 ed."

utils/collections/getIssueMetadata.ts
```diff
@@ -0,0 +1,54 @@
+import { Collection } from 'types';
+
+const getIssuePrintIssn = (collection) => {
+	if (collection.metadata.printIssn) {
+		return collection.metadata.printIssn;
+	}
+	return collection.metadata && collection.metadata.printIssn;
+};
```
3mcd, line 8: I love the use of small functions here, but if I understand their purpose, we might be able to simplify their implementation a bit.

Starting with the return statement, it looks like we're assuming that a collection may not have a `metadata` property. If that's the case, the first line of the function (line 4) is cause for concern, because if `metadata` is undefined, we'll have a `TypeError: Cannot read property of undefined` before we ever get to line 7.

These functions really just access a deeply nested property, returning `undefined` if an object in the chain is also undefined. The easiest way to simplify while preserving the original intent might be to use the [Optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) like so:

```ts
const getIssuePrintIssn = (collection) => collection.metadata?.printIssn;
```

client/components/LayoutEditor/layoutEditor.scss
```diff
@@ -46,6 +46,19 @@
...
 		}
+		.controls-row {
+			display: flex;
+			margin: 0 0.5em;
+			& > .input-field-component {
+				margin: 0;
+			}
+			& > *:not(:last-child) {
+				margin-right: 0.5em;
+			}
```
3mcd, line 57: I think the `&` might be redundant here. You should be able to remove them for these two rulesets, but not the `&:not` below.

client/components/LayoutEditor/LayoutEditorCollectionHeader/getMetaDataChecklist.tsx
```diff
@@ -0,0 +1,205 @@
...
+					collection={collection}
+				/>
+			)}
+			{collection.kind === 'conference' && (
+				<PreviewConferenceElementFields
+					content={content}
+					onChange={onChange}
+					collection={collection}
+				/>
+			)}
```
3mcd, line 200: You can maybe simplify this like (we can chat about this later):
```ts
const metadataViewsByCollectionKind = {
  issue: PreviewIssueElementFields,
  book: PreviewBookElementFields,
  conference: PreviewConferenceElementFields,
}
const Metadata = (props: Props) => {
  // ...
  const MetadataView = metadataViewsByCollectionKind[collection.kind]
  return <MetadataView content={content} onChange={onChange} ... />
}
```

client/components/LayoutEditor/LayoutEditorCollectionHeader/getMetaDataChecklist.tsx
```diff
@@ -0,0 +1,205 @@
...
+				checked={location ? !hideLocation : false}
+				onChange={() => onChange({ hideLocation: !hideLocation })}
+				label="Location"
+			/>
+			<Checkbox
+				disabled={!conferenceDate}
+				checked={conferenceDate ? !hideConferenceDate : false}
+				onChange={() => onChange({ hideConferenceDate: !hideConferenceDate })}
+				label="Conference Date"
+			/>
```
3mcd, line 170: Some repetition in these components that we might be able to handle programmatically, lets pair program something during our chat later?

client/components/Layout/layoutCollectionHeader.scss
```diff
@@ -5,87 +5,88 @@ $row-spacing: 10px;
```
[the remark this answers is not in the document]
qweliant: it was flex-wrap: wrap; used to wrap flexed content

utils/collections/getMetadata.ts
```diff
@@ -0,0 +1,28 @@
...
+	if (field === 'publicationDate') return `Published ${formatDate(data)}`;
+	if (field === 'edition') return `${data} ed.`;
+
+	if (field === 'theme') return `${data}`;
+	if (field === 'acronym') return ` ${data}`;
+	if (field === 'location') return `${data}`;
+	if (field === 'date') return `${formatDate(data)}`;
+
+	return data;
+};
```
3mcd, line 28: It might be a bit more efficient to make an object like:

```ts
export const metadataFormatters = {
	printIssn: data => `ISSN: ${data}`,
	electronicIssn: data => `e-ISSN: ${data}`,
	volume: data => `Volume ${data}`,
	issue: data => `Issue ${data}`,
	printPublicationDate: data => `Printed ${formatDate(data)}`,
	publicationDate: data => `Published ${formatDate(data)}`,

	isbn: data => `ISBN: ${data}`,
	copyrightYear: data => `Copyright © ${data}`,
	publicationDate: data => `Published ${formatDate(data)}`,
	edition: data => `${data} ed.`,

	theme: data => `${data}`,
	acronym: data => ` ${data}`,
	location: data => `${data}`,
	date: data => `${formatDate(data)}`,
}
```

And get the formatted value like `metadataFormatters[field](data)` instead.

types/collection.ts
```diff
@@ -14,7 +14,7 @@ export type Collection = {
 	viewHash?: string;
 	editHash?: string;
 	metadata?: { [k: string]: any };
-	kind?: 'tag' | 'issue' | 'book' | 'conference';
+	kind: 'tag' | 'issue' | 'book' | 'conference';
```
3mcd, line 17: 👍

client/components/Layout/LayoutCollectionHeader.tsx
```diff
@@ -17,10 +21,43 @@ type Props = {
 	content: LayoutBlockCollectionHeader['content'];
 };
 
+const MetadataDiv = (props: Props) => {
```
3mcd, line 24: Since this component doesn't render a `div`, I think there is room to name it more clearly. Maybe something like `MetadataList`, `MetadataFields`, `MetadataDict`, etc.

PR #1596, opened 2021-10-06
fix: Don't block forever on Cite.async

idreyn:
There is currently [an issue with Crossref](https://twitter.com/CrossrefSupport/status/1445788491275587584) that prevents Citation.js from loading structured citation data — in turn, this is causing some Pubs to not load. We should consider caching this data — a quicker fix is to let these citations error out if they do not load quickly enough.

_Test plan:_
Create a new Pub and add a citation with a DOI as its structured data. Refresh the page and see that it loads.


gabestein (approved):
Thank you!

PR #1606, opened 2021-10-12
update to prosemirror-reactive@0.2.0

idreyn:
So @3mcd will remember that about a year ago, we introduced the `prosemirror-reactive` library to solve a bunch of our Prosemirror state management issues with a React-like library for stateful computation inside of node attributes. I also introduced a kind of clunky API for accessing the library's built-in hooks by retrieving them from the `this` dynamically bound to the the hook function:

```ts
function useTicker(this: Hooks) {
	const { useState, useEffect } = this;
	const [ticks, setTicks] = useState(0);

	useEffect(() => {
		const interval = setInterval(() => setTicks((t) => t + 1), 1000);
		return () => clearInterval(interval);
	}, []);

	return ticks;
}
```

TS forces you to annotate the type of `this` to use it, which obviates the cleverness of using `this` instead of a function argument. It also required you to manually pass the `Hooks` object around to children, and to use the `function` keyword instead of arrow functions when writing these hooks. [An upstream change](https://github.com/pubpub/prosemirror-reactive/pull/4) now exports built-in hooks as globals, the same way React does, allowing for a new usage:

```ts
import { useState, useEffect } from '@pubpub/prosemirror-reactive';

const useTicker = () => {
	const [ticks, setTicks] = useState(0);

	useEffect(() => {
		const interval = setInterval(() => setTicks((t) => t + 1), 1000);
		return () => clearInterval(interval);
	}, []);

	return ticks;
}
```

The changes to the PubPub repo in this PR are updates to our usages of `prosemirror-reactive` to take advantage of this change.

_Test plan:_
I created a new Pub and verified the following features are working:
- Reference node labelling
- Node counting
- Structured data rendering in citations and footnotes

...which gives me a good sense that reactive attributes are working as expected.


3mcd (approved):
LGTM!

Issue #1607, opened 2021-10-12
Create tags from Pub header


gabestein:
Talk to Gabe, Ian, and Deepak.

PR #1627, opened 2021-10-18
enhancement: Create Tag Collection from Pub Header

qweliant:
addresses issue #1607 

Users can now create and add tag collections from the pub header
image
image

Test:

go to a pub of your choice and click the plus icon in the header. enter the name of a non existent collection. select the new item or hit enter. the new collection should now exist along with the new pub being part of the collection


client/components/PubCollectionsListing/PubCollectionsListing.tsx
```diff
@@ -29,6 +28,9 @@ type Props = {
 	renderTriggerButtonForQueryList?: () => React.ReactNode;
 	renderDragElementInPortal?: boolean;
 };
+type TagToCreate = {
+	tagNameToCreate: string;
```
idreyn, line 32: This should actually be `tagTitleToCreate` since Collections have "titles" not "names". Sorry for the misdirection!

client/components/PubCollectionsListing/PubCollectionsListing.tsx
```diff
@@ -203,12 +231,14 @@ const PubCollectionsListing = (props: Props) => {
 					itemPredicate={(query, collection) => fuzzyMatchCollection(collection, query)}
 					items={canAddCollections}
 					itemRenderer={renderAvailableCollection}
-					emptyListPlaceholder="No Collections match this search."
-					searchPlaceholder="Search for Collections"
+					emptyListPlaceholder="Create new collection?"
```
idreyn, line 234: This should be `Create a new Tag with this name`

client/utils/collections/api.ts
```diff
@@ -1,6 +1,17 @@
 import { apiFetch } from 'client/utils/apiFetch';
 
 const collectionPubsRoot = '/api/collectionPubs';
+const collection = '/api/collections';
+
+export const createTagCollection = ({ title, communityId }) =>
+	apiFetch(collection, {
+		method: 'POST',
+		body: JSON.stringify({
```
idreyn, line 9: I wonder if we should pass `isPublic: true` here to make these public by default — maybe create a parameter for this too. Paging @gabestein

idreyn (changes requested):
This is looking good so far. The last piece of this is that we need to avoid showing this new UI to users who don't actually have permission to create Collections — you can check `scopeData.activePermissions.canManageCommunity` to see whether that's the case.

client/utils/collections/api.ts
gabestein: yup, I think that's right.

qweliant: @idreyn what do you mean by create a parameter? 

is this what you are getting at? Why if it is always true?
`export const createTagCollection = ({ title, communityId, isPublic }) =>
	apiFetch(collection, {
		method: 'POST',
		body: JSON.stringify({
			communityId,
			title,
			kind: 'tag',
			isPublic,
		}),
	});`

qweliant:
> This is looking good so far. The last piece of this is that we need to avoid showing this new UI to users who don't actually have permission to create Collections — you can check `scopeData.activePermissions.canManageCommunity` to see whether that's the case.

My bad, I assumed only admins have permission to add a pub to a collection and that it was taken care of with line 231

`if (canAddCollections.length > 0) { `

should i just conditionally render a `QueryListDropdown` component for users versus admins? ~~Also seems like a good time to ask: how are roles set up? Can users who are non admins be granted permisson to create collections.~~ never mind. I just looked at the roles for members in settings.

client/utils/collections/api.ts
idreyn: This is what I mean — and because it may be useful to let future callers of the function decide whether the tag should be public or not. You could argue that this is premature! But I do think it's best for functions to default to a less surprising/destructive action (in this case, creating a Collection that's initially public) and let callers opt-in to the more surprising behavior. So the function signature would be

```ts
({ title, communityId, isPublic = false }) => ...
```

idreyn:
> My bad, I assumed only admins have permission to add a pub to a collection and that it was taken care of with line 231

This is broadly but not totally true (and to be most technically correct, swap "manager" for "admin" there). If you are a Pub manager, you can also add a Pub to a Collection if one of these is true:

1. You are also a manager of the Collection
2. The Collection has `isRestricted = false` meaning that it accepts additions by anyone

This logic is encapsulated here: https://github.com/pubpub/pubpub/blob/13e71a9f750c7a695d3bc16a83ff3a2c723a63da/utils/collections/permissions.ts#L4

client/components/QueryListDropdown/QueryListDropdown.tsx
```diff
@@ -17,6 +17,14 @@ type Props<Item> = {
 	itemRenderer: ItemRenderer<Item>;
 	usePortal?: boolean;
 	query?: string;
+	newItemFromQuery?: (query: string) => Item;
```
idreyn, line 20: Since this is getting wordy, one trick I'd suggest using is to pull these definitions off the `QueryList` component:

```ts
type SharedQueryListProps = Pick<
	React.ComponentProps<typeof QueryList>,
	'createNewItemFromQuery' | 'createNewItemRenderer'
>;

type Props<Item> = SharedQueryListProps & { ... };
``` 

Breaking this down:

- The [TypeScript `typeof` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof) gives you the type that TS has inferred for a variable. This is, sadly, not to be confused with the [runtime `typeof` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof) which does something totally different. But the former can only appear in type expressions, and the latter only in runtime code.
- `React.ComponentProps<T>` tells you the props of a React component with type `T`.
- `Pick<T>` you've already used here!

This produces the original prop names from `QueryList` with the "create" prefix, which I think we should keep anyway. There are probably even more props in this definition that could be picked from the `QueryList` definition instead of manually written out!

client/components/PubCollectionsListing/PubCollectionsListing.tsx
```diff
@@ -196,19 +226,32 @@ const PubCollectionsListing = (props: Props) => {
 		);
 	};
 
+	const emptyListPlaceholderText = scopeData.activePermissions.canManageCommunity
+		? 'Create a new Tag with this name'
```
idreyn, line 230: As far as I can tell this text won't ever come up, because the `createNewItemRenderer` is always going to render a placeholder that prevents the list from being empty. Instead, I think you want to modify `renderNewItem` so that it tells users that clicking this item will create a new Collection:

```ts
renderCollectionTitle({ title: `Create new tag: ${query}`, ... })
```

client/components/PubCollectionsListing/PubCollectionsListing.tsx
```diff
@@ -196,19 +226,32 @@ const PubCollectionsListing = (props: Props) => {
...
+		: 'No Collections match this search.';
+
+	const searchPlaceholderText = scopeData.activePermissions.canManageCommunity
+		? 'Search for Collections (or create a Tag)'
+		: 'Search for Collections';
+
+	const createNewItemProps = scopeData.activePermissions.canManageCommunity
```
idreyn, line 237: One thing that occurs to me is that if there's no `emptyListPlaceholderText` for Community managers, it might make more sense to fold all three of these variable declarations into a single choice between two props objects:

```ts
const createNewItemProps = canManageCommunity
    ? { createNewItemFromQuery, createNewItemRenderer, searchPlaceholderText }
    : { emptyListPlaceholderText, searchPlaceholderText };
```

(omitting values for clarity)

client/utils/collections/api.ts
```diff
@@ -1,6 +1,18 @@
 import { apiFetch } from 'client/utils/apiFetch';
 
 const collectionPubsRoot = '/api/collectionPubs';
+const collection = '/api/collections';
+
+export const createTagCollection = ({ title, communityId, isPublic = true }) =>
```
idreyn, line 6: I'd set `isPublic = false` here (realizing my previous comment about this was a little hard to interpret)

idreyn (changes requested):
I tested this out and it's looking great for users with and without permissions to create new Tags. The only thing that needs attention is making sure tag-creating users see some text that lets them know they can create a tag (detailed below).

Some food for thought: when users click to create a new Tag, how do they know that something is happening? Is there somewhere we could put a spinner or similar?

idreyn:
(also, sorry for letting this languish for so long)

client/components/PubCollectionsListing/PubCollectionsListing.tsx
qweliant: ngl, i stared at this piece of code for seven minutes wondering if i was going crazy because a formatted string was being passed with title...Then i tried it...then it worked....thank you?

client/components/PubCollectionsListing/PubCollectionsListing.tsx
qweliant: there is an type error thrown requesting the emptyListPlaceholder prop if i remove it

client/components/PubCollectionsListing/PubCollectionsListing.tsx
```diff
@@ -226,27 +225,24 @@ const PubCollectionsListing = (props: Props) => {
...
-
 	const createNewItemProps = scopeData.activePermissions.canManageCommunity
-		? { createNewItemFromQuery: handleCreateTagFromQuery, createNewItemRenderer: renderNewItem }
-		: {};
-
+		? {
+				createNewItemFromQuery: handleCreateTagFromQuery,
+				createNewItemRenderer: renderNewItem,
+				searchPlaceholder: 'Search for Collections (or create a Tag)',
+				emptyListPlaceholder: '',
```
idreyn, line 233: (I am pretty sure this can be omitted)

idreyn (approved):
LGTM!

PR #1643, opened 2021-11-01
dev: Upgrade to Prettier 2.4.1

idreyn:
Doing a bit of upgrading while we spin up our Next Big Project. The huge changeset here was produced by running:
```
npm run lint -- --fix
```
to automatically apply the [new changes in Prettier's heuristics](https://prettier.io/blog/2020/03/21/2.0.0.html) to our source. Highlights here include better method chaining and parens around lone arrow function parameters (`(x) => x` instead of `x => x`).

There are two other Prettier-related packages in our `package.json`:

```
eslint-config-prettier
eslint-plugin-prettier
```

these have seen major version bumps since we installed them, but they don't have a changelog and seem to work with the new version so I'm leaving well enough alone.


3mcd (approved):
Awesome! LGTM.

PR #1644, opened 2021-11-01
enhancement: add AMA to citation styles

qweliant:
add AMA citations to the options citation style

### DIFF
--- a/utils/citations.ts
+++ b/utils/citations.ts
@@ -44,7 +44,7 @@ export const citationStyles: CitationStyle[] = [
 	{ key: 'elife', name: 'ELife', path: './citeStyles/elife.csl' },
 	{ key: 'frontiers', name: 'Frontiers', path: './citeStyles/frontiers.csl' },
 	{ key: 'mla', name: 'MLA', path: './citeStyles/modern-language-association.csl' },
-	{ key: 'apa', name: 'AMA', path: './citeStyles/american-medical-association.csl' },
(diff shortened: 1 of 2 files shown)


gabestein (approved):
LGTM! 

Tiny thing: we could probably just get away with "AMA." I don't think they release new additions very often, and the 11th is from 2020. We have APA 7th only because it is very recent and so some people are still using 6th.

utils/citations.ts
```diff
@@ -44,6 +44,7 @@ export const citationStyles: CitationStyle[] = [
 	{ key: 'elife', name: 'ELife', path: './citeStyles/elife.csl' },
 	{ key: 'frontiers', name: 'Frontiers', path: './citeStyles/frontiers.csl' },
 	{ key: 'mla', name: 'MLA', path: './citeStyles/modern-language-association.csl' },
+	{ key: 'apa', name: 'AMA 11th edition', path: './citeStyles/american-medical-association.csl' },
```
gabestein, line 47: should be `ama`, not `apa`.

gabestein (changes requested):
Wait I take it back, sorry. Tiny little language issue.

gabestein (changes requested):
Sorry, one more thing I added late to the other review: `CitationStyleKind` will need to be updated to include ama.

qweliant:
> Sorry, one more thing I added late to the other review: `CitationStyleKind` will need to be updated to include ama.

didnt see this before i re-requested, i will change it

gabestein (approved):
👍 for real this time! Thanks!

PR #1645, opened 2021-11-01
dev: Upgrade to TypeScript 4.4.4

idreyn:
Hot on the heels of #1643, this PR catches us up to the [state of the art in TypeScript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html). The code changes here are in reaction to changes in `tsc` and `eslint` behavior, and mostly fall under these categories:

- The default type of the variable bound in a `catch` clause is now `unknown` instead of `any`
- TypeScript now makes us explicitly import types from e.g. `firebase` and `React`
- Inference has improved and we can remove some `@ts-expect-error` comments

There are a few other TS-related packages I left alone here. `ts-jest` should be updated with `jest` and `ts-loader` with webpack. `ts-node` we could update but there's so much in the changelog that I'm honestly kind of scared to do so.

_Test plan:_
Passing CI gets us most of the way there, but I will test `PubSuspendWhileTyping`, `NewEdgeEditor`, and `useFocusTrap` to make sure I didn't inadvertently change their behavior.


client/containers/Pub/PubDocument/PubDiscussions/Discussion/Discussion.tsx
```diff
@@ -30,7 +30,7 @@ type Props = {
 	searchTerm?: string;
 };
 
-const sortThreadComments = (threadComments: ThreadComment[], sortType: SortType) => {
+const sortThreadComments = (threadComments: ThreadCommentType[], sortType: SortType) => {
```
3mcd, line 33: Thoughts on using a common prefix to never have type aliases clash with variable names? Like `T` or `$`? I'm not strongly opinionated, maybe something to discuss in a dev huddle.

3mcd (approved):
LGTM! Thank you so much for this!

client/containers/Pub/PubDocument/PubDiscussions/Discussion/Discussion.tsx
idreyn: This happens a ton on the server where TS definitions collide with model singletons. I've started to just:

```ts
import * as types from 'types`
```

and then refer to `types.CollectionPub`, reserving `CollectionPub` for the model itself. I didn't do that here because this `ThreadCommentType` is something weird and one-off.

I don't have strong opinions about prefixes, either. For readability I think `$` is better than `T`, but the meaning is less clear at a glance.

3mcd: Cool, I like the `import *` trick, so I'll start doing that too 👍

PR #1646, opened 2021-11-02
Bh activity digest emails wip

howebenhowe:
Addresses #1473 

## summary
Communities, especially their managers, would like to get email-delivered digests of developments within their communities. This PR uses the activity item system to meet that feature need. It fetches activity items for current community and renders them into email friendly HTML with inline styles, viewable at `[community base url]/email/digest`.

## target
Aiming at visual parity with [these mocks](https://www.figma.com/file/RUmAt42roXLhNobGwd9yfa/email-digest-draft?node-id=25%3A639)

## implementation
+ [styled-components](https://styled-components.com/) for inline css.
+ [juice](https://github.com/Automattic/juice#readme) to to inline css properties to the style attribute for maximum email friendliness
+ a lot of activityItem sorting/butchery

## still to complete
+ send these puppies out with mailgun
+ set up chron job to run weekly
+ subscription management toggle in dash
+ digest settings to manage type/volume of activity items in digest
+ email header/metadata
+ any kind of tests
+ currently, operates for current logged in user/community only -> wire up for arbitrary community

## attention reviewers!
+ the topmost component background/color/text is unreliable from community to community. I'm taking all suggestions for how to manage the different results visible between, for example, the email for the Contours community vs Fermentology vs Qualitative Criminology
+ how to capitalize the 'y' in `you created a Release of this pub` when you are forbidden from using css classes?
+ more reusable component breakdown for future email templating
+ CSS best practices, as I was merely shooting for visual match

## testing
Choose any three communities that have regular activity for the last week. At a minimum, check out the [demo community](demo.duqduq.org). For each of them, walk through the following:
+ log in to a community admin account
+ access each of their community activity dashboards
+ in parallel window, pull up the digest email preview by navigating to `[community base url]/email/digest`. If in dev. environment, modify the `PUBPUB_LOCAL_COMMUNITY` variable to, for example, `fermentology` then access `localhost:9876/email/digest` to see the rendered fermentology digest email. If on duqduq, you would go to fermentology.duqduq.org/email/digest
+ confirm that the all activity items from within the previous week are present in the digest
+ confirm that the activity items are grouped  by the affected object (pub/community/collection/page)
+ proof rendering of copy and design parity with the figma link above.

## questions
+ should we limit the number of ActivityItems returned? It's currently (arbitrarily) set to 150 items


server/utils/email/digest.ts
```diff
@@ -0,0 +1,81 @@
...
+					),
+				title: getAffectedObject(activityItemsGroupedByObjectId[objectId][0])?.title,
+				icon: getAffectedObjectIcon(activityItemsGroupedByObjectId[objectId][0]),
+			},
+		}),
+		{},
+	);
+
+	const communityItems: GroupedActivityItems = pick(dedupedActivityItems, communityId);
+	const pubItems: GroupedActivityItems = omit(dedupedActivityItems, communityId);
```
howebenhowe, line 74: This is the meat and potatoes of the affair, sorting and grouping activity items so that they can be displayed as: "Here's all the distinct kinds of things that have happened to pub *ABCD* in the last week!"

server/activityItem/fetch.ts
```diff
@@ -120,8 +121,13 @@ const applyFiltersToWhereQuery = (whereQuery: any, filters: ActivityFilter[]) =>
 const fetchActivityItemModels = async (
 	options: Required<FetchActivityItemsOptions>,
 ): Promise<types.ActivityItem[]> => {
-	const { scope, limit, offset } = options;
+	const { scope, limit, offset, since } = options;
 	const whereQuery = {
+		...(since && {
+			timestamp: {
+				[Op.gte]: since!,
```
idreyn, line 128: The `!` is not necessary here...TS will know that `since` is truthy on the right side of `since &&`

server/utils/email/digest.ts
```diff
@@ -0,0 +1,81 @@
+import pick from 'lodash.pick';
+import omit from 'lodash.omit';
+import { GroupedActivityItems } from 'client/components/Email';
+
+import { ActivityItem } from 'types';
+import { fetchActivityItems } from 'server/activityItem/fetch';
+
+const getAffectedObject = (item) =>
```
idreyn, line 8: `item` should be typed as `ActivityItem`

server/utils/email/digest.ts
```diff
@@ -0,0 +1,81 @@
...
+	item.pubId
+		? { id: item.pubId, title: item.payload.pub.title }
```
idreyn, line 10: ⚠️ These payload titles are sometimes going to be out of date...this is frequently true with Pubs, which start out as "Untitled Pub..." but are usually quickly renamed. So we need to look inside the `associations` from `fetchActivityItems()` to get the latest title:

```ts
const getAffectedObject = (item: ActivityItem, associations: ActivityAssociations) => {
	if (item.pubId) {
		const pub = associations.pub[item.pubId];
		return { id: item.pubId, title: pub.title };
	}
	// ...and so on
};
```

server/utils/email/digest.ts
```diff
@@ -0,0 +1,81 @@
...
+
+	const { activityItems, associations } = await fetchActivityItems({ scope, limit: 150, since });
+	const activityItemsGroupedByObjectId: Record<
+		string,
+		KeyedActivityItem[]
+	> = activityItems.reduce((result, item) => {
+		const objectId = getAffectedObject(item).id;
+		const payloadKeys = Object.keys(item.payload)
+			.sort()
+			.join();
```
idreyn, line 46: It might be helpful to `join('-')` or something, guarding on principle against two colliding strings formed from smooshing different sets of words together.

server/utils/email/digest.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export const getDigestData = async (initialData) => {
+	const {
+		communityData: { id: communityId },
+		scopeData: { scope },
+		loginData: { id: userId },
+	} = initialData;
+	const since = new Date(Date.now() - 7 * 24 * 60 * 60 * 1000).toISOString();
+
+	const { activityItems, associations } = await fetchActivityItems({ scope, limit: 150, since });
```
idreyn, line 38: Personally I think we should get rid of the limit altogether (I can see active communities blowing way past 150 items in a given week) and see if that gets us into any trouble.

server/utils/email/digest.ts
```diff
@@ -0,0 +1,81 @@
...
+	const activityItemsGroupedByObjectId: Record<
```
idreyn, line 39: I think this transformation from `activityItems` to `activityItemsGroupedByObjectId` could be its own function.

server/routes/email.tsx
```diff
@@ -0,0 +1,87 @@
...
+	try {
+		const initialData = await getInitialData(req, true);
+		const {
+			communityData: community,
+			loginData: { id: userId },
+		} = initialData;
+		const { templateSlug } = req.params;
+		if (
+			!hostIsValid(req, 'community') ||
+			process.env.NODE_ENV === 'production' ||
```
idreyn, line 69: `process.env.NODE_ENV === 'production'` corresponds to whether or not we're running a production _build_ of our code. To check whether we're in the production _environment_ use `isProd()` from `utils/environment`. This way we'll be able to see this route on our non-prod Heroku deployments!

client/components/Email/CommunityHeader.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+	return (
+		<Section
+			color={headerColor}
+			backgroundColor={backgroundColor}
+			backgroundImage={backgroundImage}
+			logo={logo}
+			alignment="left"
+		>
+			<H1Style>{props.title}</H1Style>
```
idreyn, line 52: Haven't cross-referenced the mockups but IMO we should include the Community title in here somewhere.

idreyn (changes requested):
I appreciate the amount of attention to detail here and overall it's looking great! There are a few issues I've marked with ⚠️ that need to be addressed to land, and most everything else is stylistic suggestions/curiosity.

One larger-scale thing: I do think this digest would benefit from a "Collection News" section, and I don't think it would be too hard to add within the framework you've come up with here.

idreyn:
> the topmost component background/color/text is unreliable from community to community. I'm taking all suggestions for how to manage the different results visible between, for example, the email for the Contours community vs Fermentology vs Qualitative Criminology

Could you elaborate here? Is it sometimes unreadable, or is it that you don't always reproduce the Community header style?

> how to capitalize the 'y' in `you created a Release of this pub` when you are forbidden from using css classes?

What happens if you add `text-transform: capitalize` to `MessageSpanStyle`?

One quick-and-dirty thing you could do is add a `capitalizeYou` property to `ActivityRenderContext` and then conditionally render `"You"` [here](https://github.com/pubpub/pubpub/blob/e62308698b9510b6c5e65d73d5f6e607f5cb433c/client/utils/activity/titles/user.ts#L13). That would cause problems if we ever wanted to use it anywhere besides the beginning of the sentence, though.

client/components/Email/digest/DigestIntro.tsx
```diff
@@ -0,0 +1,88 @@
...
+				<td>
+					<Section backgroundColor={fadedBackgroundColor}>
+						<BaseTableStyle>
+							<tr>
+								<TextCellStyle>
+									This digest is a compilation of activity in the&nbsp;
+									<a href={communityUrl(props.community)}>
+										{props.community.title}
+									</a>
+									&nbsp;community during the week of&nbsp;
```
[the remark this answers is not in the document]
3mcd: If you do need spaces between elements and text nodes you can use the old `{' '}` trick

client/components/Email/Button.tsx
```diff
@@ -0,0 +1,72 @@
...
+import React from 'react';
+
+import { BaseTableStyle } from './shared';
+
+const TableStyle = styled(BaseTableStyle).attrs({
+	align: 'center',
+})<StyleProps>`
+	background-color: ${(props) => props.backgroundColor};
+	text-align: center;
+	width: ${(props) => props.width}px !important;
```
3mcd, line 11: Just curious why we needed to use `!important` here?

3mcd (approved):
@idreyn did a thorough review and I don't see any other issues other than the small `!important` thing, so I'm gonna go ahead and approve. Great work!

client/components/Email/CommunityHeader.tsx
idreyn: Update: never mind

idreyn (approved):
Header image hell notwithstanding this looks good to land on my end! Great work.

howebenhowe:
@idreyn I think I'm all set to merge this into activity-digest-emails, but for your approval

Issue #1612, opened 2021-10-13
Build Submission model, queries, types, API


idreyn:
Submission model:
```
status: 'incomplete' | 'submitted' | 'accepted' | 'rejected'
submissionWorkflow: SubmissionWorkflow
```

- Update Pub model to have zero or on submissions
- Write `createSubmission()` and `updateSubmisisonStatus()` queries
- Write `canModerateSubmissions()` permission check which is just Collection managers
- Add `Submission` type to types directory
- Add `PUT` handler for `/api/submissions` to allow status change.
- Write simple API test

PR #1648, opened 2021-11-02
Submissions model, query, api, and tests

howebenhowe:
implements #1612 

Testing via API tests.
Largely boilerplate.

Question this raises: are we only ever interacting with `Submissions` directly when we are changing their statuses? For creation/deletion, that would only occur when the `Spub` is created/deleted, yes? So operations on `Submissions` then are typically side-effect like?

### DIFF
--- a/server/submission/api.ts
+++ b/server/submission/api.ts
@@ -7,10 +7,10 @@ import { createSubmission, updateSubmission, destroySubmission } from './queries
 const getRequestIds = (req) => {
 	const user = req.user || {};
 	return {
+		id: req.body.id || null,
 		userId: user.id,
 		communityId: req.body.communityId,
 		collectionId: req.body.collectionId,
-		submissionId: req.body.submissionId || null,
 	};
 };
 
@@ -49,6 +49,6 @@ app.delete(
 			throw new ForbiddenError();
 		}
 		await destroySubmission(req.body);
-		return res.status(200).json(req.body.submissionId);
+		return res.status(200).json(req.body.id);
 	}),
 );
--- a/server/submission/api.ts
+++ b/server/submission/api.ts
@@ -1,0 +1,54 @@
+import app, { wrap } from 'server/server';
+import { ForbiddenError } from 'server/utils/errors';
+
+import { getPermissions } from './permissions';
+import { createSubmission, updateSubmission, destroySubmission } from './queries';
+
+const getRequestIds = (req) => {
+	const user = req.user || {};
+	return {
+		userId: user.id,
+		communityId: req.body.communityId,
+		collectionId: req.body.collectionId,
+		submissionId: req.body.submissionId || null,
+	};
+};
+
+app.post(
+	'/api/submissions',
+	wrap(async (req, res) => {
+		const ids = getRequestIds(req);
+		const permissions = await getPermissions(ids);
+		if (!permissions.create) {
+			throw new ForbiddenError();
+		}
+		const newSubmission = await createSubmission(req.body, req.user.id);
+		return res.status(201).json(newSubmission);
+	}),
+);
+
+app.put(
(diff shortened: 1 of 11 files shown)


server/submission/queries.ts
```diff
@@ -0,0 +1,34 @@
...
+export const createSubmission = async (inputValues, actorId = null) =>
+	Submission.create(
+		{
+			pubId: inputValues.pubId,
+			status: inputValues.status,
+		},
+		{ actorId },
+	);
+
+export const updateSubmission = async (inputValues, updatePermissions, actorId = null) => {
```
idreyn, line 12: This function (and in general, all new functions) should have type annotations on its arguments:

```ts
(inputValues: Partial<Submission>, updatePermissions: keyof Submission, actorId: null | string = null) => ...
```

I would consider renaming `inputValues` to `patch` or `update` and `updatePermissions` to `allowedFields`. I know these names are in heavy rotation in older code, but they don't really make sense — "permissions" in particular points to a totally different concept in our codebase.

Also, since there's only ever going to be one updatable field for the forseeable future, I think it might be simpler to have an `updateSubmissionStatus` function instead of this one.

server/submission/queries.ts
```diff
@@ -0,0 +1,34 @@
...
+	const filteredValues: Record<string, any> = {};
```
idreyn, line 13: This whole thing could be replaced by `lodash.pick`!

server/submission/queries.ts
```diff
@@ -0,0 +1,34 @@
...
+		}
+	});
+	return Submission.update(filteredValues, {
+		where: { id: inputValues.submissionId },
+		actorId,
+		individualHooks: true,
+	}).then(() => filteredValues);
+};
+
+export const destroySubmission = (inputValues, actorId = null) =>
```
idreyn, line 26: We won't call this anywhere in our currently-planned work (but there's not really any harm in keeping it around)

idreyn (commented):
Dropping a few comments your way — I know you're still working on the permissions stuff

server/submission/api.ts
```diff
@@ -0,0 +1,54 @@
...
+
+app.post(
+	'/api/submissions',
+	wrap(async (req, res) => {
+		const ids = getRequestIds(req);
+		const permissions = await getPermissions(ids);
+		if (!permissions.create) {
+			throw new ForbiddenError();
+		}
+		const newSubmission = await createSubmission(req.body);
```
idreyn, line 25: (Don't forget to pass the user ID into the `actorId` here)

howebenhowe: I actually took the opposite approach, and pulled all the activityItem stuff out, so that it can be added in its own PR. Seemed prudent.

server/submission/queries.ts
howebenhowe: This was copied over as boilerplate, seems pretty much unnecessary with a model that has basically one editable field.

server/submission/queries.ts
howebenhowe: Amen. I am inclined to do this as a homework assignment across the board, but it's not pressing. A little back end semantic meaning tune-up could go a long way for the Qwelian's and Ben's of the world, just starting to get familiar with the back end.

server/submission/queries.ts
```diff
@@ -0,0 +1,34 @@
+import { Submission } from 'server/models';
+
+export const createSubmission = async (inputValues, actorId = null) =>
```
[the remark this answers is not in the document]
howebenhowe: with it, except right now we're working with `incomplete` as the initial status

idreyn (commented):
LGTM. 🥇 first merge onto the `submissions` branch!

idreyn (approved):
whoops there we go

Issue #1617, opened 2021-10-13
Build SubmissionWorkflow model, queries, API


idreyn:
Model fields:

```
enabled: boolean
instructionsText: JSON
emailText: JSON
layoutBlockContent: JSON
targetEmailAddress: string
```

- Update Collection model to have optional `SubmissionWorkflow`
- Queries: `createSubmissionWorkflow()`, `updateSubmissionWorkflow()`, `destroySubmissionWorkFlow()`
- Permissions: `canManageSubmissionWorkflow()` true for managers of associated Collection
- API: `POST`, `PUT`, `DELETE` for `/api/submissionWorkflows`
- Simple API test
- new `SubmissionWorkflow` type in types directory

PR #1650, opened 2021-11-03
feature: Submission Workflow API, Model, Queries, and Tests

qweliant:
addresses issue #1617 

The model, query, api, and tests for submissions workflow is now available. 

Test plan:

run test suite using `npm run test-dev server/submissionWorkflow`


qweliant:
im not sure what type to make the JSON fields. currently they are any

idreyn:
> im not sure what type to make the JSON fields. currently they are any

Since they're going to hold ProseMirror document content, let's use the `DocJson` type. I have another version of the `SubmissionWorkflow` type that you can grab from #1658!

server/submissionWorkflow/queries.ts
```diff
@@ -0,0 +1,42 @@
...
+			enabled,
+			instructions,
+			afterSubmittedText,
+			email,
+			layoutBlock,
+		}),
+	);
+};
+
+export const updateSubmissionWorkflow = async (inputValues, updatePermissions) => {
```
idreyn, line 24: 5️⃣ There are a few things that need some attention here in `updateSubmissionWorkflow`:

As I mentioned over in `permissions.ts`, you can simplify this logic by noting that the `updatePermissions` is constant for all users that have _any_ permissions to call this function, and remove them entirely. Another benefit of this is that it decouples queries from any notion of permissions, so they can be easily called elsewhere on the server, outside of permissions-aware API routes.

Now we want a type for the lone `inputValues` argument. I'd rename this argument to `update`, and give it the type  `Partial<Omit<types.SubmissionWorkflow, 'id' | 'createdAt' | 'updatedAt'>>` — and probably factor the `Omit<...>` into a type that's shared with the create query above.

This puts some bounds on what TypeScript will allow us to pass in here, but that does nothing to prevent runtime code from sneaking extra values into the SQL query. To do that, let's unpack all the variables we actually want to allow folks to change:

```ts
const { collectionId, enabled, instructionsText, emailText, ... } = update;
await SubmissionWorkflow.update({ enabled, instructionsText, ... }, { where: { collectionId } });
```

It's good practice when writing Sequelize queries to pass in an object literal that explicitly lays out the props that are being updated. When we write `Model.update(objectFromElsewhere)` I'm always a little suspicious that `objectFromElsewhere` might have some nefarious, unanticipated props — even when it's defined very nearby.

server/submissionWorkflow/api.ts
```diff
@@ -0,0 +1,66 @@
...
+	createSubmissionWorkflow,
+	updateSubmissionWorkflow,
+	destroySubmissionWorkFlow,
+} from './queries';
+
+const getRequestIds = (req) => {
+	const user = req.user || {};
+	return {
+		userId: user.id,
+		communityId: req.body.communityId,
```
idreyn, line 15: 3️⃣ You should be able to eliminate `communityId` from all of your API code — providing a `collectionId` should be enough to uniquely identify a Collection.

server/submissionWorkflow/api.ts
```diff
@@ -0,0 +1,66 @@
...
+	wrap(async (req, res) => {
+		const requestIds = getRequestIds(req);
+		const permissions = await getPermissions(requestIds);
+		if (!permissions.update) {
+			throw new ForbiddenError();
+		}
+		const updatedValues = await updateSubmissionWorkflow(
+			{
+				...req.body,
+				submissionWorkflowId: req.body.id,
```
idreyn, line 44: 4️⃣ Consider that this value was never checked by `getPermissions`. What would happen in an attacker provided the ID of a `Collection` they own along with the ID of a `SubmissionWorkflow` they don't own? Your code should protect against this — I'd suggest by using `collectionId` instead of `id` in your `SubmissionWorkflow.update()` query. This is a good case to consider when writing API unit tests, too.

server/submissionWorkflow/queries.ts
```diff
@@ -0,0 +1,42 @@
...
+			filteredValues[key] = inputValues[key];
+		}
+	});
+	await SubmissionWorkflow.update(filteredValues, {
+		where: { id: inputValues.submissionWorkflowId },
+	});
+	return filteredValues;
+};
+
+export const destroySubmissionWorkFlow = (inputValues) => {
```
idreyn, line 38: This could probably just take a single `id: string`.

types/submissonWorkflow.ts
```diff
@@ -0,0 +1,15 @@
+import { DocJson } from 'types';
+import { LayoutBlockSubmissionBanner } from 'utils/layout';
+
+export type EnabledStatus = true | false;
```
idreyn, line 4: Don't think this gets used anywhere.

server/submissionWorkflow/queries.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const createSubmissionWorkflow = async ({
+	collectionId,
+	enabled,
+	instructions,
+	afterSubmittedText,
+	email,
+	layoutBlock,
+}) => {
+	return Collection.findOne({ where: { id: collectionId } }).then(() =>
```
idreyn, line 12: Good thinking here, but be aware that this promise will resolve whether or not a Collection was actually found.

idreyn (changes requested):
Thanks for getting this up quickly — there are a lot of elements to consider here. There are some patterns I see here that have been adapted from older PubPub code, including some that have caused us trouble that we've moved away from; I should apologize for not calling those out earlier, and I hope my feedback here will communicate the "why" as much as the "what" on the requested changes.

I'm trying a new thing here using numbered emojis to call out major code areas that need attention in the order I think it makes sense to address them. If you hit 1️⃣ through 5️⃣ and then go back and address remaining comments, you should be in good shape! I'd also appreciate knowing whether this ordering turned out to be helpful or not :)

server/submissionWorkflow/model.ts
```diff
@@ -0,0 +1,27 @@
+export default (sequelize, dataTypes) => {
+	return sequelize.define(
+		'SubmissionWorkflow',
+		{
+			id: sequelize.idType,
+			enabled: { type: dataTypes.BOOLEAN, allowNull: false },
+			instructions: { type: dataTypes.JSONB },
```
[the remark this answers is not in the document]
qweliant: sorry, for some reason i thought your comment in slack meant combine the email fields.

server/submissionWorkflow/model.ts
```diff
@@ -0,0 +1,27 @@
...
+			layoutBlock: {
+				type: dataTypes.JSONB,
+			},
+			targetEmailAddress: { type: dataTypes.STRING },
+		},
+		{
+			classMethods: {
+				associate: (models) => {
+					const { Collection, SubmissionWorkflow } = models;
+					SubmissionWorkflow.belongsTo(Collection, {
```
idreyn, line 18: So I know these Sequelize association method names are actively misleading. One tip for using them is that, when using `belongsTo`, all of these things need to match up:

- The model provided as the first argument
- The name of the `as` property
- The foreign key

Here you have a mismatch, that results in a `submissionWorkflow.submissionWorkflow` property which is actually a Collection. If you change `as` to `'collection'` you'll get `submissionWorkflow.collection`, which is a useful thing to have. To make this work you'll also need to add a (nullable) `collectionId` property to this model, which I wanna say you previously had?

Also, let's drop the `onDelete: 'CASCADE'`. If a Collection is deleted, I think we want to err on the side of keeping its submission workflow around in the database.

server/submissionWorkflow/__tests__/api.test.ts
```diff
@@ -0,0 +1,184 @@
...
+		})
+		.expect(200);
+
+	const updatedSubmission = await SubmissionWorkflow.findOne({
+		where: { collectionId: collection.id },
+	});
+	expect(updatedSubmission).toMatchObject({ enabled });
+});
+
+it('forbids a non-Community manager or attacker from updating a workflow', async () => {
```
idreyn, line 134: Sweet

server/submissionWorkflow/api.ts
idreyn: Please do give this a shot (and let me know if it's not working, that's indicative of a problem with `getScope` that we need to fix)

server/submissionWorkflow/api.ts
```diff
@@ -0,0 +1,57 @@
...
+	updateSubmissionWorkflow,
+	destroySubmissionWorkFlow,
+} from './queries';
+
+const getRequestIds = (req) => {
+	const user = req.user || {};
+	return {
+		userId: user.id,
+		communityId: req.body.communityId,
+		collectionId: req.body.id || null,
```
idreyn, line 16: You're passing in `req.body.collectionId` from your tests — I think that's what you want to be using here.

idreyn (changes requested):
Few more things to take care of.

qweliant:
Should i check if `updateSubmissionWorkflow` actually finds a row in the DB? I have not noticed that and wanted to know why we don't throw an error if there isn't an existing row in the DB.

```export const updateSubmissionWorkflow = async (update: Update) => {
	const {
		collectionId,
		enabled,
		instructionsText,
		emailText,
		layoutBlockContent,
		targetEmailAddress,
	} = update;
	await SubmissionWorkflow.update(
		{ enabled, instructionsText, emailText, targetEmailAddress, layoutBlockContent },
		{ where: { collectionId } },
	);
};```

File: server/submissionWorkflow/model.ts [hunk 11]
```diff
@@ -0,0 +1,27 @@
+export default (sequelize, dataTypes) => {
+	return sequelize.define(
+		'SubmissionWorkflow',
+		{
+			id: sequelize.idType,
+			collectionId: { type: dataTypes.UUID, allowNull: false },
+			enabled: { type: dataTypes.BOOLEAN },
```
idreyn, line 7: Ah wait before you land please do make these `allowNull: false`

server/submissionWorkflow/model.ts
```diff
@@ -0,0 +1,27 @@
...
+				type: dataTypes.JSONB,
+			},
+			targetEmailAddress: { type: dataTypes.STRING },
+		},
+		{
+			classMethods: {
+				associate: (models) => {
+					const { Collection, SubmissionWorkflow } = models;
+					SubmissionWorkflow.belongsTo(Collection, {
+						as: 'Collection',
```
idreyn, line 20: We'll want this to be `'collection'` so the property is `someSubmissionWorkflow.collection` instead of `someSubmissionWorkflow.Collection`.

idreyn (approved):
Looking fantastic, thanks for following up with all that. There are just a few more things to look at in the `model.ts` but I don't want to block you from landing.

idreyn:
> Should i check if `updateSubmissionWorkflow` actually finds a row in the DB?

You're welcome to do this (it's probably good practice, but missing from a lot of our code). The pattern would be:

```ts
const model = await Model.findOne(...);
model.update(update);
await model.save();
```

PR #1652, opened 2021-11-07
Bh/submission activity item

howebenhowe:
Implements `submission-created`, `submission-deleted`, and `submission-status-changed` activity items.

TODO: include submission activity items in the activity feed, but combine submission creation and pub creation as they are the same thing from the user's perspective.


howebenhowe:
@idreyn Mr. `ActivityItem`, how does a person work on the renderers and title functions for an activity item when we don't have any `submission` activity in the DB yet? Got anything slicker and more reflective of user generated data than one-off insert statements into the DB?

stubstub/activity.ts
```diff
@@ -29,16 +28,26 @@ export const expectCreatedActivityItem = <Result extends Record<string, any>>(
...
+		toMatchObject: async (values) => {
+			const result = await promise;
+			await finishDeferredTasks();
+			const { query, assertion } = splitQueryFromAssertion(values);
+			const item = await ActivityItem.findOne({
+				where: { ...query, createdAt: { [Op.gte]: cutoffTime } },
+			});
+			expect(item).toMatchObject(assertion);
+			return result;
+		},
```
3mcd, line 51: Was splitting this function in two a matter of preference (e.g. to keep `toMatchObject` closer to Jest's version), or was it necessary for some other reason?

client/containers/DashboardOverview/helpers/QuickActions.tsx
```diff
@@ -6,7 +6,7 @@ require('./quickActions.scss');
 
 export type QuickAction = {
 	icon: IconName;
-	label: React.ReactNode;
+	label: string;
```
3mcd, line 9: Super minor (feel free to ignore): at a quick glance it looks like we use string-only values for quick action labels. But I wonder if we could preserve accepting an arbitrary React.Node and instead use the `href` as the key?

3mcd (approved):
LGTM! But I'd still implore @idreyn to take a quick glance since he's 👑 of activity items.

PR #1658, opened 2021-11-09
feature: Submissions tab

### LINKED ISSUE #1613: Add Submissions tab for Collection dashboard
- Add a new route in `server/routes` (plenty of boilerplate there to copy)
- Add new tab to Dashboard side menu
- Check a feature flag before showing tab (can make check without flag existing in db)
- Place to put following pieces of UI on one dashboard tab:
    - Submission instructions etc (#1614)
    - List of submitted Pubs (#1615)

### LINKED ISSUE #1614: Add SubmissionWorkflow configuration to Submissions tab
- Disabled if `collection.pageId !== null` because then we can't render the Collection layout
- Submissions enabled/disabled switch
- Instructions, completion text, email text as ProseMirror editors (ask Ian which one)
- Layout block details and button text editors
- Submission email address (use community email as fallback?)

idreyn:
This PR adds a Submissions tab to the Collection scope of the dashboard, hidden behind the `submissions` feature flag. It also has a first pass at a `SubmissionWorkflowEditor`. Follow-on work for this PR will include:

- Querying and sending real `SubmissionWorkflow` model from #1617
- Creating a new `SubmissionWorkflowEditor` wrapper that persists a real model via the API
- Switching between viewing submitted Pubs and editing an existing workflow

Because there is a lot of integration work left to do, there are a lot of dead ends (and odds and ends) in this PR that I will flag with my own review comments.

_Test plan:_
- Enroll yourself in the `submissions` feature flag (or just log in as PubPub Team)
- Navigate to a Collection in the dashboard and verify that you can see the Submissions tab
- Make sure you can use all the workflow editor controls, and that your WIP persists via `localStorage` when refreshing
- Click "Save and continue" and verify that you are taken to a mostly-blank dashboard

_Screenshot:_
image

Resolves #1613
Resolves #1614


client/components/ClientOnly/ClientOnly.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
idreyn, line 1: Rationale for this is that using `localStorage` in a feature prevents us from server-side rendering it. The `useLocalStorage` hook checks that the context rendered here is present, which makes it impossible to accidentally use browser-specific features during SSR.

client/components/Layout/LayoutSubmissionBannerSkeleton.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
idreyn, line 1: Eventually, when we figure out what a `LayoutSubmissionBanner` looks like, it will delegate its look 'n' feel to this component. Bit of a dead end on the `onSubmitClicked` handler for now.

client/components/MinimalEditor/minimalEditor.scss
```diff
@@ -8,10 +8,16 @@ $border-style: 1px solid #ddd;
         border-radius: 2px;
         padding: 0.5em;
         cursor: text;
+        > .editor {
```
idreyn, line 11: This has been driving me crazy since day one

client/components/SubmissionEmail/SubmissionEmail.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type Props = {
+	community: Community;
+	customText: React.ReactNode;
+	submissionTitle: string;
+	submissionUrl?: string;
+	submitterName: React.ReactNode;
+};
+
+const SubmissionEmail = (props: Props) => {
```
idreyn, line 14: I imagine we'll eventually use this component to render the text that gets emailed to submitters (and it'll be kept in sync with the preview on the dashboard)

client/containers/DashboardSubmissions/DashboardSubmissions.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import { ClientOnly, DashboardFrame } from 'components';
+
+import NewSubmissionWorkflowEditor from './NewSubmissionWorkflowEditor';
+import { EditableSubmissionWorkflow } from './types';
+
+require('./dashboardSubmissions.scss');
+
+const DashboardSubmissions = () => {
+	const [workflow, setWorkflow] = useState<null | EditableSubmissionWorkflow>(null);
```
idreyn, line 12: Eventually we'll use `usePersistableState` here

client/containers/DashboardSubmissions/NewSubmissionWorkflowEditor.tsx
```diff
@@ -0,0 +1,76 @@
...
+		enabled: false,
+		bannerContent: {
+			title: '',
+			body: getEmptyDoc(),
+		},
+	};
+};
+
+type Props = {
+	onWorkflowCreated: (w: EditableSubmissionWorkflow) => unknown;
```
idreyn, line 26: Eventually this component will be responsible for POSTing `/api/submissionWorkflows`, and it'll call this with the whole model (ID and all)

client/containers/DashboardSubmissions/StartWorkflowCallout.tsx
```diff
@@ -0,0 +1,61 @@
...
+			{url}
+		</a>
+	));
+
+	const pageLinkedToCollection = communityData.pages.find(
+		(page) => page.id === activeCollection.pageId,
+	);
+
+	return (
+		<Callout
```
idreyn, line 38: This could have so much more info in it...TK

client/utils/featureFlags.ts
```diff
@@ -1,6 +0,0 @@
-import { getClientInitialData } from './initialData';
```
idreyn, line 1: So it turns out that this doesn't work at all because `getClientInitialData()` doesn't work during the SSR stage. I think there's a more careful version of this that we could write, but for now we'll just get feature flags from `usePageContext()` — I updated the docs to reflect this.

client/utils/localStorage.ts
```diff
@@ -0,0 +1,48 @@
+export type LocalStorable = Record<string, any>;
+
+export type LocalStorageWriter<T extends LocalStorable> = {
+	setItem: (item: T) => void;
+	getItem: () => T;
+	getChild: <U extends LocalStorable>(key: string) => LocalStorageWriter<U>;
+};
+
+// See https://en.wikipedia.org/wiki/Private_Use_Areas
+const separator = '\uE000';
```
3mcd, line 10: TIL

client/utils/localStorage.ts
```diff
@@ -0,0 +1,48 @@
...
+	}
+};
+
+const createLocalStorageWriter = <T extends LocalStorable>(key: string): LocalStorageWriter<T> => {
+	const setItem = (item: T) => {
+		localStorage.setItem(key, JSON.stringify(item));
+	};
+
+	const getItem = (): T => {
+		return JSON.parse(localStorage.getItem(key) ?? 'null');
```
3mcd, line 30: I've never thought about doing it this way (in one line without a if-statement/ternary) before!

3mcd (approved):
LGTM!

client/containers/App/SideMenu/SideMenu.tsx
```diff
@@ -82,7 +102,7 @@ const SideMenu = () => {
 		},
 		{
 			title: 'Members',
-			icon: 'people' as const,
```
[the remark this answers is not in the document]
idreyn: AFAIK the `as const` is needed because these are mutable object properties. Since I added the `MenuItem` type, TS narrows these to `IconName`.

Issue #1620, opened 2021-10-13
Update getManyPubs to understand Submissions


idreyn:
Query on:
- Presence/absence of submission
- Submission state

PR #1665, opened 2021-11-16
enhancement: query many submissions

qweliant:
issue number #1620

Query pubs based the presence of a submission status

### DIFF
--- a/server/pub/queryMany.ts
+++ b/server/pub/queryMany.ts
@@ -50,7 +50,7 @@ const createInnerWhereClause = (query: PubsQuery) => {
 };
 
 const createJoins = (query: PubsQuery) => {
-	const { scopedCollectionId, term, hasReviews } = query;
+	const { scopedCollectionId, term, hasReviews, submissionStatus } = query;
 	return (builder: QueryBuilder) => {
 		if (scopedCollectionId) {
 			builder.innerJoin(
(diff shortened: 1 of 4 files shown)


server/pub/queryMany.ts
```diff
@@ -67,6 +67,12 @@ const createJoins = (query: PubsQuery) => {
 		if (typeof hasReviews === 'boolean') {
 			builder.leftOuterJoin('ReviewNews', 'Pubs.id', 'ReviewNews.pubId');
 		}
+		if (submissionStatus) {
+			builder.innerJoin('Submissions', {
+				'Submissions.pubId': 'Pubs.id',
+				'Submissions.status': knex.raw('some(?::string[])', [submissionStatus]),
```
qweliant, line 73: should knex.raw be builder.where

idreyn: I don't think so. The object `{ 'Submissions.status': X }` is acting as the `ON` clause of a `JOIN` query (kind of like a `WHERE`)  and this one key-value pair renders down to `"Submissions"."status" = X`. The `X` here is the query builder carefully injecting our statuses array into the query and casting it to an array of strings with `::string[]`. The `some()` is checking to see if that equality is true of _any_ element in the array (it is a weird syntax).

idreyn (commented):
This looks good to me. Does it work?

PR #1672, opened 2021-11-17
Integrate Submission and SubmissionWorkflow models

idreyn:
Since these were done in parallel, they need a little stitching together. Specifically, the Submission model needs a `submissionWorkflowId`, and its API needs to check for the presence of an enabled workflow before allowing users to start a submission.

_Test plan:_
```
npm run test-dev server/submission
```

### DIFF
--- a/server/submission/queries.ts
+++ b/server/submission/queries.ts
@@ -44,7 +44,7 @@ export const updateSubmission = async (patch: UpdateOptions, sendEmail = false) 
 	await submission.save();
 	if (sendEmail) {
 		if (previousStatus === 'incomplete' && status === 'submitted') {
-			// Send a freaking email
+			// Send an email from here
 		}
 	}
 
--- a/server/submission/queries.ts
+++ b/server/submission/queries.ts
@@ -1,25 +1,61 @@
-import { Submission } from 'server/models';
-import { Submission as SubmissionType, managerStatuses, submitterStatuses } from 'types';
+import { createPub } from 'server/pub/queries';
+import { Collection, Submission, SubmissionWorkflow } from 'server/models';
+import {
+	Submission as SubmissionType,
+	SequelizeModel,
+	managerStatuses,
+	submitterStatuses,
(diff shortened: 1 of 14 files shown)


server/submission/queries.ts
```diff
@@ -1,25 +1,61 @@
...
+
+type UpdateToStatus = typeof updateToStatuses;
+type UpdateOptions = Partial<SubmissionType> & { status: UpdateToStatus };
+
+export const updateSubmission = async (patch: UpdateOptions, sendEmail = false) => {
+	const { id, status } = patch;
+	const submission: SequelizeModel<SubmissionType> = await Submission.findOne({ where: { id } });
+	const previousStatus = submission.status;
+	submission.status = status;
+	await submission.save();
```
qweliant, line 44: What is effected by await submission.save()?

server/submission/queries.ts
```diff
@@ -1,25 +1,61 @@
...
+	if (sendEmail) {
+		if (previousStatus === 'incomplete' && status === 'submitted') {
+			// Send an email from here
+		}
+	}
 
-export const updateSubmission = async (patch: PatchType) =>
 	Submission.update(
 		{ status: patch.status },
```
qweliant, line 52: why not just pass status here?

qweliant, line 52: why not just pass status here?

idreyn: I could see an `updateSubmissionStatus` function being an appropriate way to handle this, but we need to pass in `{id, status}` anyway so kinda it might as well be a single object. But I didn't write this function :)

server/submission/queries.ts
idreyn: It's making the same kind of SQL query as `Submission.update()` — when you assign new values to the properties on a Sequelize model object, it keeps track of them in a "dirty" object that contains the values that haven't made it to the server yet. Calling `.save()` makes the query to persist those.

PR #1673, opened 2021-11-17
feature(?): Top secret Release Discussions dialog

### LINKED ISSUE #1668: Write script to make draft discussions public

idreyn:
Resolves #1668

In response to a recurring need from the content team to manually move discussions from draft to release, here's a little widget that does this. For now it's only for the `pubpub-team` user, but we could make it part of the release process in the future. To use it, click the little arrow-out icon in the Discussions section:

image

Then choose which ones you want to move:

(Don't know why that vertical spacing is goofy but I'm out of time to fix it.)

_Test plan:_
- Open a Pub draft, add some discussions, and then make a Release. Return to the draft and use this newfangled button to move the discussions to the release
- Make sure this button does not appear if you are not logged in as the PubPub team, if there are no discussions, or if you're viewing the draft.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -52,6 +52,7 @@
 		"import/no-unresolved": [2, { "ignore": ["\\config.js$"] }],
 		"import/prefer-default-export": 0,
 		"jsx-a11y/click-events-have-key-events": 0,
+		"jsx-a11y/accessible-emoji": 0,
 		"jsx-a11y/label-has-associated-control": [2, {"controlComponents": ["Switch"]}],
 		"lines-between-class-members": ["error", "always", { "exceptAfterSingleLine": true }],
 		"max-classes-per-file": 0,
--- a/client/containers/Pub/PubDocument/PubBottom/Discussions/DiscussionsReleaseDialog.tsx
+++ b/client/containers/Pub/PubDocument/PubBottom/Discussions/DiscussionsReleaseDialog.tsx
@@ -1,0 +1,97 @@
+import React, { useCallback, useState } from 'react';
+import { Button, Callout, Checkbox, Classes, Dialog, IDialogProps, Label } from '@blueprintjs/core';
+
+import { Community, Pub, PubPageData, PubPageDiscussion } from 'types';
+import { apiFetch } from 'client/utils/apiFetch';
+import { usePageContext } from 'utils/hooks';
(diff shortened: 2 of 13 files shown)


.eslintrc
```diff
@@ -52,6 +52,7 @@
 		"import/no-unresolved": [2, { "ignore": ["\\config.js$"] }],
 		"import/prefer-default-export": 0,
 		"jsx-a11y/click-events-have-key-events": 0,
+		"jsx-a11y/accessible-emoji": 0,
```
idreyn, line 55: This is deprecated

gabestein (approved):
Awesome! A little thanksgiving treat. Works great in a little test. 

One note that I think is not blocking, but should make us think how we eventually make it public (which I think we should): you can send a discussion over the line on a part of the document that hasn't been released, which causes it to temporarily dematerialize. Amazingly, it rematerializes on the release when you then push the release (assuming you haven't made any changes to the draft).

To avoid situations like that, in an eventual public version of this feature, I might suggest explicitly tying it to the release process rather than making it separate. That is, all the discussions that could be moved show up when you create a release, and you can select the ones you want (or choose all) to be moved.

client/containers/Pub/PubDocument/PubBottom/Discussions/DiscussionsReleaseDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+			className="discussions-release-dialog-component"
+			onClose={onClose}
+			isOpen={isOpen}
+			title="Release Discussions"
+		>
+			<div className={Classes.DIALOG_BODY}>
+				<p>
+					Select discussions to move from the draft to the latest Release. Once they're
+					moved, you won't be able to undo this action (though you may be able to ask Ian
+					to do it for you 😊)
```
3mcd, line 79: Blocker: replace "Ian" with `developerWithSpareTime()` :joy:

(jk of course)

3mcd (approved):
LGTM!

client/containers/Pub/PubDocument/PubBottom/Discussions/DiscussionsReleaseDialog.tsx
idreyn: `Ian` was chosen by fair dice roll and is guaranteed to be random.

PR #1675, opened 2021-11-18
Remove Bluebird

3mcd:
This PR removes our `bluebird` dependency. This should shave off a few hundred KB from our Heroku slug size. This necessitated that we port few Bluebird utilities, specifically `Promise.each`, `Promise.map` and `Promise.mapSeries`. The helpers live in a new file at utils/async.ts.

In addition, Bluebird's `promisify` was replaced with Node's `util.promisify`. As far as I can tell, the two functions work exactly the same way. 

There may be some minor risk in these changes, but I've done my best to ensure that the each/map/mapSeries ports work 1:1 with unit tests.

_Test Plan_

I think a quick smoke test should suffice. @idreyn could you maybe help me find the best code to test `Promise.map`/`asyncMap`'s concurrency option?

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
 		"amqplib": "^0.5.3",
 		"aws-sdk": "^2.444.0",
 		"big-json": "^2.0.2",
-		"bluebird": "^3.5.3",
 		"body-parser": "^1.19.0",
 		"bourbon": "^5.1.0",
 		"camelcase-css": "^2.0.1",
--- a/server/activityItem/__tests__/fetch.test.ts
+++ b/server/activityItem/__tests__/fetch.test.ts
@@ -1,4 +1,3 @@
-/* global describe, it, expect, beforeEach, beforeAll, afterAll */
 import { modelize, setup, teardown } from 'stubstub';
 
 import { ActivityAssociations, ActivityAssociationType } from 'types';
--- a/utils/async.ts
+++ b/utils/async.ts
@@ -64,12 +64,12 @@ export async function asyncMap<T, R>(
 					cursor++;
 					pending++;
 					Promise.resolve(next)
-						.then(function onVisitedPromiseResolve(value) {
+						.then((value) => {
 							return iteratee(value, index, resolvedLength);
 						})
 						.then(
 							// eslint-disable-next-line no-loop-func
-							function onVisitedPromiseUnwrapped(value) {
+							(value) => {
 								pending--;
 								results[index] = value;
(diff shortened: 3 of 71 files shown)


utils/async.ts
```diff
@@ -0,0 +1,98 @@
+import { assert } from 'utils/assert';
```
3mcd, line 1: Ported Bluebird utils are here.

package.json
```diff
@@ -47,7 +47,6 @@
 		"amqplib": "^0.5.3",
 		"aws-sdk": "^2.444.0",
 		"big-json": "^2.0.2",
-		"bluebird": "^3.5.3",
```
3mcd, line 50: RIP

server/activityItem/__tests__/fetch.test.ts
```diff
@@ -1,4 +1,3 @@
-/* global describe, it, expect, beforeEach, beforeAll, afterAll */
```
3mcd, line 1: I also made a small tweak to our eslint config so we don't have to define jest globals in our test files anymore.

idreyn:
> @idreyn could you maybe help me find the best code to test Promise.map/asyncMap's concurrency option?

All of the callsites appear to either be in tools that don't get run regularly or in codepaths that are a little hard to trigger. I think the easiest one would be to create a Pub that is contained in many Collections, and then check that when you update its `ScopeSummary` (e.g. by adding a Discussion to it) that it updates all of the Collections.

utils/async.ts
```diff
@@ -0,0 +1,98 @@
...
+					const next = resolvedList[index];
+					cursor++;
+					pending++;
+					Promise.resolve(next)
+						.then(function onVisitedPromiseResolve(value) {
+							return iteratee(value, index, resolvedLength);
+						})
+						.then(
+							// eslint-disable-next-line no-loop-func
+							function onVisitedPromiseUnwrapped(value) {
```
idreyn, line 72: Looks like there's nothing to do here but disable the linter, but I'm mildly curious whether it produces the same complaint for arrow functions.

idreyn (approved):
Looks good to me! A few more smoke testing suggestions:

- Login flow
- Password reset flow
- The `asyncForEach` in `createPub` — you can test this by creating a new Banner Block witha Create Pub button on a page/collection layout, adding many default collections to it, and then using the button to create a Pub and verifying that the Pub is in all of those Collections.
- You could use the `clone` tool to make sure `asyncMap` works to clone all of the Pubs in a Community.

PR #1687, opened 2021-12-01
update/fix: Use publisher for copyright

### LINKED ISSUE #596: add cc badge images to pub footer license
### Required

---

**What's your request?**
Add badge images for cc licenses to the Pub footer license section.

It's easy to mis-read the CC licenses as text since they all start and end similarly. It would be nice to display the badge in addition to the text, since they're more recognizable and easier to parse at a quick glance.

### Optional

---

gabestein:
Ended up being a larger one than I anticipated. This PR:

- Uses the publisher field for the Pub copyright license text, if filled out
- Fixes the copyright license display for PDF and JATS
- Fixes(?) a bug where the copyright date for a book was always displaying the prior year due to timezone issues
- Resolves #596 by adding images to cc-by licenses in the pub footer, because why not while I was down there?

Flew a little blind here and had to add some data from up in the rendering tree, so definitely open to suggestions of better ways to organize. But I think it's pretty straightforward.

Screen Shot 2021-12-01 at 16 56 48
Screen Shot 2021-12-01 at 16 56 42
Screen Shot 2021-12-01 at 17 05 01
Screen Shot 2021-12-01 at 17 06 32

_Test Plan_
1. Create a pub in a community that has the premium license flag turned on and no publisher set in settings, and set the license of that pub to copyright.
2. Make sure the license is `Copyright (c) {Community Title} {Pub Creation Date}. All rights reserved.`
3. Download PDF and JATS and make sure the license matches.
4. Add a publisher to community settings. Visit the pub and make sure the license is: `Copyright (c) {Publisher} {Pub Creation Date}. All rights reserved.`
5. Check the PDF and JATS to make sure they match (you will have to make body changes to regen).
6. Add the pub to a public book collection with a copyright year different from this year set (e.g. 2020). Make sure the license is: `Copyright (c) {Community Title} {Book Copyright Year}. All rights reserved.`
7. Check the PDF and JATS to make sure they match (you will have to make body changes to regen).
8. Add the pub to a public issue collection with a publication year different from this year set (e.g. 2020). Make sure the license is: `Copyright (c) {Community Title} {Pub Release Year}. All rights reserved.`
9. Check the PDF and JATS to make sure they match (you will have to make body changes to regen).
10. Remove the pub from those collections and release it, setting a custom release date with a year different than this one (e.g. 2020). Make sure the license is: `Copyright (c) {Community Title} {Custom Publication Date Year}. All rights reserved.`
11. Check the PDF and JATS to make sure they match (you will have to make body changes to regen).


3mcd:
Hey @gabestein, there's a 99% chance I missed something, but I'm getting stuck at at test step #8. I've added the [test Pub](https://qubqub-pr-1687.herokuapp.com/dash/pub/7gfelsrj/settings) to a public journal issue [collection](https://qubqub-pr-1687.herokuapp.com/dash/collection/export-primary-collection-issue/settings) with the following publication date(s):

Screen Shot 2021-12-06 at 2 20 54 PM

When I make a change to the Pub, and re-export to PDF, I still see the pub creation year (2021).

client/containers/Pub/PubDocument/PubBottom/pubBottom.scss
```diff
@@ -1,35 +1,53 @@
...
+		}
+	}
+}
+
+@media screen and (max-width: 720px) {
+	.pub-bottom-license {
+		.license-link {
+			flex-direction: column;
+			.license-image {
+				margin: 0 0 10px 0;
```
idreyn, line 49: $0.02: I'd prefer the `margin-x` property here — I can never remember what order these go in.

utils/pub/pubDates.ts
```diff
@@ -77,8 +77,8 @@ export const getPubCopyrightYear = (
 		if (metadata) {
 			const { copyrightYear, date, publicationDate } = metadata;
 			const dateSource = copyrightYear || date || publicationDate;
-			if (dateSource) {
-				return dateFormat(dateSource, 'yyyy');
+			if (dateSource && new Date(dateSource).getDate()) {
```
idreyn, line 80: Could you leave a comment about what's happening here?

workers/tasks/export/html.tsx
```diff
@@ -176,20 +190,28 @@ const renderFrontMatter = ({
...
 	attributions,
 	licenseSlug,
+	copyrightYear,
 	publisher,
 }: PubMetadata) => {
 	const affiliations = [
 		...new Set(attributions.map((attr) => attr.affiliation).filter((x) => x)),
 	];
 	// do not put community title if this is a book
+	const pubPublisher =
```
idreyn, line 207: This is another thing that feels like it should go into a utility function somewhere, if only so we're forced to name it.

workers/tasks/export/pandoc.ts
```diff
@@ -91,7 +93,12 @@ const createYamlMetadataFile = async (pubMetadata: PubMetadata, pandocTarget: st
 			title: communityTitle,
 		},
 		copyright: {
-			text: license.full,
+			text:
+				license.short === 'Copyright'
```
idreyn, line 97: I also feel like this should be in a utility function, maybe in `utils/licenses.js`?

client/components/LicenseSelect/LicenseSelect.tsx
```diff
@@ -33,10 +33,7 @@ const LicenseSelect = (props: Props) => {
 	const { communityData } = usePageContext();
 	const currentLicense = getLicenseBySlug(pubData.licenseSlug)!;
 	const pubCopyrightYear = getPubCopyrightYear(pubData);
-	let pubPublisher = communityData.title;
-	if (communityData.id === '78810858-8c4a-4435-a669-6bb176b61d40') {
-		pubPublisher = 'Massachusetts Institute of Technology';
-	}
+	const pubPublisher = communityData.publishAs || communityData.title;
```
idreyn, line 36: Since this is being rewritten in two places, and seems important to name, I'd suggest a utility function in `utils/license.js` or maybe a `utils/community.ts` with the following signature:

```ts
getPublisherString(community: Pick<Community, 'publishAs' | 'title'>): string
```

idreyn (changes requested):
Lots of important details being rendered in this code, and even though the logic is very simple it feels important to be able to encapsulate them in some utility functions. More than anything else this will help us give a concise name to behaviors like "picking the publisher string" so we can refer to and change them easily later.

client/containers/Pub/PubDocument/PubBottom/pubBottom.scss
gabestein: Heh. My habits from before anyone compiled CSS are still going strong.

utils/pub/pubDates.ts
gabestein: In the code or here?

Long story short: this fixes two bugs. 

1. Currently, if you add something in that meta field that is not interpretable as a date, it will break any Pubs in the collection. That's what checking `new Date(...)` in the conditional prevents. I've also added some form validation to prevent it, so we could probably remove this. 

2. There's a timezone-related bug where, if your local time is > UTC, the copyright will always be displayed as the year prior to what's actually entered, because it stores it as midnight UTC on January 1. `Date.UTC(...)` is a bit of a hack to make it always display the year as entered, which is the intended behavior independent of timezone. It does this by asserting that the date is actually February 1 UTC.

gabestein:
> Hey @gabestein, there's a 99% chance I missed something, but I'm getting stuck at at test step 8. I've added the [test Pub](https://qubqub-pr-1687.herokuapp.com/dash/pub/7gfelsrj/settings) to a public journal issue [collection](https://qubqub-pr-1687.herokuapp.com/dash/collection/export-primary-collection-issue/settings) with the following publication date(s):
> 
> Screen Shot 2021-12-06 at 2 20 54 PM
> 
> When I make a change to the Pub, and re-export to PDF, I still see the pub creation year (2021).
> 
> Screen Shot 2021-12-06 at 2 28 17 PM

Interesting...I actually think this is a question for @Dawit-Tegbaru. In a hypothetical scenario where a journal is published in 2020, but an article from it comes out in 2021...should the copyright be 2020 or 2021? I presume the latter, actually, since you couldn't copyright something before it existed (and I think this is why TMB is closing issues prior to year-end), but it's a weird grey area.

workers/tasks/export/pandoc.ts
gabestein: Not sure the best way to do this, because it's already a bit of a hack. Copyright is the only license that relies on info about the pub and community. So if we turn it into a helper, we're forcing ourselves to always pass that information to it. That might just be the way to do it, but it feels a bit inefficient given that most licenses don't need any additional info.

3mcd (approved):
Gonna go ahead and approve this since @idreyn left a thorough technical review, and everything worked fine in testing except the one callout I made.

Dawit-Tegbaru:
> 

So something published in 2022 could have a 2021 copyright statement. The year is typically based on the accept date (or more officially) the time copyright was obtained - which usually happens right after acceptance. 

E.g. 
Accepted ->  November 2021
Published - Jan 2022
Copyright 2021

gabestein:
Gotcha...because we don't have a field for accepted date, I think our current behavior is probably the ideal one. I've adjusted the testing plan for it.

workers/tasks/export/pandoc.ts
gabestein: Ended up putting it all into the existing getLicenseBySlug function. See what you think. Could probably make it more typescript-y, although it gets called from places with two different styles of metadata objects, so idk.

gabestein:
ok @idreyn I'm ready for your next pass.

utils/pub/pubDates.ts
idreyn: There are functions called `isValidDate` and `getLocalDateMatchingUtcCalendarDate` in `utils/dates.ts` that I think you might be able to use here. It might be nice to modify `isValidDate` to take `string | Date` and do the cast to `Date` internally if necessary.

workers/tasks/export/pandoc.ts
idreyn: That's the right call, IMO. The fact that it's a hack makes it even _more_ necessary to encapsulate it in a function where the "hackiness" is contained by contract with the caller. I'd argue that we should take it one step further and write a new `getLicenseForPub` function:

```ts
// Taking the License type to be the object that `getLicenseBySlug` returns...
type RenderedLicense = License & {
    summary?: string;
}

type PubWithCollections = Pub & { 
    collectionPubs: DefinitelyHas<CollectionPub, 'collection'>[];
}

const getLicenseForPub = (pub: PubWithCollections, community: Community): RenderedLicense => {
    // call getPubCopyrightYear and getPublisherString in here...
}
```

There's no getting around the fact that we _do_ need a Community and a Pub in scope to correctly compute a Pub's license now, and while I see the temptation to move that responsibility to the caller of `getLicenseBySlug` like this...

```ts
const pubCopyrightYear = getPubCopyrightYear(pubData as any);
const publisherString = getPublisherString(communityData);
const { link, full, slug, summary } = getLicenseBySlug(
    pubData.licenseSlug,
    publisherString,
    pubCopyrightYear,
)!;
```

I weigh that against two other factors:

1. It seems unlikely we'll ever use `getLicenseBySlug(pub, publisherString, pubCopyrightYear)` to do anything other than this exact computation on the results of `getPubCopyrightYear` and `getPublisherString`. And _this_ block of computation still relies on having a Pub and a Community in scope. So we might as well encapsulate it.
2. Having a Pub and a Community to work with in our codebase is _really_ common. We almost always have these at hand when we need them, from `scopeData` on the server and `usePageContext` on the client.

Happy to get on a call and discuss the implementation of `getLicenseForPub` in more detail (or to do it myself). I'm only being persnickety because I have a strong sense that this will lead to much more predictable behavior down the line.

utils/licenses.js
```diff
@@ -58,6 +58,12 @@ export const licenses = [
 	},
 ];
 
-export const getLicenseBySlug = (slug = 'cc-by') => {
-	return licenses.find((ls) => ls.slug === slug);
+export const getLicenseBySlug = (slug = 'cc-by', publisherString, copyrightYear) => {
+	const license = licenses.find((ls) => ls.slug === slug);
+	license.summary = `${license.slug.toUpperCase()} ${license.version}`;
```
idreyn, line 63: This will modify the object in `licenses` for future callers. You'd want to shallow-copy it like this before mutating it:

```ts
const license = { ...licenses.find((ls) => ls.slug === slug) };
```

idreyn (changes requested):
Left a few more comments about encapsulation here.

gabestein:
Third time's the charm?

idreyn (approved):
Looks great! Thanks for following up with all of my requested changes.

gabestein:
Thanks! I don't know why that test is failing, though...not showing up locally at all, and we assert the existence of license (which requires a slug)...
```
  ● Storyshots › containers/Pub/PubDocument/PubBottom/PubBottom › no-discussions
    TypeError: Cannot read property 'slug' of undefined
      73 | 	const renderedLicense = {
      74 | 		...license,
    > 75 | 		summary: `${license.slug.toUpperCase()} ${license.version}`,
         | 		                    ^
      76 | 	};
      77 | 	const pubCopyrightYear = getPubCopyrightYear(pub);
      78 | 	const publisherString = getPublisherString(community);
      at getLicenseForPub (utils/licenses.ts:75:23)
      at LicenseSection (client/containers/Pub/PubDocument/PubBottom/LicenseSection.tsx:18:56)
```

idreyn:
> Thanks! I don't know why that test is failing, though...not showing up locally at all, and we assert the existence of license (which requires a slug), on line 72...

You'll need to add a `licenseSlug` to `pubData` in `PubBottom.stories.tsx`. (The `!` assertion is just a claim we make to TypeScript to make it turn a blind eye to nullable values — it won't compile down to any kind of runtime assertion checking.)

PR #1701, opened 2021-12-08
update: Change language from submitted to pending

qweliant:
this just changes the fields 'submitted' to 'pending'. this feels more in line with how the field is being used and has been talked about

test:
run submission test to be sure changes do not effect expected behavior


howebenhowe (approved):
Looks good!

PR #1702, opened 2021-12-08
Implement "notification cadence" in PubPub

idreyn:
I've had this crazy idea for a tool that prevents compulsive notification-checking for a while, that I call _notification cadence_. The rules go like this:

1. You pick a cadence `c`, which is a time interval, given in minutes, that we store
2. When you check your notifications, we mark this time as `t0`
3. Your notifications are now frozen in time — no more new ones will appear for `c` minutes
4. At `t1 >= t0 + c` you can check again and we'll show you all your notifications, including those created on `[t0, t1]`
5. Your notifications are now frozen again until `t1 + c`. The cycle continues...

(For `c = 0` you'll receive notifications instantly)

We have a field in `UserNotificationPreferences` for notification cadence, and a dropdown that lets you select one:

image

This PR just implements the feature.

_Test plan:_
```
npm run test-dev server/userNotification
```

### DIFF
--- a/server/userNotification/__tests__/api.test.ts
+++ b/server/userNotification/__tests__/api.test.ts
@@ -3,12 +3,19 @@ import uuid from 'uuid/v4';
 
 import { UserNotification } from 'server/models';
 import { modelize, login, setup, teardown } from 'stubstub';
+import { fetchUserNotifications } from '..';
 
 const communityId = uuid();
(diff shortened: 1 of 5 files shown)


3mcd (approved):
This is really cool, and thanks for the comments in the test

PR #1708, opened 2021-12-13
dev: Add type definitions to `usePageContext`

idreyn:
The `usePageContext()` hook is everywhere but we always have to cross-reference other callsites to remind ourselves what values it provides. No longer! I have defined its return value in TypeScript and fixed all the resulting little type issues.

_Test plan:_
Going out on a bit of a limb to say there shouldn't be any behavioral changes (except for a bug I found in `PubDiscussions` that the new types caught). So a typecheck should be sufficient?


3mcd (approved):
Beautiful! LGTM

PR #1710, opened 2021-12-15
feature: Submission Tab Styling

qweliant:
Add details row design. This adds new color bocks and submission dates to the details row. This can be seen below the contributors row
image

Test: 
Run storybook test
 
Check Submission tab in collection overview and compare submission status and submitted date with database.
Pending = yellow box | should have a submission date
Accepted = green box | should have a submission date
Declined = red box | should have a submission date
Incomplete = grayed out box, no submission date


client/containers/DashboardOverview/overviewRows/CollectionOverviewRow.tsx
```diff
@@ -25,6 +25,37 @@ type Props = {
 	onToggleOpen?: () => unknown;
 };
 
+const labelPairs = (iconLabelPairs: IconLabelPair[]) => {
```
idreyn, line 28: I think this would have a natural home as a utility that gets imported from `labels.tsx` instead of duplicated across several files. It should also have a more verb-y name, maybe `renderLabelPairs`.

client/containers/DashboardOverview/overviewRows/OverviewRowSkeleton.tsx
```diff
@@ -22,7 +16,7 @@ type Props = {
 	byline?: React.ReactNode;
 	rightElement?: React.ReactNode;
 	darkenRightIcons?: boolean;
-	iconLabelPairs: IconLabelPair[];
+	iconLabelPairs: React.ReactNode;
```
idreyn, line 19: Since this can now take _any_ content, `iconLabelPairs` is too specific a name. Maybe just `details`?

client/containers/DashboardOverview/overviewRows/PubOverviewRow.tsx
```diff
@@ -19,13 +24,46 @@ type Props = {
...
+		</div>
+	);
 };
 
 const PubOverviewRow = (props: Props) => {
 	const {
 		pub,
 		className,
 		inCollection,
+		isSubmission,
```
idreyn, line 66: Gonna nitpick here and say that this should be `hasSubmission` to be consistent with how we model `Submissions` and `Pubs`.

client/containers/DashboardOverview/overviewRows/PubOverviewRow.tsx
```diff
@@ -38,16 +76,40 @@ const PubOverviewRow = (props: Props) => {
...
+	// for when a pub has a submission but is being rendered in collection or community overview
+	if (pub.submission) {
+		const { status } = pub.submission;
+		const iconLabelPairs = isSubmission
+			? labelPairs([getSubmissionStatusLabel(status), getSubmissionTimeLabel(pub)])
+			: labelPairs([
+					...getScopeSummaryLabels(pub.scopeSummary),
+					getPubReleasedStateLabel(pub),
+			  ]);
+		return (
```
idreyn, line 89: Removing duplication here will make it easier for readers to understand how these two codepaths diverge, and to make future changes. As far as I can tell the only difference between this `<OverviewRowSkeleton>` and the one rendered underneath is the `iconLabelPairs` prop. Let's hone in on that — we need a function like:

```ts
renderIconLabelPairs(pub: Pub, hasSubmission: boolean): React.ReactNode
```

The conditional logic you've added here should be contained in this function, which will return the thing that we can unconditionally pass into `<OverviewRowSkeleton>`.

client/containers/DashboardOverview/overviewRows/labels.tsx
```diff
@@ -78,3 +76,52 @@ export const getPubReleasedStateLabel = (pub: Pub) => {
...
 	};
 };
+
+export const getSubmissionStatusLabel = (
+	submissionStatus: Partial<SubmissionStatus>,
+): IconLabelPair => {
+	if (submissionStatus === 'accepted') {
+		return {
+			label: 'Accepted',
+			icon: 'symbol-square' as const,
```
idreyn, line 86: Does this work if you remove `as const`? I kinda think it should but I'm not sure.

client/containers/DashboardOverview/overviewRows/labels.tsx
```diff
@@ -78,3 +76,52 @@ export const getPubReleasedStateLabel = (pub: Pub) => {
...
+	}
+	return {
+		label: 'Incomplete',
+		icon: 'symbol-square' as const,
+		intent: 'none',
+	};
+};
+
+export const getSubmissionTimeLabel = (pub: Pub) => {
+	const publishedDate = getPubSubmissionDate(pub);
```
idreyn, line 112: This is not a "published date" — rename accordingly.

client/containers/DashboardSubmissions/SubmissionItems.tsx
```diff
@@ -76,7 +77,12 @@ const SubmissionItems = (props: Props) => {
 
 	const renderPubs = () => {
 		return pubs.map((pub) => (
-			<PubOverviewRow pub={pub} key={pub.id} leftIconElement="manually-entered-data" />
+			<PubOverviewRow
+				pub={pub}
+				key={pub.id}
+				leftIconElement="manually-entered-data"
+				isSubmission={true}
```
idreyn, line 84: You can just write `isSubmission` instead of `isSubmission={true}` here (if you like)

tools/migrations/2021_12_15_addSubmittedAtField.js
```diff
@@ -0,0 +1,9 @@
+export const up = async ({ sequelize, Sequelize }) => {
```
idreyn, line 1: This is great and very thoughtful but I don't think we'll actually need to run this. Sequelize _should_ create new columns for us when we add them to our models (and fwiw, since this model isn't in production yet, we can always just wipe out the table on dev instead of performing a migration of any complexity if it saves us trouble)

types/iconLabel.ts
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
idreyn, line 1: The `IconLabelPair` type should live closer to the overview row components (maybe in `labels.tsx`?) The top-level `types` directory is for types that we expect to use all over the codebase.

utils/pub/pubDates.ts
```diff
@@ -85,3 +85,13 @@ export const getPubCopyrightYear = (
 	const pubPublishedDate = getPubPublishedDate(pub);
 	return pubPublishedDate ? dateFormat(pubPublishedDate, 'yyyy') : dateFormat('yyyy');
 };
+
+export const getPubSubmissionDate = (pub: Pub) => {
+	if (pub.submission?.submittedAt) {
+		// This is a date string representing a time at midnight UTC for a given date.
```
idreyn, line 91: This is a good comment but I actually think (a variation of) it should live in `getLocalDateMatchingUtcCalendarDate`.

idreyn (changes requested):
Requesting changes, but this has the bones of a good clean refactor. Some general thoughts:

1. Nice job generalizing `IconLabelPair` to encompass the additions we need to make. It makes the refactoring of `OverviewRowSkeleton` to take `React.ReactNode` instead of `IconLabelPair[]` temporarily unnecessary, but this is a good move. It will now be easy to render the Member avatars or some other stuff down there that truly isn't an icon-label pair.

2. There's a lot of de-duplication that can be done in this PR. I'll mention specifically where in the comments, but I bet you'll find it yourself if you step into a reader's mindset and consider where the code could be made easier to understand and change 6-12 months from now.

3. Names! I have a few suggestions for more precise or descriptive variable names, and I also want to encourage you to think more about variable names generally, to help signal your intent to readers, and as a tool to clarify your own design thinking. I think there are outsize returns for a little bit more focus here.

4. For changes that have a major visual component, screenshots should be _de rigueur_ on our PRs.

client/containers/DashboardSubmissions/SubmissionItems.tsx
qweliant: i do like explicit declaration here

client/containers/DashboardOverview/overviewRows/labels.tsx
```diff
@@ -78,3 +84,101 @@ export const getPubReleasedStateLabel = (pub: Pub) => {
...
+			label: (
+				<>
+					<em>Submitted</em>&nbsp;{getDateLabelPart(submissiondDate)}
+				</>
+			),
+			icon: 'time' as const,
+		};
+	}
+	return {
+		label: 'Update Status',
```
idreyn, line 132: What does this mean? If there's no submission date yet, we should just not render anything.

client/containers/DashboardOverview/overviewRows/labels.tsx
```diff
@@ -78,3 +84,101 @@ export const getPubReleasedStateLabel = (pub: Pub) => {
...
+	}
+	return {
+		label: 'Incomplete',
+		icon: 'symbol-square',
+		intent: 'none',
+	};
+};
+
+export const getSubmissionTimeLabel = (pub: Pub) => {
+	const submissiondDate = getPubSubmissionDate(pub);
```
idreyn, line 120: Small typo here.

client/containers/DashboardOverview/overviewRows/labels.tsx
```diff
@@ -78,3 +84,101 @@ export const getPubReleasedStateLabel = (pub: Pub) => {
...
+						{iconElement}
+						{label}
+					</div>
+				);
+			})}
+		</div>
+	);
+};
+
+export const renderDetailsRow = (pub: Pub, hasSubmission: boolean): React.ReactNode => {
```
idreyn, line 168: Okay, I see the problem now with my suggestion to call this `hasSubmission` — the obvious check to see where a pub "has submission" is just to check `pub.submission`. I guess I'd call this something like `showSubmissionDetails`?

idreyn (approved):
Screenshot look great — a few more things inline but I don't wanna block you!

client/containers/DashboardOverview/overviewRows/labels.tsx
qweliant: pretty much, i just didn't want to have something look like this

```
...
if (pub.submission) {
     const { status } = pub.submission;
     const detailsRow = pub.submission
	     ? renderLabelPairs([getSubmissionStatusLabel(status), getSubmissionTimeLabel(pub)])
	     : renderLabelPairs([
	           ...getScopeSummaryLabels(pub.scopeSummary),
	           getPubReleasedStateLabel(pub),
	     ]);
     return detailsRow;
......
```
because a pub may have a submission, but render in the collections or community overview tab. explicitly passing isSubmission seems to sign post if the `PubOverviewRow ` component being rendered is from the `SubmissionOverview` render the submissions details row. hasSubmission feels more natural. 

`renderDetailsRow` is rendering and `IconLabelPair` (or whatever else we want there) wherever `PubOverviewRow` is being called so what about `renderDetails`?

client/containers/DashboardOverview/overviewRows/labels.tsx
qweliant: That makes sense, for some reason I was thinking a pub without a submission status still needs to be rendered

PR #1711, opened 2021-12-15
fix: Export abstract to JATS with <p> wrapping intact

### LINKED ISSUE #1695: wrap abstracts in <p> tags for pandoc exports
There was still one validation error in the new file.JATS does not allow PCDATA directly in abstract. It should be wrapped with a `<p>` or other display elements (https://jats.nlm.nih.gov/archiving/tag-library/1.2/element/abstract.html).

idreyn:
Resolves #1695

An `<abstract>` tag in well-formed JATS should have its text wrapped in `<p>` tags. This change tweaks our exporter to make that happen.

_Test plan:_
Create a new Pub with an "Abstract" H1 and add some text to the following paragraph. The JATS XML export of the Pub should have an `<abstract>` tag containing a `<p>` tag which itself contains the abstract content.


gabestein (approved):
LGTM!

PR #1712, opened 2021-12-15
feat: activity digest email tool

3mcd:
**Description**

This PR expands on @howebenhowe's work:
- adds the a column `receivesDigestEmail` to the member table
- adds a [`Switch` input](https://github.com/pubpub/pubpub/pull/1712/files#diff-d4d0f63a20184adab7795572538b339f39370fc06a7d52108ae4f1f45fac0bc1R53) to the Activity dashboard page which toggles the member's `receivesDigestEmail` field. It currently only appears for members of a community with either `"admin"` or `"manage"` permission
- added a tool named [`emailActivityDigest`](https://github.com/pubpub/pubpub/pull/1712/files#diff-e1085e0eaeee53bff6ca2b04fc3d6e0b2c796c92439595470ed3c56aea43c0f6) which iterates batches of communities and renders/sends activity digests to each community member where `member.receivesDigestEmail==true`
- widens the `Digest.tsx` template a bit (from 600px to 640px), centers it on the X-axis, and adds an [invisible preview](https://github.com/pubpub/pubpub/pull/1712/files#diff-6395f0fa3f065d18adb2462152cdd22e8d8173aae0e537849da5bf1acc0d8c42R23) that appears in some email clients' subject lines, but not in the email body

**Screenshots**

Gmail subject line with hidden preview:
Screen Shot 2022-01-05 at 3 34 18 PM

Email content (showing increased width and centering):
Screen Shot 2022-01-05 at 3 01 46 PM

"Enroll" switch:
Screen Shot 2022-01-05 at 3 46 53 PM

**Test Steps**

* Give yourself admin or manage permission to a community.
* Visit the Activity dashboard page for that community, and flip the "Enroll in activity digest emails" switch in the top-right of the page switch on (it should be off by default).
* Run the new activity digest tool: `npm run tools emailActivityDigest`
  * _If you want to speed this up, you can copy `where: { subdomain: <community subdomain> }` after line 220 in `server/community/queries.ts`_
* Wait a few seconds and you should get an email!


gabestein (changes requested):
Super, super exciting! Game-changing! And other superlatives. From a ux/functionality perspective, looks great to me. Emails send as expected, look amazing, and are super helpful for both more and less active communities.

A few very small things:

- URLs for everything other than the community itself appear to be relative (e.g. `/user/eric`). Might be a result of running it locally?
- Add "weekly" to the toggle text (`Enroll in weekly Activity Digest Emails`).
- As discussed: instead of News, which feels a bit word, have the sections be `Community`, `Pubs`, `Collections` (do we even have a collections segment?).
- If there is no activity in Pubs, the header for the section still shows up. I think that's actually probably better than not showing the section at all, but we might want to add 'No activity this week.' in those empty state cases for all sections.
- As discussed: no need for unsubscribe link. I think it's totally fine, given the opt-in nature of this email, to just have the link to the activity page.

client/containers/DashboardActivity/DashboardActivity.tsx
```diff
@@ -43,7 +46,30 @@ const DashboardActivity = (props: Props) => {
 	});
 
 	return (
-		<DashboardFrame className="dashboard-activity-container" title="Activity">
+		<DashboardFrame
+			className="dashboard-activity-container"
+			title="Activity"
+			controls={
+				canEnrollInActivityDigestEmails && (
```
idreyn, line 53: ❗️ We also only want to show this the Community scope, right? As-is this appears on Collection/Pub activity dashboards but still targets the parent Community.

client/containers/DashboardActivity/DashboardActivity.tsx
```diff
@@ -43,7 +46,30 @@ const DashboardActivity = (props: Props) => {
...
+								.updateMember({
+									member,
+									update: {
+										receivesDigestEmail: (e.target as HTMLInputElement).checked,
+									},
+									scopeIds: { communityId: scope.communityId },
+								})
+								.then(setMember)
+						}
+						label="Enroll in Activity Digest Emails"
```
idreyn, line 68: $0.02 — I'd use "subscribe" over "enroll" here

client/containers/DashboardActivity/DashboardActivity.tsx
```diff
@@ -43,7 +46,30 @@ const DashboardActivity = (props: Props) => {
...
-		<DashboardFrame className="dashboard-activity-container" title="Activity">
+		<DashboardFrame
+			className="dashboard-activity-container"
+			title="Activity"
+			controls={
+				canEnrollInActivityDigestEmails && (
+					<Switch
+						checked={member.receivesDigestEmail}
+						className="parent-approval-switch"
+						onChange={(e) =>
```
idreyn, line 57: Would you mind factoring out a local `setEnrolledInActivityDigest(enrolled: boolean) => Promise<Partial<Member>>` here? The inlining is visually busy.

client/utils/activity/render.ts
```diff
@@ -10,5 +10,5 @@ export const renderActivityItem = (
 	// This cast avoids a TypeScript error:
 	// Expression produces a union type that is too complex to represent. ts(2590)
 	const renderer = activityItemRenderers[item.kind] as ActivityItemRenderer<any>;
-	return renderer(item, context) as RenderedActivityItem;
+	return (renderer?.(item, context) as RenderedActivityItem) || null;
```
idreyn, line 13: ❗️Looking at my IDE I don't think this is necessary (and I also don't know how TS is accepting `null ∈ RenderedActvitiyItem` with our compiler settings! Could you (1) try reverting this and see if it's okay and (2) if not, add a `!` non-null assertion to the extraction of `renderer` on the previous line?

My high-level goal here is for callers and readers of this function to be confident that they'll unconditionally get a `RenderedActivityItem` for any `ActivityItem`.

server/community/queries.ts
```diff
@@ -207,3 +210,17 @@ export const isUserAffiliatedWithCommunity = async (userId: string, communityId:
 	const isHere = counts.some((c) => c > 0);
 	return isHere;
 };
+
+export const iterAllCommunities = async function* (limit = 10): AsyncGenerator<types.Community[]> {
```
idreyn, line 214: 🏆 Achievement unlocked: first use of generators in production PubPub code!

tools/emailActivityDigest.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const mg = mailgun.client({
+	username: 'api',
+	key: process.env.MAILGUN_API_KEY!,
+});
+
+const memberQueryOptions = {
+	include: [includeUserModel({ attributes: ['email'], as: 'user' })],
+	permissions: { [Op.or]: ['admin', 'manage'] },
+	raw: true,
```
idreyn, line 20: Can I prod at what `{ raw, nest }` are here for? If it's related to memory concerns, could you add a comment about that?

idreyn (changes requested):
This is so cool. I've really enjoyed getting all the test emails in my inbox (and thanks @howebenhowe for your half of the work on them, too!)

The only things that _need_ to be addressed are labelled with ❗ and the rest are just suggestions.

client/components/Email/Wrapper.tsx
```diff
@@ -13,15 +14,45 @@ const TableStyle = styled(BaseTableStyle).attrs({
...
+			style={{
+				display: 'none',
+				fontSize: '1px',
+				lineHeight: '1px',
+				maxHeight: '0px',
+				maxWidth: '0px',
+				opacity: 0,
+				overflow: 'hidden',
+			}}
+			// eslint-disable-next-line react/no-danger
```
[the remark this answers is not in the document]
3mcd: Agreed!!

client/containers/DashboardActivity/DashboardActivity.tsx
3mcd: So "subscribe to"?

client/utils/activity/render.ts
3mcd: So Gabe discovered this issue as he was testing this PR. The activity dashboard page was crashing because the `activityItemRenderers` map didn't have an entry for a submissions-related activity item. This was a quick fix without having to block testing/merging this PR until we had renderer(s) written for the culprits. Would you like me to implement those as part of this PR?

tools/emailActivityDigest.ts
3mcd: Yeah, memory concerns, although now that I'm using a sliding window I probably don't need them.

client/containers/DashboardActivity/DashboardActivity.tsx
idreyn: I think so. Enrolling makes it sound...arduous...to me 😄

client/utils/activity/render.ts
idreyn: Oh, I should have seen that. Don't worry about it for now — I'll make a card to fix this up on the project board.

3mcd: Sounds good :D

idreyn:
side note, I love the "feat" prefix on this PR — I'd like to imagine that it's not a "feature" but a ["feat of strength"](https://www.youtube.com/watch?v=HX55AzGku5Y)

idreyn (approved):
Thanks for the changes, LGTM!

PR #1732, opened 2022-01-04
update: Integrate Workflow editor and Submissions listing into dashboard

idreyn:
This PR makes the Submission Workflow editor and the Submissions listing coexist on the same dashboard tab. I've settled on using two different routes/container for this: the listing is at `/dash/collection/:slug/submissions` and the workflow editor is at `/dash/collection/:slug/submissions/worfklow`. When you click on the Submissions tab for the first time you'll be redirected to create a workflow. This has the nice side effect of using the breadcrumbs as they were intended:

image

While I was in here I also added a button on the Submissions listing to enable/disable submissions:

image

_Test plan:_
- For a Collection without a `SubmissionWorkflow`, clicking the Submissions tab should redirect you to the workflow creator. 
- Click "Save and Continue" to be redirected to the submissions listing
- Check that the "Edit Workflow" button takes you back to the workflow editor, which allows you to make changes
- Check that the enable/disable submissions button works.


client/containers/DashboardSubmissionWorkflow/ExistingSubmissionWorkflowEditor.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+	const {
+		state: workflow,
+		update: updateWorkflow,
+		persist,
+		hasChanges,
+		isPersisting,
+	} = usePersistableState(initialWorkflow, async (update, full) => {
+		await updateSubmissionWorkflow(update, activeCollection.id);
+		// Give the usePersistableState hook disable its onBeforeUnload before the parent redirects
```
3mcd, line 35: I'm not 100% sure what this comment means, could you help clarify it for me?

3mcd (approved):
Nice work!

client/containers/DashboardSubmissionWorkflow/ExistingSubmissionWorkflowEditor.tsx
idreyn: Oh oops that was grammatically incoherent but even then still kind of hard to parse. I wrote a better one.

PR #1743, opened 2022-01-09
Qdt/pubheader skeleton

qweliant:
This adds the three tabs that will eventually be the spub header
image

test:

run storybook in terminal. check that the PubSubmissionHeader component renders and switches between views


client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,72 @@
...
+$header-margin-top: 0px;
+$header-margin-bottom: 26px;
+$header-margin-bottom-mobile: 13px;
+$header-margin-total: $header-margin-top + $header-margin-bottom;
+$header-margin-total-mobile: $header-margin-top + $header-margin-bottom-mobile;
+$title-elements-spacing: 15px;
+$mobile-bottom-buttons-padding-top: 15px;
+$mobile-bottom-buttons-spacing: 10px;
+
+.pub-header-component {
```
idreyn, line 13: It's important (and non-obvious) that when you `require('./spubHeader.scss')` you're loading those styles _globally_, for the whole application, so it's not just for cosmetics that this class needs to have a unique name.

Looking at the difference between these styles and those in the existing `pubHeader.scss`, it looks like this is a copy-paste of that file with some stuff taken out. I'm tempted to say that this is fine for MVP-mode development, but I also wonder how far you'd get by factoring the shared styles into a `@mixin pub-header-basic` that's then `@include`-ed from both files.

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,72 @@
+@import 'styles/variables.scss';
+@import './themes.scss';
+
+$header-margin-top: 0px;
```
idreyn, line 4: Looks like these don't get used here and should be safe to delete

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,72 @@
...
+		text-align: left;
+
+		box-decoration-break: clone;
+		@include smaller-viewport {
+			font-size: 32px;
+		}
+	}
+}
+
+.bp3-tabs {
```
idreyn, line 46: All of these `.bp3-xyz` classes need to be scoped inside the (to be renamed) `pub-header-component` class up there, otherwise they'll target every control of this type in the application.

client/containers/Pub/PubSubmissionHeader/themes.scss
```diff
@@ -0,0 +1,269 @@
+$mobile-viewport-cutoff: 750px;
```
idreyn, line 1: This duplicated file is full of SCSS mixins, which you can think of as like compile-time CSS functions. And like a directory full of functions, it'd be most sensible to

```scss
@import "../PubHeader/themes.scss";
```

from `spubHeader.scss` rather than duplicating them here.

client/containers/Pub/PubSubmissionHeader/PubSubmissionHeader.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import PubSubmissionHeader from './PubSubmissionHeader';
+// import { locationData, loginData, communityData } from 'utils/storybook/data';
+
+storiesOf('containers/Pub/PubSubmissionHeader', module).add('default', () => (
+	<PubSubmissionHeader />
```
idreyn, line 8: This component will need to take a `Submission` as a prop, so I'd suggest that you start constructing a fake one here:

```ts
const submission: Submission = {
    instructionsText: { ... },
    // and all the rest
}
```

Since this type has so many `DocJson` properties you might find the `getEmptyDoc()` function in `components/Editor` helpful for spoofing them. And if you really, you can probably get away with defining the `submission` type as `Pick<Submission, 'a' | 'b' | 'c'>` where `'a' | 'b' | 'c'` are only the properties you currently need (which for now is just the `instructionsText`.

client/containers/Pub/PubSubmissionHeader/PubSubmissionHeader.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import PubSubmissionHeader from './PubSubmissionHeader';
+// import { locationData, loginData, communityData } from 'utils/storybook/data';
```
idreyn, line 5: Zap commented-out code

idreyn (changes requested):
So this is actually looking really slick, but there are three big structural things that need to be addressed:

1. Your component needs to mount `DocJson` in an `Editor`
2. Example data should be moved into the `.stories.tsx` file and passed in as a prop
3. Your CSS needs to be scoped and de-duplicated

I go into all of this in some detail below but let me know if you want to hop on a call and get into it some more!

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,43 @@
+.spub-header-component {
+	border-bottom:	border-bottom: 1px solid #d2d2d2;
+;
+	padding: 200px;
```
idreyn, line 4: Wait...what is this massive padding? 😆 I assume it's helpful for Storybook, but please remove it before landing.

client/containers/Pub/PubSubmissionHeader/PubSubmissionHeader.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+			meta: {},
+		},
+		content: [],
+	},
+	instructionsText: {
+		type: 'doc',
+		attrs: {
+			meta: {},
+		},
+		content: [
```
idreyn, line 35: This H1 content is actually the `submissionWorkflow.title` — it should be moved out of here and rendered using an `<h1>` in React in the `InstructionsTab`.

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,43 @@
...
+		text-align: left;
+
+		box-decoration-break: clone;
+	}
+
+	.tabs{
+		color: var(--community-accent-dark);
+		max-width: 900px;
+		> .bp3-tab-list {
+			position: absolute;
```
idreyn, line 23: I'm not sure what any of these properties (`position`, `width`, `height`, `padding-top`) are doing. My hunch is that you don't want any of them.

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,43 @@
...
+			width: 1005.79px;
+			height: 0px;
+			padding-top: 5px;
+			.tab {
+				color: var(--community-accent-dark);
```
idreyn, line 28: The `text-align`, `font-size`, and `font-style` properties here don't need to be specified because they're already either the defaults, or provided by Blueprint.

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,43 @@
+.spub-header-component {
+	border-bottom:	border-bottom: 1px solid #d2d2d2;
+;
+	padding: 200px;
+
+	h1 {
```
idreyn, line 6: We might want to give this a more specific class name, like `submission-workflow-title`. This will differentiate it from any `h1` elements in the user-provided `instructionsText`.

client/containers/Pub/PubSubmissionHeader/PubSubmissionHeader.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const workflow: SubmissionWorkflow = {
+	id: '12089e31h3f23bf-f23f23f23f2-f23f32bu234v23-f2323f',
+	createdAt: '',
+	updatedAt: '',
+	enabled: true,
+	title: 'A Submission Work-work-work-work-work-work-flow',
+	collectionId: '',
+	targetEmailAddress: '',
+	emailText: {
```
idreyn, line 16: If you like, you could call `getEmptyDoc()` to produce these.

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,43 @@
...
+		line-height: 56px;
+		letter-spacing: 0em;
+		text-align: left;
+
+		box-decoration-break: clone;
+	}
+
+	.tabs{
+		color: var(--community-accent-dark);
+		max-width: 900px;
```
idreyn, line 21: What is this value for?

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
```diff
@@ -0,0 +1,43 @@
...
+				font-weight: 500;
+				line-height: 17px;
+				letter-spacing: 0.15em;
+				text-align: left;
+				padding: 15px;
+				text-transform: uppercase;
+			}
+		}
+		.tab-panel {
+			padding-top: 33px;
```
idreyn, line 40: After removing the `position: absolute`, something happens where this looks like it may be too much padding in Storybook.

idreyn (changes requested):
Let's chat about this CSS a bit more over a call...my comments are kind of annoyingly hard to respond to in isolation but I think they point towards a fairly compact solution if I can explain them as a whole.

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
idreyn: Oh, one more. I think we lost the `$header-font` here.

qweliant: How would i use the classname for a DocJSON

client/containers/Pub/PubSubmissionHeader/spubHeader.scss
qweliant: its suppose to change the color of the tabs, but how that is being handled has been updated

Issue #1683, opened 2021-11-30
Add Submission accept/decline controls to `SubmissionPubRow`


idreyn:
The `SubmissionPubRow` component from #1682 needs controls to accept/decline a submission. We've modeled these as thumbs up/down buttons. These should launch a new `Dialog` with the following controls:

- A text area to edit the email that will be sent to the submitter (using the fields #1681)
- A button to commit the acceptance/rejection and send the email
- A button that does the same thing but skips sending the email
- A cancel button

We can use the `DialogLauncher` component to launch these dialogs when the buttons are clicked.

The "thumbs up/down" buttons should not be visible on Pubs that have already been accepted or declined.

PR #1748, opened 2022-01-11
feature: Send emails for Submissions

### LINKED ISSUE #1621: Send emails when Submission is submitted
In response to submission model status change to `submitted`

- `to: <all Pub member emails>`
- `from: noreply-xxx@pubpub.org`
- `reply-to: <collection.submissionWorkflow.email>`
- `cc: <collection.submissionWorkflow.email>`

Include any custom `SubmissionWorkflow` email text in body.

idreyn:
Resolves #1621

This PR causes emails to be sent on the following `Submission.status` state changes:

```
incomplete -> pending
pending -> { accepted, declined }
```

Notes as bullet points because brain hurts

- All of these have specific messages (defined by a `SubmissionEmailKind`) that can be configured in a `SubmissionWorkflow`, and then (in the case of accepted/declined) customized in the UI before sending (via #1683).
- You can pass `skipEmail` to the `/api/submissions` POST handler if you don't want to send an email
- The email will be sent to all the Pub-level Members of the Spub, as well as the `targetEmailAddress` in the workflow
- It will be addressed to the submitted Pub's _earliest created_ Member — this will typically be the person who initiated the Submission, but maybe we should choose the first name in the byline, or do some _et al._ nonsense?)

_Test plan:_

Kind of light for now but you can make sure that emails are at least being generated with my changes to the unit tests:

```
npm run test-dev server/submission/
```


3mcd (approved):
Looks great!

PR #1750, opened 2022-01-12
Bh/submissions accept decline controls

### LINKED ISSUE #1683: Add Submission accept/decline controls to `SubmissionPubRow`

howebenhowe:
Resolves #1683

This PR adds approve/decline/delete controls to submission overview rows and invites editing/preventing emails to be sent to author(s) of submissions.

_Test plan_:

1. create three new submissions, for example for the collection "A Long Collection Name That Everyone Regrets Choosing!"
1. visit that collections Submissions overview
1. mark one of your new, `pending` status submissions as accepted, augment the email copy with some of your own thoughts, then submit
1. mark another one as declined (thumbs-down) and augment the copy
1. delete the third one

Currently, deleted pubs don't disappear until the page is refreshed. Other changes are stored temporarily in UI until a page refresh syncs back/front end anew.

### DIFF
--- a/client/containers/DashboardSubmissions/ArbitrationMenu.tsx
+++ b/client/containers/DashboardSubmissions/ArbitrationMenu.tsx
@@ -1,9 +1,11 @@
 import React from 'react';
 import { Button, Dialog } from '@blueprintjs/core';
 
-import { Submission } from 'types';
+import { Submission, DefinitelyHas, Pub } from 'types';
 import { apiFetch } from 'client/utils/apiFetch';
 import { Icon, IconName, DialogLauncher } from 'client/components';
+
+require('./arbitrationMenu.scss');
 
 const arbitrationOptions = [
 	{
@@ -19,6 +21,7 @@ const arbitrationOptions = [
 	},
 	{
 		icon: 'thumbs-down',
+		targetStatus: 'declined',
 		actionTitle: 'Decline',
 		actionHelpText: 'Would you like to decline this Submission?',
 		onSubmit: (submission: Submission) => () => {
@@ -34,6 +37,7 @@ const arbitrationOptions = [
 	{
 		icon: 'endorsed',
 		actionTitle: 'Accept',
+		targetStatus: 'accepted',
 		actionHelpText: 'Would you like to accept this Submission?',
 		onSubmit: (submission: Submission) => () => {
 			apiFetch('/api/submissions', {
@@ -48,7 +52,7 @@ const arbitrationOptions = [
 ];
 
 type Props = {
-	pub: any;
+	pub: DefinitelyHas<Pub, 'submission'>;
 };
 
 type DialogProps = {
@@ -77,45 +81,35 @@ const VerdictDialog = (props: DialogProps) => {
 	);
 };
 
-const ArbitrationMenu = (props: Props) => {
-	const { pub } = props;
-	return (
-		<div
-			style={{
-				display: 'grid',
-				gridTemplateColumns: 'repeat(3, 1fr)',
-				gridTemplateRows: '1fr',
-				gridColumnGap: '40px',
-			}}
-		>
-			{arbitrationOptions.map((option, index) => {
-				return (
-					<div style={{ gridColumn: index + 1 }} key={option.actionTitle}>
-						<DialogLauncher
-							renderLauncherElement={({ openDialog }) => (
-								<Button
-									minimal
-									small
-									icon={<Icon icon={option.icon as IconName} iconSize={20} />}
-									onClick={openDialog}
-								/>
-							)}
-						>
-							{({ isOpen, onClose }) => (
-								<VerdictDialog
-									handleSubmission={option.onSubmit(pub.submission)}
-									actionTitle={option.actionTitle}
-									isOpen={isOpen}
-									onClose={onClose}
-									actionHelpText={option.actionHelpText}
-								/>
-							)}
-						</DialogLauncher>
-					</div>
-				);
-			})}
-		</div>
-	);
-};
+const ArbitrationMenu = (props: Props) => (
+	<div className="arbitration-menu">
+		{arbitrationOptions.map((option, index) => (
+			<div style={{ gridColumn: index + 1 }} key={option.actionTitle}>
+				{props.pub.submission.status !== option.targetStatus && (
+					<DialogLauncher
+						renderLauncherElement={({ openDialog }) => (
+							<Button
+								minimal
+								small
+								icon={<Icon icon={option.icon as IconName} iconSize={20} />}
+								onClick={openDialog}
+							/>
+						)}
+					>
+						{({ isOpen, onClose }) => (
+							<VerdictDialog
+								handleSubmission={option.onSubmit(props.pub.submission)}
+								actionTitle={option.actionTitle}
+								isOpen={isOpen}
+								onClose={onClose}
+								actionHelpText={option.actionHelpText}
+							/>
+						)}
+					</DialogLauncher>
+				)}
+			</div>
+		))}
+	</div>
+);
 
 export default ArbitrationMenu;
--- a/client/containers/DashboardSubmissions/ArbitrationMenu.tsx
+++ b/client/containers/DashboardSubmissions/ArbitrationMenu.tsx
@@ -1,13 +1,13 @@
 import React from 'react';
 import { Button } from '@blueprintjs/core';
 
-import { DefinitelyHas, Pub } from 'types';
+import { DefinitelyHas, Pub, SubmissionWorkflow } from 'types';
 import { Icon, IconName, DialogLauncher } from 'client/components';
 import VerdictDialog from './VerdictDialog';
 
 require('./arbitrationMenu.scss');
 
-const arbitrationOptions = [
+const getArbitrationOptions = (submissionWorkflow: SubmissionWorkflow) => [
 	{
 		icon: 'cross',
 		actionTitle: 'Delete',
@@ -20,6 +20,7 @@ const arbitrationOptions = [
 		completedName: 'declined',
 		apiMethod: 'PUT',
 		status: 'declined',
+		initialEmailText: submissionWorkflow.declinedText,
 	},
 	{
 		icon: 'endorsed',
@@ -27,6 +28,7 @@ const arbitrationOptions = [
 		completedName: 'accepted',
 		apiMethod: 'PUT',
 		status: 'accepted',
+		initialEmailText: submissionWorkflow.acceptedText,
 	},
 ];
 
@@ -37,7 +39,7 @@ type Props = {
 const ArbitrationMenu = (props: Props) => {
 	return (
 		<div className="arbitration-menu">
-			{arbitrationOptions.map((option, index) => (
+			{getArbitrationOptions(props.pub.submission.submissionWorkflow).map((option, index) => (
 				<div style={{ gridColumn: index + 1 }} key={option.actionTitle}>
 					{props.pub.submission.status !== option.status && (
 						<DialogLauncher
--- a/client/containers/DashboardSubmissions/ArbitrationMenu.tsx
+++ b/client/containers/DashboardSubmissions/ArbitrationMenu.tsx
@@ -34,38 +34,38 @@ const getArbitrationOptions = (submissionWorkflow: SubmissionWorkflow) => [
 
 type Props = {
 	pub: DefinitelyHas<Pub, 'submission'>;
+	onJudgePub: (pubId: string, status?: string) => void;
 };
 
-const ArbitrationMenu = (props: Props) => {
-	return (
-		<div className="arbitration-menu">
-			{getArbitrationOptions(props.pub.submission.submissionWorkflow).map((option, index) => (
-				<div style={{ gridColumn: index + 1 }} key={option.actionTitle}>
-					{props.pub.submission.status !== option.status && (
-						<DialogLauncher
-							renderLauncherElement={({ openDialog }) => (
-								<Button
-									minimal
-									small
-									icon={<Icon icon={option.icon as IconName} iconSize={20} />}
-									onClick={openDialog}
-								/>
-							)}
-						>
-							{({ isOpen, onClose }) => (
-								<VerdictDialog
-									isOpen={isOpen}
-									onClose={onClose}
-									{...option}
-									pub={props.pub}
-								/>
-							)}
-						</DialogLauncher>
-					)}
-				</div>
-			))}
-		</div>
-	);
-};
+const ArbitrationMenu = (props: Props) => (
+	<div className="arbitration-menu">
+		{getArbitrationOptions(props.pub.submission.submissionWorkflow).map((option, index) => (
+			<div style={{ gridColumn: index + 1 }} key={option.actionTitle}>
+				{props.pub.submission.status !== option.status && (
+					<DialogLauncher
+						renderLauncherElement={({ openDialog }) => (
+							<Button
+								minimal
+								small
+								icon={<Icon icon={option.icon as IconName} iconSize={20} />}
+								onClick={openDialog}
+							/>
+						)}
+					>
+						{({ isOpen, onClose }) => (
+							<VerdictDialog
+								isOpen={isOpen}
+								onClose={onClose}
+								{...option}
+								pub={props.pub}
+								onJudgePub={props.onJudgePub}
+							/>
+						)}
+					</DialogLauncher>
+				)}
+			</div>
+		))}
+	</div>
+);
 
 export default ArbitrationMenu;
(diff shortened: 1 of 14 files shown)


client/containers/DashboardSubmissions/VerdictDialog.tsx
```diff
@@ -0,0 +1,154 @@
...
+	initialEmailText?: DocJson;
+};
+
+const PreSubmissionBody = (props: PreSubmissionBodyProps) => {
+	const [shouldSendEmail, setShouldSendEmail] = useState(!!props.initialEmailText);
+	const [customEmailText, setCustomEmailText] = useState(props.initialEmailText || getEmptyDoc());
+	return (
+		<>
+			<div className={Classes.DIALOG_BODY}>
+				<p>Would you like to {props.actionTitle} this submission?</p>
```
idreyn, line 38: `props.actionTitle.toLowerCase()` so the action doesn't appear capitalized here.

client/containers/DashboardSubmissions/arbitrationMenu.scss
```diff
@@ -0,0 +1,6 @@
+.arbitration-menu {
```
idreyn, line 1: This should be `arbitration-menu-component`

client/containers/DashboardSubmissions/verdictDialog.scss
```diff
@@ -0,0 +1,24 @@
+.verdict-dialog {
```
idreyn, line 1: This should be `verdict-dialog-component`

client/containers/DashboardSubmissions/arbitrationMenu.scss
```diff
@@ -0,0 +1,6 @@
...
+	display: grid;
```
idreyn, line 2: I wonder whether a:

```css
display: flex;
> *:not(:last-child) {
    margin-right: 40px;
}
```
might do the job better here. That way you don't have to specify the `gridColumn` in React, and it scales to support more (or fewer) buttons, which eliminates this kind of goofy state:

image

client/containers/DashboardSubmissions/VerdictDialog.tsx
```diff
@@ -0,0 +1,154 @@
...
+			{submissionError ? (
+				<Callout intent="warning" title="There was an error creating this submission." />
+			) : updatedSubmission ? (
+				<PostSubmitBody
+					isHandlingSubmission={isHandlingSubmission}
+					completedName={props.completedName}
+					onClose={props.onClose}
+				/>
+			) : (
+				<PreSubmissionBody
```
idreyn, line 142: Nit: if this is called `PreSubmissionBody`, I think the other component should be called `PostSubmissionBody`. I also wonder if a better noun than "submission" could be used here to refer to the act of submitting the form — it's uniquely ambiguous here because we're dealing with `Submission` objects as well :)

client/containers/DashboardSubmissions/VerdictDialog.tsx
```diff
@@ -0,0 +1,154 @@
...
+			lazy={true}
+			title={props.actionTitle}
+			className="verdict-dialog"
+			isOpen={props.isOpen}
+			onClose={props.onClose}
+		>
+			{submissionError ? (
+				<Callout intent="warning" title="There was an error creating this submission." />
+			) : updatedSubmission ? (
+				<PostSubmitBody
```
idreyn, line 136: I can't get this to ever show up in testing. Under what circumstances does it appear?

client/containers/DashboardSubmissions/VerdictDialog.tsx
```diff
@@ -0,0 +1,154 @@
...
+	};
+	return (
+		<Dialog
+			lazy={true}
+			title={props.actionTitle}
+			className="verdict-dialog"
+			isOpen={props.isOpen}
+			onClose={props.onClose}
+		>
+			{submissionError ? (
```
idreyn, line 133: Would you consider a `renderInner()` function with a set of `if` statements? The nested ternaries are a little hard to follow here.

client/containers/DashboardSubmissions/VerdictDialog.tsx
```diff
@@ -0,0 +1,154 @@
...
+	);
+};
+
+const VerdictDialog = (props: Props) => {
+	const [isHandlingSubmission, setIsHandlingSubmission] = useState(false);
+	const [updatedSubmission, setUpdatedSubmission] = useState(null);
+	const [submissionError, setSubmissionError] = useState(null);
+	const handleSubmission = (customEmailText: DocJson, shouldSendEmail: boolean) => {
+		setIsHandlingSubmission(true);
+		apiFetch('/api/submissions', {
```
idreyn, line 106: In the event you get rid of the `delete` option, you can use the `apiFetch.put` method (and then you don't have to stringify the body).

client/containers/DashboardSubmissions/verdictDialog.scss
```diff
@@ -0,0 +1,24 @@
...
+	.email-text-header {
+		text-transform: uppercase;
+		font-size: 12px;
+		color: #888;
+		margin-bottom: 5px;
+	}
+	.minimal-editor-component {
+			height: 150px;
+			overflow: auto;
```
idreyn, line 10: Do you happen to remember what problems the `overflow: auto`, `min-height: 100%`, and `white-space: break-spaces` solve? I have been trying for years to get this component to Do The Right Thing by default, and if you're having CSS frustrations around it that may be a sign that I should tweak something internally.

client/containers/DashboardSubmissions/VerdictDialog.tsx
```diff
@@ -0,0 +1,154 @@
...
+type Props = {
+	isOpen: boolean;
+	onClose: () => unknown;
+	actionTitle: string;
+	completedName: string;
+	apiMethod: string;
+	status?: string;
+	initialEmailText?: DocJson;
+	pub: DefinitelyHas<Pub, 'submission'>;
+	onJudgePub: (pubId: string, status?: string) => void;
```
idreyn, line 21: `status` can be narrowed to `SubmissionStatus`

client/containers/DashboardSubmissions/VerdictDialog.tsx
```diff
@@ -0,0 +1,154 @@
...
+	const [customEmailText, setCustomEmailText] = useState(props.initialEmailText || getEmptyDoc());
+	return (
+		<>
+			<div className={Classes.DIALOG_BODY}>
+				<p>Would you like to {props.actionTitle} this submission?</p>
+				<p className="email-text-header">
+					<Icon icon="manually-entered-data" iconSize={12} />
+					{'  '}Email to Authors
+				</p>
+				<WorkflowTextEditor
```
idreyn, line 43: This doesn't make sense for the delete case (we won't ever send an email when the submission is deleted)

client/containers/DashboardSubmissions/SubmissionItems.tsx
```diff
@@ -68,39 +85,31 @@ const SubmissionItems = (props: Props) => {
 		},
 	});
 
+	const [localStats, setLocalStats]: [{ [pubId: string]: SubmissionStatus }, any] = useState({});
```
idreyn, line 88: Slightly cleaner way to do this:

```tsx
const [localStats, setLocalStats] = useState<{ [pubId: string: SubmissionStatus }>({});
```
this will properly type both items in the `[value, setter]` array.

idreyn: Could we lengthen these names to `localStatuses`? "stats" makes me think it's "statistics"

client/containers/DashboardSubmissions/SubmissionItems.tsx
```diff
@@ -68,39 +85,31 @@ const SubmissionItems = (props: Props) => {
...
-	const renderPubs = () => {
-		return pubs.map((pub) => (
-			<PubOverviewRow
-				pub={pub}
-				key={pub.id}
-				leftIconElement="manually-entered-data"
-				hasSubmission={true}
-			/>
-		));
+	const trackLocalStatus = (pubId: string, status?: string) => {
```
idreyn, line 96: Perhaps `updateLocalStatus` would be a better name — I don't know whether "track" is a read-ish or write-ish operation from a glance. Also, this can be wrapped in a `useCallback()` and `status` could be narrowed to `SubmissionStatus`.

client/containers/DashboardSubmissions/SubmissionItems.tsx
```diff
@@ -112,9 +121,30 @@ const SubmissionItems = (props: Props) => {
 				filters={overviewSearchFilters}
 			/>
 			<OverviewRows>
-				{renderPubs()}
+				{(pubs as DefinitelyHas<Pub, 'submission'>[])
```
idreyn, line 124: Could we make `DefinitelyHas<Pub, 'submission'>[]` the type of the `pubs` prop?

idreyn (changes requested):
Looking well on its way! The `VerdictDialog` is feeling really nice to use. I'm primarily concerned with this "delete submission" button — I don't think it needs to exist, and it simplifies a lot of the code if it's removed.

howebenhowe:
@idreyn re. deleting pubs, i was operating from a combination of the card and the [final figma view](https://www.figma.com/file/ERHvqJqCVBZ7dH2ZQBGlp8/SPUB-vision?node-id=223%3A1942), hence the delete functionality. Happy to eliminate it--we sure it's not a concern that the `declined` filter view will get clogged with accidental/spam submissions?

client/containers/DashboardSubmissions/SubmissionItems.tsx
howebenhowe: don't think so? We're pulling it out of both `useManyPubs` and `props.initialPubs`. I'm down to put it to `props.initialPubs`, but I don't understand how we could expect presence of submissions on pubs returned from `usemanyPubs`

howebenhowe: nevermind, TS ain't complaining so neither am I.

idreyn:
> @idreyn re. deleting pubs, i was operating from a combination of the card and the [final figma view](https://www.figma.com/file/ERHvqJqCVBZ7dH2ZQBGlp8/SPUB-vision?node-id=223%3A1942), hence the delete functionality. Happy to eliminate it--we sure it's not a concern that the `declined` filter view will get clogged with accidental/spam submissions?

Yes, I think this is just a miscommunication in design/dev handoff. I don't know what @djagdish's intention for the delete button was, but my guess is that it should delete the Pub itself and not just the Submission object. It would be annoying for the person managing the Submissions queue to find a Pub that was so odious that they press the "delete" button rather than just declining it, and then later find out that that the Pub is still bumping around their Community. At any rate, I apologize for not accounting for it in #1683.

client/containers/DashboardSubmissions/VerdictDialog.tsx
howebenhowe: Good catch. Got some logic conflict here. You can get it to appear by starting in the 'all section' and then either approving or declining a submission. The conflict at play is that after changing a status, rather than letting the row languish in a filter view that ought to exclude it (e.g. a freshly 'declined' submission in the 'pending' view) it is omitted from the set of rows displayed. Since `VerdictDialog` is a child of the row itself, the success message is not dropped out of the dom together with it's parent row.

client/containers/DashboardSubmissions/VerdictDialog.tsx
howebenhowe: I am reluctant to break from chained ternary, but I think it's just a matter of taste. Using a `renderInner()` would mean either (a) passing extraneous props into it only to discard them if the necessary component doesn't need them or (b) priming a few components somewhere inside the `VerdictDialog` component before the return value. Those each smell funkier to me.

Here's what converted me a year or so ago:
https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340

Overall, I'm still on the lookout for the best approach to multiple-choice conditional rendering.

client/containers/DashboardSubmissions/verdictDialog.scss
howebenhowe: `overflow: auto` keeps the formatting bar in view once the email copy runs long

`white-space: break-space` covers for when the placeholder text is longer than the input is wide. This is nearly impossible, and can be removed.

Can't recall why I included `min-height: 100%`, so I'll remove that too.

client/containers/DashboardSubmissions/arbitrationMenu.scss
howebenhowe: I'm happy either way. In the `all` filter view with `grid` the delete / decline / approve buttons stay vertically aligned by effect tolerant of the missing buttons, but at the cost of the goofy state you show above. Comparison, from the figma mock ups:
Screenshot from 2022-01-24 01-55-12

idreyn: That's a great point! In that case, the "delete" button should definitely go at the end, though, since it's always visible. I think it should also use the `trash` icon, because the `x` is mostly used to mean "Remove from Collection" elsewhere and folks may assume that's the action here, too.

client/containers/DashboardSubmissions/VerdictDialog.tsx
idreyn: Fine by me if you feel strongly about it, though I think the readability argument breaks down when the expressions don't mostly fit onto a single line like this:

```ts
const value =  (!conditionA)
    ? valueC
    : (conditionB)
    ? valueA
    : valueB;
```


 Neither (a) or (b) is necessarily what I'm imagining though. This is a good example of what I mean — the inner React elements are being constructed within the `return` statement, just in an inner function definition.

https://github.com/pubpub/pubpub/blob/a8b1c090a8c523c853d31af520657b630d510b76/client/components/PubShareDialog/PubShareDialog.tsx#L159

Also on (b) — at some point in the last two years @3mcd kind of convinced me that defining React nodes outside of a `return` statement is a totally fine thing to do:

https://github.com/pubpub/pubpub/blob/2965db0ed81ce2f24dd41b59f02cfc0521796edc/client/utils/activity/renderers/itemRenderer.tsx#L28

Just so you know where I am on the carousel of JS code fashions.

client/containers/DashboardSubmissions/ArbitrationMenu.tsx
```diff
@@ -0,0 +1,106 @@
...
+						completedName="deleted"
+						onSubmit={() =>
+							apiFetch('/api/pub', {
+								method: 'DELETE',
+								body: JSON.stringify({
+									id: props.pub.submission.pubId,
+								}),
+							}).then(() => props.pub.submission)
+						}
+						pub={props.pub as DefinitelyHas<Pub, 'submission'>}
```
idreyn, line 44: This cast should no longer be necessary, right?

client/containers/DashboardSubmissions/ArbitrationMenu.tsx
```diff
@@ -0,0 +1,106 @@
...
+					<Button
+						minimal
+						small
+						icon={<Icon icon={'cross' as IconName} iconSize={20} />}
+						onClick={openDialog}
+					/>
+				)}
+			>
+				{({ isOpen, onClose }) => (
+					<VerdictDialog
```
idreyn, line 30: You might find a `ConfirmDialog` helpful here if you don't want to squeeze this functionality into the `VerdictDialog`. Either way.

client/containers/DashboardSubmissions/ArbitrationMenu.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const ArbitrationMenu = (props: Props) => (
+	<div className="arbitration-menu-component">
+		<div style={{ gridColumn: 1 }}>
+			<DialogLauncher
+				renderLauncherElement={({ openDialog }) => (
+					<Button
+						minimal
+						small
+						icon={<Icon icon={'cross' as IconName} iconSize={20} />}
```
idreyn, line 24: Does this work without the cast?

client/containers/DashboardSubmissions/ArbitrationMenu.tsx
howebenhowe: verily

client/containers/DashboardSubmissions/ArbitrationMenu.tsx
howebenhowe: Yup.

PR #1754, opened 2022-01-18
fix: allow DOI updates for Pubs without releases

### LINKED ISSUE #1745: DOI suffix generation

1. Working on a draft for HDSR and need to generate a DOI suffix (not trying to deposit it)
2. When you click on "generate suffix" a message pops up that says "there was a problem generating the doi suffix"

...

DOI suffix generates

https://hdsr.mitpress.mit.edu/dash/pub/kpn45eyx/settings

Chrome

===

Screen Shot 2022-01-10 at 3 24 37 PM

Rebecca at HDSR related to CA

3mcd:
Fixes #1745

This PR moves the releases check in our DOI generation's `assertUserAuthorized` function into a new function called `pubExistsAndIsMissingReleases`. Removing this check allows users to generate DOIs for Pubs without releases.

The new `pubExistsAndIsMissingReleases` function is called in the Crossref preview/deposit code because we still want to guarantee that a Pub has at least one release before depositing.

### Test Steps

* Create a new Pub. Don't release it.
* Go into the Pub settings, you should be able to generate a DOI without error!

### DIFF
--- a/server/doi/api.ts
+++ b/server/doi/api.ts
@@ -1,12 +1,10 @@
-import xmlbuilder from 'xmlbuilder';
-
+import { Release } from 'server/models';
 import app, { wrap } from 'server/server';
 import { ForbiddenError } from 'server/utils/errors';
-import { Release } from 'server/models';
 import { parentToSupplementNeedsDoiError } from 'utils/crossref/createDeposit';
-
-import { getDoiData, setDoiData, generateDoi } from './queries';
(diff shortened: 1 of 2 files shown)


gabestein (changes requested):
Thanks for this! Updating the suffix manually works great. But clicking 'generate suffix' still surfaces an error. I think most folks just generate suffixes, so it would be nice to be able to do that as well if possible.

PR #1758, opened 2022-01-18
Activity Digest Emails

3mcd:
This PR implements all of @howebenhowe's fantastic work, plus the digest email tool, plus a feature flag for the subscribe switch. This is an amalgamation of lots of already-reviewed work (other than the feature flag, but that's trivial), so no QA required.

After this is merged into master, I'll hook up the scheduler plugin and kick off a test run :)


idreyn (approved):
LGTM!

PR #1764, opened 2022-01-24
fix: Make Editor read-only until Firebase finishes connecting

### LINKED ISSUE #1753: Refreshing the page can cause lost work
1. The editor had an issue connecting; a small yellow "Connecting..." message was displayed where there is usually green "Connected"
2. I assumed this was an innocuous bug, and no problem. I worked in an open tab for several days
3. While editing I pressed Ctrl+R with the intention of "redo"-ing a change I had "undo"-ed with Ctrl-Z, causing the page to refresh.
4. My work was lost
...

Ideally in a situation like this, the contents of the edit buffer should have been cached locally and taken precedence over the copy stored on the server so that my work would not be lost.

===

idreyn:
Resolves #1753

We've had a few reports of folks losing their work because `Editor` lets them add content in a stalled `connecting` state. This change makes the editor read-only until `status = 'connected'`. I've done a little work moving the responsibility for monitoring the connection around, and a little type-safety/cosmetic overhaul.

_Test plan:_

Check that:
- The editor works and persists your changes after refreshing
- You cannot type while the status bar says "Connecting..." (maybe try throttling your connection here)
- The collab-status indicator works as expected: `connecting -> connected -> (saving <-> saved) -> disconnected`

### DIFF
--- a/client/components/Editor/plugins/collaborative/document.ts
+++ b/client/components/Editor/plugins/collaborative/document.ts
@@ -3,7 +3,11 @@ import { receiveTransaction, sendableSteps } from 'prosemirror-collab';
 import { Step } from 'prosemirror-transform';
 import { uncompressStepJSON } from 'prosemirror-compress-pubpub';
 
-import { storeCheckpoint, createFirebaseChange } from '../../utils';
+import {
+	storeCheckpoint,
+	createFirebaseChange,
+	getFirebaseConnectionMonitorRef,
+} from '../../utils';
 
 /*
 Rough pipeline:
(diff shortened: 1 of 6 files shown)


howebenhowe:
I _think_ this is failing the testing instructions. When toggled to slow 3G speed, I was able to type in the editor while the indicator status still read `connecting` and then those edits were erased once the editor hopped to `connected`.
See screenshots:
Screenshot from 2022-01-25 11-37-21
Screenshot from 2022-01-25 11-39-24

howebenhowe (approved):
The code looks clean but my grasp on firebase is nascent at best. I have been lucky thus far. It does mean a firebase-savvy individual might want to pass eyes over this one.

gabestein:
Was closing out some tickets and found the one that spurned this PR. So, bumping it to see if we can get it across the finish line.

3mcd (approved):

idreyn:
Good catch @howebenhowe, I fixed it now!

PR #1765, opened 2022-01-26
update: Some glue for Submissions

idreyn:
This PR chamfers some edges on the Submissions dashboard that's coming together across tasks.

- Move a bunch of state management for individual Spubs out of `PubRow` and `SubmissionItems` into a new `SubmissionRow` that wraps `PubRow`. This component is responsible for managing its own `submission.status` state, and hiding itself if the Pub has been deleted. It also renders Submission-specific details (like the status) instead of delegating that to the more general `PubRow`.
- Add a `cacheQueries` option to the `useManyPubs()` monstrosity. By default the hook caches Pubs from different filters/queries to `/api/pubs/many` — these filters correspond to e.g. the Pending and Accepted tabs of the Submissions listing. This is a great experience on the Overview dashboard tab but creates problems in the Submissions tab, where Pubs may move _between_ filters. Setting `cacheQueries: false` forces to hook to re-fetch Pubs on every filter change, which is closer to ideal here.
- Make the `OverviewSearchGroup` filter tabs take an `OverviewSearchFilter` instead of `Partial<PubsQuery>` as its list of filters (so we can reference them by id) and work as either a controlled or uncontrolled component by adding a `filter` prop — together these are helpful for a parent component that needs to know which tab is currently selected.
- Some copy changes

_Test plan:_
- In the Submissions dashboard, move Pubs between the Pending, Accepted, and Rejected tabs. Verify that when you switch tabs a reload of all Pubs in the new tab occurs, and the _updated_ list of Pubs in that tab's submission state is always shown
- Also check that the filter tabs on the Community and Collection overview pages are working

### DIFF
--- a/client/containers/DashboardSubmissions/arbitrationMenu.scss
+++ b/client/containers/DashboardSubmissions/arbitrationMenu.scss
@@ -1,6 +1,9 @@
 .arbitration-menu-component {
-	display: grid;
-	grid-template-columns: repeat(3, 1fr);
-	grid-template-rows: 1fr;
-	grid-column-gap: 40px;
+	display: flex;
+	flex-direction: row;
+	align-items: center;
+
+	> *:not(:first-child) {
+		margin-left: 20px;
+	}
 }
--- a/client/utils/useManyPubs/useManyPubs.ts
+++ b/client/utils/useManyPubs/useManyPubs.ts
@@ -1,10 +1,11 @@
 import { useCallback, useEffect, useState } from 'react';
-import { useUpdateEffect } from 'react-use';
+import { useUpdate, useUpdateEffect } from 'react-use';
 
 import { Pub } from 'types';
 import { usePageContext } from 'utils/hooks';
 import { indexByProperty } from 'utils/arrays';
 import { apiFetch } from 'client/utils/apiFetch';
+import { useLazyRef } from 'client/utils/useLazyRef';
 
 import {
 	getStartLoadingPubsState,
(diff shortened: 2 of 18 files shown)


client/containers/DashboardSubmissions/arbitrationMenu.scss
```diff
@@ -1,6 +1,9 @@
 .arbitration-menu-component {
-	display: grid;
-	grid-template-columns: repeat(3, 1fr);
-	grid-template-rows: 1fr;
-	grid-column-gap: 40px;
+	display: flex;
```
idreyn, line 2: I ended up changing this to get the buttons to align vertically

client/utils/useManyPubs/useManyPubs.ts
```diff
@@ -63,80 +67,92 @@ export const useManyPubs = <P extends Pub = Pub>(
...
+			isLoading: currentQueryState.isLoading,
+			hasLoadedAllPubs: currentQueryState.hasLoadedAllPubs,
+			pubs: currentQueryState.orderedPubs.pubsInOrder as P[],
 		},
 		allQueries: {
-			isLoading: Object.values(manyPubsState).some((s) => s.isLoading),
+			isLoading: Object.values(manyPubsState.current).some((s) => s.isLoading),
 		},
 	};
 };
```
3mcd, line 158: I like these changes. @howebenhowe mentioned implementing something like this a while back and you pulled it off great!

3mcd (approved):
LGTM

client/utils/useManyPubs/useManyPubs.ts
idreyn: It is kind of scary that the custom hooks I write all eventually get complicated enough that I need to use refs as an escape hatch to avoid the stale-closure problem. Oh well!

Issue #1731, opened 2022-01-03
Build the `SubmitPubDialog`


idreyn:
In #1730 we added a "Submit Pub" button to the Spub header. This task adds a modal that is triggered when that button is clicked, which actually submits the Pub.

The modal has two basic states:

- Pre-submit:
  - Confirmation text: "Are you sure you're ready to submit this Pub?" or something
  - Cancel and Submit buttons
  - Clicking Submit calls `/api/submissions` _and_ merges `submission.abstract` into Pub. We could choose to merge the abstract on the client or as a side-effect of POST to `/api/submissions` on the server. Hmm.
- Post-submit:
  - Big green Blueprint `Callout` with `intent="success"`
  - Let the user know that their submission is under review
  - Maybe(?) show the `submissionWorkflow.emailText` — need clarification
  - Button to show user their submission, i.e. refresh the page (lol)

The `DialogLauncher` is a handy component to know about here.

PR #1790, opened 2022-02-07
Add dialog popup for submission to preview tab

howebenhowe:
#1731

WIP, missing (only, I think) the email text display, which may or may not be desirable. If anyone is reading this, please comment below, should we include the 'thanks for submitting' email text in the success `Callout` after a contributor hit's submit?

### DIFF
--- a/client/containers/Pub/SpubHeader/PreviewTab/PreviewTab.tsx
+++ b/client/containers/Pub/SpubHeader/PreviewTab/PreviewTab.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
 import { Icon } from '@blueprintjs/core';
 
-import { PubPageData } from 'types';
-import { GridWrapper, PubByline } from 'components';
+import { PubPageData, Submission, DefinitelyHas } from 'types';
+import { GridWrapper, PubByline, DialogLauncher } from 'components';
 import { usePageContext } from 'utils/hooks';
 
+import SubmitDialog from './SubmitDialog';
 import PubHeaderBackground from '../../PubHeader/PubHeaderBackground';
 import ResponsiveHeaderButton from '../../PubHeader/ResponsiveHeaderButton';
 import { getHistoryButtonLabelForTimestamp } from '../../PubHeader/DraftReleaseButtons';
@@ -14,7 +15,7 @@ require('../../PubHeader/draftReleaseButtons.scss');
 type Props = {
 	historyData: any;
 	updateHistoryData: any;
-	pubData: PubPageData;
+	pubData: PubPageData & { submission: DefinitelyHas<Submission, 'submissionWorkflow'> };
 };
 
 const PreviewTab = (props: Props) => {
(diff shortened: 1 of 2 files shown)


client/containers/Pub/SpubHeader/PreviewTab/SubmitDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+};
+
+const SubmitDialog = (props: Props) => {
+	const { collabData } = usePubContext();
+	const { view } = collabData.editorChangeObject;
+	const [isHandlingSubmission, setIsHandlingSubmission] = useState(false);
+	const [updatedSubmission, setUpdatedSubmission] = useState(null);
+	const [submissionErr, setSubmissionErr] = useState(null);
+	const onSubmit = () => {
+		const abstract = props.submission.abstract || getEmptyDoc();
```
idreyn, line 24: For cleanliness consider breaking the Prosemirror manipulation out into its own local helper function:

```ts
addAbstractToEditor(abstract: DocJson, view: EditorView): void
```

client/containers/Pub/SpubHeader/PreviewTab/SubmitDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+				status: 'pending' as SubmissionStatus,
+			})
+			.then((submissionRes) => setUpdatedSubmission(submissionRes))
+			.catch((err) => setSubmissionErr(err))
+			.finally(() => setIsHandlingSubmission(false));
+	};
+	return (
+		<Dialog isOpen={props.isOpen} onClose={props.onClose}>
+			{submissionErr ? (
+				<Callout intent="warning" title="There was an error updating this submission." />
```
idreyn, line 45: `There was an error submitting this Pub` may more accurately capture this from the user's perspective.

client/containers/Pub/SpubHeader/PreviewTab/SubmitDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+			) : updatedSubmission ? (
+				<>
+					<div className={Classes.DIALOG_BODY}>
+						<Callout intent="success" title="Submitted!">
+							Your pub has been submitted and is under review.
```
idreyn, line 50: I'd suggest `Your Pub has been submitted for review`. The way it's written implies (to me) that someone is currently actively looking at it.

client/containers/Pub/SpubHeader/PreviewTab/PreviewTab.tsx
```diff
@@ -14,14 +16,15 @@ require('../../PubHeader/draftReleaseButtons.scss');
 type Props = {
 	historyData: any;
 	updateHistoryData: any;
-	pubData: PubPageData;
+	pubData: PubPageData & { submission: DefinitelyHas<Submission, 'submissionWorkflow'> };
```
idreyn, line 19: Could we merge this information into `PubPageData`, so that it has:

```ts
submission?: DefinitelyHas<Submission, 'submissionWorkflow'>
```

then the type of `pubData` across the `SpubHeader` and its children would be:

```ts
pubData: DefinitelyHas<PubPageData, 'submission'>;
```

By the way, we have more informative types for `{ historyData, updateHistoryData }` than `any`:

```ts
historyData: PubHistoryState;
updateHistoryData: (patch: Partial<PubHistoryState>) => unknown;
```
I saw you mention elsewhere that using `any` to type these props made you queasy, and it should! Everywhere you see `any` as a prop type it's an artifact of our automatic best-effort JS -> TS conversion from late 2020 — we shouldn't be using it for new props.

idreyn (changes requested):
This looks really good! Just a few requests for minor refactors and copy changes.

PR #1795, opened 2022-02-09
feature(?): Render unstructured citations as mixed-citation in JATS (and use element-citation otherwise)

idreyn:
This PR makes two major changes to our JATS export pipeline. The first is straightforward and the second is stupidly complicated:

1. When exporting with JATS, we turn on the `element_citations` target flag to use the `element-citation` instead of `mixed-citation` JATS element for references — this produces a maximally structured citation.

2. Citations that have unstructured data will be rendered by this pipeline as a `ref` object with a unique ID containing an empty `element-citation` object. We'll transform these to meaningful `mixed-citation` objects by:

- Parsing the post-pandoc JATS XML
- Looking for empty `ref[id=$X] > element-citation` nodes for all Notes with `{ id: $X }` that have HTML unstructured content, but no structured content
- For each of these nodes, replace the empty `element-citation` with a `mixed-citation`. We populate this `mixed-citation` with the JATS equivalent of the unstructured content (generated by additional Pandoc calls)
- Render the updated XML back to a string

Note that we are (still) dropping the unstructured content of citations that _also_ have structured content.

_Test plan:_
I haven't actually done much testing yet, but broadly:
- Non-JATS export formats should see no behavior changes
- Structured citations in JATS should be rendered as `element-citation` instead of `mixed-citation`
- Unstructured citations should be rendered as `mixed-citation` and should contain rich text content.

Thanks to @gabestein and @3mcd for many hours of debugging towards a solution to this.


workers/tasks/export/notes.ts
```diff
@@ -30,51 +34,22 @@ const enrichNoteWithStructuredResult = (
...
-	}
-	return null;
-};
-
 const getCslJsonForNote = (
 	note: Note,
 	hash: string,
 	renderedStructuredValues: RenderedStructuredValues,
-) => {
+): [boolean, Record<string, any>] => {
```
3mcd, line 41: Could we maybe type this tuple or name its elements? I guess it's kind of obvious by looking at the function implementation but it wouldn't be so clear otherwise.

3mcd (approved):
LGTM!

Issue #1684, opened 2021-11-30
Add Release modal for Pubs in the "accepted" state in the Submissions tab


idreyn:
Depends on #1682 

This is possibly below the line for MVP but just to spec it out: the `PubReleaseDialog` is pretty much ready to be added to the dashboard (and is a very useful thing to have there!) We just need to make some changes to its type signature:

- It should accept a `Pub` instead of a `PubPageData`
- It should accept an `historyKey?: number` rather than a `historyData: { latestKey?: number}`. The Release API should handle null history keys just fine and default to releasing the latest draft of the Pub

Just like in #1683 we can use the `DialogLauncher` component to launch this modal from a `SubmissionPubRow`.

PR #1796, opened 2022-02-09
feature: pub release dialog

howebenhowe:
#1684

Adds a 'release pub' button to the Submission Overview Rows for `accepted` spubs.

_Test plan:_

1. navigate to Submissions Overview
2. filter pubs based on `accepted`  status
3. click "social-media" icon for releasing pubs
4. proceed through pub-release process, same as from a pub draft

If it differs in any way from the release experience in the pub draft view, that's a problem.

### DIFF
--- a/client/components/PubReleaseDialog/PubReleaseDialog.tsx
+++ b/client/components/PubReleaseDialog/PubReleaseDialog.tsx
@@ -17,16 +17,17 @@ import { usePageContext } from 'utils/hooks';
 
 import { apiFetch } from 'client/utils/apiFetch';
 import { ClickToCopyButton, MinimalEditor } from 'components';
-import { Release, PubPageData } from 'types';
+import { Release, PubPageData, Pub } from 'types';
 
 require('./pubReleaseDialog.scss');
 
 type Props = {
-	historyData: {
+	historyKey?: number;
+	historyData?: {
 		latestKey?: number;
 	};
 	isOpen: boolean;
-	pubData: PubPageData;
+	pubData: PubPageData | Pub;
 	onClose: () => unknown;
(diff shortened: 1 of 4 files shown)


client/components/PubReleaseDialog/PubReleaseDialog.tsx
```diff
@@ -100,7 +98,7 @@ const PubReleaseDialog = (props: Props) => {
 			<div>
 				<p className="text-info">{label}</p>
 				<ControlGroup className="url-select">
-					<InputGroup className="display-url" value={url} fill small />
+					<InputGroup className="display-url" value={url} fill small disabled />
```
idreyn, line 101: Thanks for the attention to detail here. An alternative fix that doesn't need any CSS changes is:
```tsx
onChange={() => {}}
```
presented for your consideration.

client/containers/DashboardSubmissions/ArbitrationMenu.tsx
```diff
@@ -16,6 +16,29 @@ type Props = {
 
 const ArbitrationMenu = (props: Props) => (
 	<div className="arbitration-menu-component">
+		{props.pub.submission.status === 'accepted' && (
```
idreyn, line 19: In the `PubReleaseDialog`'s current home on the Pub page, the button that opens it is disabled if there aren't any changes in the Pub draft since the last Release. We don't have any such check here (and it would be troublesome to pipe that data in). So you can use this new button to try creating a new, duplicate Release, but it will give the user an unhelpful error once it's been attempted.

To get around this for now, I propose only showing this button if the Pub has no Releases.

idreyn (commented):
LGTM! But please see my comment about hiding the button for subsequent Releases.

PR #1803, opened 2022-02-14
Two citation fixes

idreyn:
This PR fixes two problems with citations:

- Citations with identical `unstructuredValue` are no longer deduped because the `ids` plugin is now generating random identifiers for each paragraph in the unstructured content HTML. To restore this behavior, I have added the [`striptags`](https://www.npmjs.com/package/striptags) package to get raw text from this HTML. We now dedupe on the basis of that raw text. This will restore the correct behavior retroactively to Pubs created since this breaking change.
- The citation `customLabel` attribute doesn't get preserved in copy+paste. Well, it does now!

_Test plan:_
- Create two separate citations with the same unstructured content and verify that they are deduped in the Pub bottom citations section. Refresh the page and make sure this works on the server during initial SSR.
- Create a citation with a custom label (using the control in the lower right side of the citation editor) and verify that this custom label survives a copy+paste round trip.

### DIFF
--- a/client/components/Editor/schemas/citation.ts
+++ b/client/components/Editor/schemas/citation.ts
@@ -32,12 +32,13 @@ export default {
 						id: node.getAttribute('id'),
 						value: node.getAttribute('data-value') || '',
 						unstructuredValue: node.getAttribute('data-unstructured-value') || '',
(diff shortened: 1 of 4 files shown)


howebenhowe (approved):
Code looks good, tests are clean. SHIPPPIT!

PR #1804, opened 2022-02-15
fix: Activity digest emails not sending due to deleted associations

idreyn:
This PR fixes a bug that would crash the Activity digest sending script when it encountered an object in `ActivityAssociations` that was deleted after its Activity items were created. I also added a try/catch wrapper around the code that renders and sends an individual email, so if one fails the script will plow on undeterred.

_Test plan:_
Let's land this and see whether it works tonight.


howebenhowe (approved):
Looks good! I like the swap to payload from associations--more to the point. Good call on the comments for documentation, too.

PR #1805, opened 2022-02-16
Bh/light dark theme spubheader

howebenhowe:
Quick fix changes to make the recently included Submit Pub button match the rest of the header buttons through all light/dark theme selections.

_Test plan_

1. navigate to an incomplete pub on this PR's dyno
2. go to the preview tab
3. confirm the 'submit pub' button matches the styling of the other header buttons
4. in duqduq, find the same pub and change the light/dark text background in the theme
5. go back to this pr's dyno, and confirm the theming has changed

this will be easier to test after getting the 'edit theme' functionality added


client/containers/Pub/SpubHeader/PreviewTab/PreviewTab.tsx
```diff
@@ -38,7 +38,7 @@ const PreviewTab = (props: Props) => {
 							<span className="text-wrapper">{pubData.title}</span>
 						</h1>
 						<h3 className="description pub-header-themed-secondary">
-							{pubData.description}
+							<span className="description-text-content">{pubData.description}</span>
```
idreyn, line 41: I think we can just use the existing `text-wrapper` class here.

client/containers/Pub/SpubHeader/PreviewTab/PreviewTab.tsx
```diff
@@ -38,7 +38,7 @@ const PreviewTab = (props: Props) => {
 							<span className="text-wrapper">{pubData.title}</span>
 						</h1>
 						<h3 className="description pub-header-themed-secondary">
```
idreyn, line 40: btw I think we should write `{pubData.description && <h3> ...` so that this box doesn't appear when there is no description.

idreyn (changes requested):
Tested this out and it looks great — requesting changes to push for a bit of CSS simplification...it's a small thing, but really important to keep the surface area of the header theming no more complicated than it needs to be, since it's already really complicated.

idreyn (approved):
LGTM!

PR #1808, opened 2022-02-17
Submission Contributors

qweliant:
This adds contributor functionality to the submission tab in the spub header

image

test
Run local environment and add contributors. refresh page to be sure contributors are still saved


idreyn (approved):
LGTM!

PR #1809, opened 2022-02-17
Spub Settings

qweliant:
This branch adds functionality and styling  for the submission settings tab

image

image

After changing pub theme to community color
image

test plan:
In your local dev edit a spubs settings and make sure the changes persist on refresh


client/containers/Pub/SpubHeader/spubHeader.scss
```diff
@@ -111,3 +111,15 @@ $title-elements-spacing: 15px;
 		max-width: $spub-configuration-width;
 	}
 }
+
+.small-header-button-component {
```
idreyn, line 115: Don't forget that these need to be scoped to the `.spub-header-component` selector or they're going to apply everywhere in the application!

idreyn (changes requested):
Just a few CSS things but otherwise looks great!

idreyn (changes requested):
Looking good, but we do need to scope the `.small-header-button-component` and `.file-upload-button-component` selectors to a parent component, or they'll apply globally instead of just inside the spub header, as desired.

idreyn (approved):

PR #1832, opened 2022-02-22
fix: Use regex instead of XML parser to insert mixed-citation

idreyn:
Our attempt to insert JATS `mixed-citation` elements using an XML parse/write roundtrip (in #1795) stripped too much important info from the XML. So in this PR I use regex to accomplish the same thing 🙃 I also limited the tags that can go inside a `mixed-citation` so that the JATS will validate.

_Test plan:_
Create a document with multiple unstructured citations and at least one structured one (`/pub/7w7jd9y9/draft` will do nicely) and check the JATS export:
- `element-citation` elements should remain for structured citations
- `mixed-citation` elements should appear for unstructured citations
- The only rich text content allowed inside `mixed-citation` should be `bold`, `italic`, and `ext-link`. `p` tags should be stripped.


3mcd (approved):
LGTM!

PR #1837, opened 2022-02-24
feat: add digest links; fix discussion overflow

### LINKED ISSUE #1793: Add Pub links to activity digest
- Make Pub title in digest email a link to the Pub slug link
- Resolve discussion text box overflow

3mcd:
Resolves #1793

This PR
* adds links to Pub and Collection activity item titles in digest emails
* uses `colspan` to prevent discussion snippets from overflowing

*Test Steps*

* Visit http://localhost:9876/email/digest
* Verify that Pub and Collection titles are clickable and take you to the correct URL
* Verify that discussion snippets are contained within the width of the email (i.e. they do not overflow to the right).

### DIFF
--- a/server/utils/email/digest.tsx
+++ b/server/utils/email/digest.tsx
@@ -11,6 +11,7 @@ import { ActivityAssociations, ActivityItem, Community, Scope, User } from 'type
 import { Digest } from 'components/Email';
 import { globals, reset } from 'components/Email/styles';
 import { fetchActivityItems } from 'server/activityItem/fetch';
+import { collectionUrl, pubUrl } from 'utils/canonicalUrls';
 
 type KeyedActivityItem = ActivityItem & {
 	displayKey: string;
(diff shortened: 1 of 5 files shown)


server/utils/email/digest.tsx
```diff
@@ -44,6 +45,17 @@ const getAffectedObject = (item: ActivityItem, associations: ActivityAssociation
 	};
 };
 
+const getAffectedObjectUrl = (item: ActivityItem, associations: ActivityAssociations) => {
+	const community = Object.values(associations.community)[0];
+	if (item.pubId) {
+		return pubUrl(community, associations.pub[item.pubId]);
```
idreyn, line 51: In the case where a Pub was deleted , `associations.pub[item.pubId]` will be undefined — so we gotta check for the existence of that too before calling `pubUrl()`.

server/utils/email/digest.tsx
```diff
@@ -44,6 +45,17 @@ const getAffectedObject = (item: ActivityItem, associations: ActivityAssociation
...
+	}
+	if (item.collectionId) {
+		return collectionUrl(community, associations.collection[item.collectionId]);
```
idreyn, line 54: Here too.

server/utils/email/digest.tsx
```diff
@@ -44,6 +45,17 @@ const getAffectedObject = (item: ActivityItem, associations: ActivityAssociation
...
+	}
+	return undefined;
```
idreyn, line 56: $0.02 — I think we should try to stick to `null` for a falsy fallback value, so `undefined` more strongly carries its built-in meanings of "this key doesn't exist" or "no return value".

idreyn (changes requested):
Tested and works great, but see my comments about the `associations` object!

idreyn (approved):
LGTM, thanks!

Issue #1807, opened 2022-02-17
Credit collection-level editor/creators in the auto-generated pub-level citations


avanouse:
Enhancement requests are requests to make changes to an existing PubPub feature. If you'd like to suggest a new feature, please post in the ideas section of our [https://github.com/pubpub/pubpub/discussions](discussion forum).

_Before filing your enhancement, please search existing issues to make sure a similar one doesn't already exist. If one does, please leave a comment or reaction in support of the issue rather than creating a new one._

_To fill out the request, please leave the below headers (**) in tact, and fill out each section. If a section isn't applicable, leave it blank or write n/a under it. Please do not delete any template sections._

### What's your enhancement request?
Let the automatic citation for a pub include collection-level metadata for editorship and/or authorship so that (e.g.) the editor of a collection of essays is always cited in the generated citation for the chapter *as well as* the author of the  pub (which is currently visible in this citation only when you've checked the box for it to appear on the byline).

### What would this enhancement help you accomplish?
(1) more consistently proper attribution for cases where the collection-level creator is also the primary author of the pub (authors get cited for citations chapter-by-chapter: can this  be true on clicking "cite" as well as when viewing the heading of the pub?); (2) editors can be properly credited (as editors) for the contributions they've solicited, selected, and helped through gestation

---

### What's the URL to your PubPub community?
[examples abound! Let's please talk about this!](https://autogeddon.pubpub.org/pub/jzu2d7ru)
compare to ch 2.1 in the same community.

### Who requested this? (If not you)
Me, but related to observations by Kelsey Dufresne

gabestein:
Intended behavior: when byline is checked at collection level, treat those contributors as authors on each pub within collection.

gabestein:
Notes from today's call:

- Map from CREDIT Taxonomy to CSL values, pass those into citation.js to generate pub-level citations/deposits. Gabe/Allison to work on mapping.
- Is it mapping or just mapping + logic? Gabe to research.

PR #1844, opened 2022-03-01
fix: Always show Collection-level contributors in Collection browser and CitationsModal

idreyn:
A Pub's byline will include the contributors to its primary Collection, but there are two places in the Pub header where we don't incorporate that information:

1. the Collection browser:

image

2. the Cite dropdown:

image

This PR fixes that problem, resolving #1807. This is kind of a cheap fix — what I would really like to do is create a computed/materialized `primaryCollection` association from `Pub` to `Collection` so we don't have to load _all_ a Pub's Collections and compute the primary collection in business logic — but after reading the documentation for a while, I can't figure out how to do this from Sequelize.

_Test plan:_
Create a public, non-Tag Collection and add attributions in Settings. Create a Pub inside that Collection and verify that you see those attributions:
- In the Cite dropdown from that Pub's header
- In the Collection browser from _another_ Pub in the same Collection.

### DIFF
--- a/server/collectionPub/queries.ts
+++ b/server/collectionPub/queries.ts
@@ -7,6 +7,7 @@ import {
 	PubAttribution,
 	Release,
 	includeUserModel,
+	CollectionAttribution,
 } from 'server/models';
 import { getCollectionPubsInCollection } from 'server/utils/collectionQueries';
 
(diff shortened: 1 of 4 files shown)


3mcd (approved):
LGTM!

gabestein:
me too!

gabestein:
Just pushed it live, so it'll be up in a few minutes. (Anything that's merged here shows up on duqduq immediately, so you can always test something out there, but there's a manual step to push it to prod).

avanouse:
Not sure if this is a result of tweaks here, but I just noticed the following potentially-related oddity (in Futurama community) since this started rolling:
Screen Shot 2022-03-04 at 4 52 30 PM

??

PR #1856, opened 2022-03-08
SpubHeaderToolbar

qweliant:
creates: `SpubHeaderToolbar` component 

The toolbar for submissions is now separate from the rendering of of tabs

image

image


3mcd (changes requested):

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+	const submissionTabTitle = renderTabTitle('manually-entered-data', 'Submission');
+	const contributorsTabTitle = renderTabTitle('people', 'Contributors');
+	const previewTabTitle = renderTabTitle('eye-open', 'Preview');
+	const maybeActiveClass = (tabId: string) =>
+		`${tabId === props.selectedTab ? 'active' : 'inactive'}`;
+
+	const { communityData } = usePageContext();
+
+	const lighterAccentColor = useMemo(
+		() => Color(communityData.accentColorLight).alpha(0.2),
```
idreyn, line 36: This should be a lighter version of `accentColorDark` — `accentColorLight` is generally something like white or off-white.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+		[communityData.accentColorLight],
+	);
+
+	const darkerAccentColor = useMemo(
```
idreyn, line 40: I goofed up in the spec for this task and claimed that this was a darker version of `accentColorDark`. As you correctly inferred from the Figma mockup, this is _exactly_ `accentColorDark`...so we can drop this altogether and just use the `--community-accent-dark` CSS variable.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+const renderTabTitle = (icon: IconName, title: string) => (
+	<>
+		<Icon icon={icon} /> {title}
+	</>
+);
+
+type Props = {
+	selectedTab: TabId;
+	onSelectTab: Dispatch<SetStateAction<TabId>>;
```
idreyn, line 19: So `Dispatch<SetStateAction<TabId>>` is the type of the `useState<TabId>` setter, but isn't what you want here, because it rejects callback functions that aren't `useState` setters. Let me break that down.

`SetStateAction<T>` is:
```ts
T | ((currentValue: T) => T)
```
Which is the set of things that the setter function from a `useState` hook accepts...e.g. you can say `setCount(3)` or `setCount(n => n + 1)`. Then, `Dispatch<T>` is just:
```ts
(value: T) => void
```
Together these correctly type the `useState` setter function. But in the component that calls this, we'll likely use a a callback like this to actually trigger a state change:

```ts
// In SpubHeader.tsx
const { updateSubmissionState } = usePubContext();

const setCurrentTab = useCallback((tabId: TabId) => {
    updateSubmissionState({ tabId });
}, [updateSubmissionState]);
```
The type of this function is `Dispatch<TabId>`, and TypeScript will not accept it as a prop here because it doesn't accept an argument like `(currentTab: TabId) => TabId`.

My head is spinning a bit writing all of this, but the long and short is that we don't want to assume that callback props are `useState` setters, because those functions have extra abilities (taking an updater function) that typical callbacks don't. Instead you could use one of these:

- `Dispatch<TabId>`
- `Callback<TabId>` (importing the `Callback<T>` utility from `types`)
- `(t: TabId) => unknown`

Any of these are correct, but the latter two are most common in our codebase.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+	status: SubmissionStatus;
+	showSubmitButton: boolean;
+	onSubmit: () => any;
```
idreyn, line 22: I'm dying on a hill about `unknown` vs `any` [over here](https://github.com/pubpub/pubpub/pull/1866#discussion_r827360781) in Ben's PR.

client/containers/Pub/PubSyncManager.tsx
```diff
@@ -32,7 +32,7 @@ const shimPubContextProps = {
 	historyData: {},
 	firebaseDraftRef: null,
 	updateLocalData: null as any,
-	updatePubData: null as any,
+	updatePubData: () => ({} as any),
```
idreyn, line 35: What you really want here is `(() => {}) as any` but it's not a big deal because I'm about to delete this whole file!

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+						<Tab
+							id="contributors"
+							title={contributorsTabTitle}
+							className={`tab-panel ${maybeActiveClass('contributors')}`}
+						/>
+						<Tab
+							id="preview"
+							title={previewTabTitle}
+							className={`${maybeActiveClass('preview')}`}
```
idreyn, line 95: `` `${x}` `` can be reduced to `x` here (since `x` is already a string)

client/containers/Pub/SpubHeader/SpubHeaderToolBar/spubHeaderToolbar.scss
```diff
@@ -0,0 +1,40 @@
...
+	column-gap: 340px;
+	.tabs.spub-header-component .bp3-tab-list {
+		margin: 0px;
+	}
+	.spub-header-component .bp3-tab-list .bp3-tab.inactive {
+		opacity: 0.45;
+		color: var(--community-accent-dark);
+	}
+}
+
```
idreyn, line 16: As a matter of form, everything inside of a component's SCSS definition _must_ be nested inside the outermost class wrapper. there are very few exceptions to this rule (like when React renders something outside of the component HTML using a Portal). As written this would modify any element on the page that happens to have the class `.status`.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/spubHeaderToolbar.scss
```diff
@@ -0,0 +1,40 @@
...
+.gridParent {
+	&.container {
```
idreyn, line 18: I believe we want to keep the `width: 96%` here to keep things in line with other grid elements. However, we do also want to add this:

```css
.col-12 {
    margin-top: 0;
    margin-bottom: 0;
}
```
to remove the excess vertical margin around one of the `GridWrapper` children. Apologies that this is so fussy — we will eventually migrate off of `GridWrapper`.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/spubHeaderToolbar.scss
```diff
@@ -0,0 +1,40 @@
+@import 'styles/variables.scss';
+
+.spubheader-toolbar {
```
idreyn, line 3: Per our conventions this should be `.spubheader-toolbar-component`.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/spubHeaderToolbar.scss
```diff
@@ -0,0 +1,40 @@
...
+	display: flex;
+}
+
+.status-text {
+	text-transform: capitalize;
+}
+
+.submission-button {
+	width: 119px;
+	height: 30px;
```
idreyn, line 33: I think you can drop all this and use the `minimal` prop on the button instead.

Also, please set `width: 120px`. I know it is exactly `119px` in the mockup, but this number is driving me crazy!

client/containers/Pub/SpubHeader/SpubHeaderToolBar/spubHeaderToolbar.scss
```diff
@@ -0,0 +1,40 @@
+@import 'styles/variables.scss';
+
+.spubheader-toolbar {
+	display: flex;
+	justify-content: center;
+	align-items: baseline;
```
idreyn, line 6: I think you actually want `align-items: center` here to vertically center the button in the toolbar.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/spubHeaderToolbar.scss
```diff
@@ -0,0 +1,40 @@
...
+	column-gap: 340px;
```
idreyn, line 7: Let's make our horizontal separation of the left and right elements more robust to different container sizes. Right now you have:
```css
justify-content: center;
column-gap: 340px;
```
This gets some space between the two sides, but it's always exactly `340px`, regardless of the size of the container. Where does that number come from? My hunch is that it emerges from your screen size, and it won't look right on other displays :) Instead what you want is:
```css
justify-content: space-between
```
to put as much space as possible between the two children.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+	);
+	return (
+		<div style={{ background: lighterAccentColor, color: darkerAccentColor }}>
+			<GridWrapper containerClassName="gridParent">
+				<div className="spubheader-toolbar">
+					<Tabs
+						id="spubHeader"
+						onChange={props.onSelectTab}
+						selectedTabId={props.selectedTab}
+						className="spub-header-component tabs bp3-large"
```
idreyn, line 73: I'm skeptical that you need any of these classes here:

- `.spub-header-component` is now the parent component and is dragging in some unwanted background color in Storybook. Guessing this is coming from the tabs in `SpubHeader.tsx`. <kbd>Cmd + C</kbd> <kbd>Cmd + V</kbd>...great power, great responsibility, etc.
- `.tabs` is being used to select the component in the SCSS, but you could just as easily use the `.bp3-tabs`
- `.bp3-large` is doing the right thing, but you should use the `large` prop on `Tabs` instead and let it render the appropriate class to make it big.

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+	);
+	const { pendingCount } = usePendingChanges();
+	const isSaving = pendingCount > 0;
+
+	const status = isSaving ? (
+		<strong>
+			<em>Saving</em>
+		</strong>
+	) : (
+		<p className="status-text">{props.status}</p>
```
idreyn, line 52: This should be a `span` — a `p` is a block-level element and comes with unwanted margin (...or padding? can never remember.)

client/containers/Pub/SpubHeader/SpubHeader.tsx
```diff
@@ -57,21 +56,29 @@ const SpubHeader = (props: Props) => {
 			}),
 		}).catch(() => props.updateLocalData('pub', props.pubData));
 	};
+
+	const updateSubmissionStatus = async (newSubmissionStatus: SubmissionStatus) => {
+		return apiFetch('/api/submissions', {
```
idreyn, line 61: Let me show you the wonders of `apiFetch.put()`:

```ts
apiFetch.put('/api/submissions', {
    status: newSubmissionStatus,
    id: props.pubData.submission.id,
});
```
saves you a call to `JSON.stringify()` because it does it internally :)

client/components/PubHistoryViewer/PubHistoryViewer.stories.tsx
```diff
@@ -311,6 +311,11 @@ export const initialHistoryData = {
 	timestamps: editTimestamps,
 	currentKey: 14678,
 	latestKey: 14678,
+	isViewingHistory: false,
```
idreyn, line 314: This is nice to have but curious why you made this change — some new TS error?

idreyn (changes requested):
Structurally this is looking good :+1: but there are some color/layout things to address!

client/containers/Pub/SpubHeader/SpubHeader.tsx
```diff
@@ -34,7 +33,7 @@ const SpubHeader = (props: Props) => {
 	const [abstract, setAbstract] = useState(
 		() => props.pubData.submission.abstract || getEmptyDoc(),
 	);
-	const { updatePubData } = usePubContext();
+	const [status, setStatus] = useState(() => props.pubData.submission.status);
```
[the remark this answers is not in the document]
qweliant: not even sure why thats there, thanks!

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
qweliant: image

not if i can help it

client/components/PubHistoryViewer/PubHistoryViewer.stories.tsx
qweliant: pretty much, when trying to render something in storybook it was erroring out without those types being present

client/containers/Pub/SpubHeader/SpubHeaderToolBar/spubHeaderToolbar.scss
qweliant: 🙃🙃🙃

client/containers/Pub/SpubHeader/SpubHeaderToolBar/SpubHeaderToolbar.tsx
qweliant: Callback functions have different abilities than useState functions. we want to use a callback because its like a Philips screwdriver, not a flathead.

idreyn: Good analogy. I think in this case `Dispatch<SetStateAction<TabId>>` is the phillips and `Callback<TabId>` is the flathead (React's `Dispatch<T>` and our `Callback<T>` are identical for our purposes). You can unwrap that to say that `SetStateAction<TabId>` is the phillips screwdriver and `TabId` is the flathead — you can use a `TabId` anywhere you can use a `SetStateAction<TabId>` but not vice-versa.

(it's nearly midnight so I reserve the right to wake up in the morning and realize that this is completely backwards 🤡)

idreyn: If we're talking _screws_, though, then `TabId` is the phillips. You can use it anywhere you have a flathead or phillips _screwdriver_ handy, whereas if you have a flathead screw you need a flathead screwdriver.

This is a pretty interesting analogy for the relationship between values and types.

3mcd (approved):

Issue #1849, opened 2022-03-07
Build the `SpubHeaderTab` wrapper component


idreyn:
The `SpubHeaderTab` is a very simple wrapper around `<GridWrapper>`.We'll use it as the container component for the Instructions, Submission, and Contributors tabs.

- It has the background color `#f8f8f8`
- It provides `30px` of vertical space above and below its contents

It has one tricky prop, `expandToFold`. When this prop is `true`, we'll set the `min-height` of an inner `div` so that this component expands vertically to push the global footer below the fold. We can use this technique:

1. Get a ref to the underlying HTML `div` using `useRef()`
2. Compute its vertical offset from top-of-page with `el.getBoundingClientRect().top + window.scrollY`
3. Use our `useViewport()` hook to get the total viewport height
4. The desired min-height is `(viewport height) - (vertical offset)`

We should put this computation in a `useEffect()` hook.

It should be possible to build this component in Storybook without access to a real Spub.

PR #1857, opened 2022-03-09
Create expandable tab wrapper

howebenhowe:
Implements #1849

Test: view in storybook. To check the expanded size behavior, open inspector (or like) with dimensions:responsive and monitor the `min-height` css property on the wrapper div while changing the viewport; the light-grey background wrapper should resize to fill to the bottom of the viewport.

### DIFF
--- a/client/containers/Pub/SpubHeader/SpubHeaderTab.tsx
+++ b/client/containers/Pub/SpubHeader/SpubHeaderTab.tsx
@@ -1,0 +1,36 @@
+import React, { useRef, useState, useEffect } from 'react';
+
+import { useViewport } from 'client/utils/useViewport';
+
+require('./spubHeaderTab.scss');
+
+type Props = {
+	children?: React.ReactNode;
+	expandToFold: boolean;
+};
+
+const SpubHeaderTabWrapper = (props: Props) => {
+	const elementRef = useRef<HTMLDivElement>(null);
+	const [minHeight, setMinHeight] = useState(0);
+	const { viewportHeight } = useViewport();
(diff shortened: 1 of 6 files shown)


client/containers/Pub/SpubHeader/SpubHeaderTab.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useRef, useState, useEffect } from 'react';
+
+import { useViewport } from 'client/utils/useViewport';
+
+require('./spubHeaderTab.scss');
+
+type Props = {
+	children?: React.ReactNode;
```
idreyn, line 8: No reason to make `children` optional here — if you don't have any, you won't want to use this component!

client/utils/useViewport.ts
```diff
@@ -2,15 +2,22 @@ import { useEffect, useState } from 'react';
 
 export const useViewport = () => {
 	const [viewportWidth, setViewportWidth] = useState(null);
+	const [viewportHeight, setViewportHeight] = useState(null);
```
idreyn, line 5: Dang, sorry, I thought we already had this!

You can get rid of the `@ts-expect-error` comments that this generates by explicitly telling TypeScript what the valid values are for the state: `useState<null | number>(null)`. (Same with `viewportWidth`, if you don't mind)

idreyn (changes requested):
Looks like the expanding is working properly, and thanks for fixing up `useViewport()`. The big remaining change is just to make this component use `GridWrapper` so it properly positions its children.

Issue #1851, opened 2022-03-07
Break the `SubmissionTab` into the `DetailsTab` and `ContributorsTab`


idreyn:
Here we'll break the existing `SubmissionTab` into two new top-level tabs, the `DetailsTab` and `ContributorsTab`. Each should be wrapped in a `SpubHeaderTab`.

[_Figma mocks_](https://www.figma.com/file/ERHvqJqCVBZ7dH2ZQBGlp8/SPUB-vision?node-id=941%3A2039)

Note that what I'm calling the `DetailsTab` is the "Submission" tab in the mockups — I'm choosing a different name for clarity.

## The `DetailsTab`

This is very similar to the existing `TitleSubmissionAbstract` tab, with the following changes:
- Removing the `description` field
- Adding the `DownloadChooser` which is currently in the `SpubSettings` tab. We can revert any changes made to this component's styling in previous work, and treat it as normal Blueprint buttons.
- Removing the last piece of copy (`Enter your primary submission content...`)
- a few copy changes to prompt and placeholder text per the Figma mockups

Some misc tips:
- The mockups still show some (?) info indicators on various fields, which we can safely ignore.
- All of the custom field currently inside `spubHeader.scss` should be scoped to the `.scss` file for this component

## The `ContributorsTab`
This one is easy — we'll just take the existing `ContributorsTab` and wrap it in a `SpubHeaderTab`.

howebenhowe:
Per the mocks, I have used the name `SubmissionTab` for what this card refers to as the `DetailsTab`

PR #1858, opened 2022-03-09
Bh/spubheader break submissions tab

howebenhowe:
Implements #1851 

Overall CSS is fairly close to mocks. It's still missing some spacing and font styling, but I suspect much of the css will be the shared across the unfinished Instructions tab and be informed by the new `SpubheaderToolbar`, so I put it off until those other pieces are in place, when a general CSS roundup would be more timely.


howebenhowe:
Per the mocks, I have used the name `SubmissionTab` for what this card refers to as the `DetailsTab`

client/containers/Pub/SpubHeader/SubmissionTab.tsx
```diff
@@ -46,25 +48,21 @@ const TitleDescriptionAbstract = (props: Props) => {
...
-			<Label>
-				<h2>Pub Content</h2>
-				<span className="input-prompt">
-					Enter your primary submission content in the pub body below by typing or by
-					importing files in which you already have content prepared.
-				</span>
+				<h2>Default Download File</h2>
 			</Label>
-		</>
+			<p className="submission-tab-prompt-text">
```
idreyn, line 53: Is there a CSS declaration that uses this class?

client/containers/Pub/SpubHeader/SpubHeader.tsx
```diff
@@ -82,29 +86,45 @@ const SpubHeader = (props: Props) => {
...
+						/>
+					</SpubHeaderTab>
+				}
+			/>
+			<Tab
+				className={`tab-panel ${maybeActiveClass('contributors')}`}
+				id="contributors"
+				title={contributorsTabTitle}
+				panel={
+					<SpubHeaderTab expandToFold>
```
idreyn, line 116: This one shouldn't have `expandToFold`.

idreyn (approved):
LGTM and I'm happy to defer the CSS pass until later!

One comment: My original intention was for the `SpubHeaderTab` wrapper to be used _inside_ the `ContributorsTab`, `SubmissionTab`, etc. I'm fine with the way this is, but just throwing that out there.

PR #1866, opened 2022-03-15
Add 'Begin Submission' button

howebenhowe:
Adds `Begin Submission` button. This button, when clicked, navigates you to the next tab of the spubheader (the 'Submission` tab).

### DIFF
--- a/client/containers/Pub/SpubHeader/InstructionsTab.tsx
+++ b/client/containers/Pub/SpubHeader/InstructionsTab.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Button } from '@blueprintjs/core';
 
 import { SubmissionWorkflow } from 'types';
 import { Editor } from 'components';
@@ -8,6 +9,7 @@ import SpubHeaderTab from './SpubHeaderTab';
 require('./instructionsTab.scss');
 
 type Props = {
+	onBeginSubmission: () => any;
 	submissionWorkflow: SubmissionWorkflow;
 };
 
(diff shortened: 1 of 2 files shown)


client/containers/Pub/SpubHeader/InstructionsTab.tsx
```diff
@@ -8,13 +9,15 @@ import SpubHeaderTab from './SpubHeaderTab';
...
 type Props = {
+	onBeginSubmission: () => any;
 	submissionWorkflow: SubmissionWorkflow;
 };
 
 const InstructionsTab = (props: Props) => (
 	<SpubHeaderTab expandToFold className="instructions-tab-component">
 		<h1 className="submission-workflow-title">{props.submissionWorkflow.title}</h1>
 		<Editor initialContent={props.submissionWorkflow.instructionsText} isReadOnly />
+		<Button text="Begin Submission" icon="arrow-right" onClick={props.onBeginSubmission} />
```
idreyn, line 20: Can you make this `large`?

client/containers/Pub/SpubHeader/InstructionsTab.tsx
```diff
@@ -8,13 +9,15 @@ import SpubHeaderTab from './SpubHeaderTab';
 require('./instructionsTab.scss');
 
 type Props = {
+	onBeginSubmission: () => any;
```
idreyn, line 12: This is nitpicking but I might as well say it: as I understand it, the "best" return type for callback props that don't return anything is `unknown`. This admits the same set of functions as `any`, but it forbids the component from doing willy-nilly property access like `onBeginSubmission().some.thing`.

idreyn (approved):
LGTM!

client/containers/Pub/SpubHeader/InstructionsTab.tsx
howebenhowe: Right! Meant to ask how to type props-passed functions when we expect  them to be state-setters like this one is.
Unknown is peachy, but what in `[someState, setSomeState] = useState('');` what is the return type of `setSomeState`?

Issue #1876, opened 2022-03-22
Limit the Abstract editor in the Spub header to a single paragraph


idreyn:
You'll need to modify the Prosemirror schema that gets passed in.

PR #1883, opened 2022-03-22
enhancement: one-paragraph abstracts only

howebenhowe:
implements #1876

## _Test Plan_
+ open a spub
+ navigate to the 'submission' tab of that spub
+ give the spub an abstract
+ try to make it multi-paragraph (i.e. hit 'enter') and you should be unsuccessful
+ try to make it multi-line (i.e. hit 'shift + enter') and you should have no problem
+ try to paste in an image or an `H2` copied from elsewhere, and you should be prevented

## Todo
+ is this the behavior and limitation we want? Or do we want to limit the doc to `paragraph+` and then wrap them in a single paragraph on save?


3mcd (approved):
LGTM! And for whatever my opinion is worth, I prefer this over concatenating multiple paragraphs into one.

idreyn (approved):
LGTM and I agree w/Eric that this is the preferred behavior.

Issue #1887, opened 2022-03-23
Fix "Submitted 6 hours from now"


howebenhowe:
Timestamps on `submittedAt` are saved from calls to `Date.toISOString()`, which saves a timezone as the final character of the string. By default, the timezone saved is 'Zulu time', represented by a 'Z', which is the same as GMT or UTC +00.

When this is fed into `react-timeago`, the timezone consideration is lost, and so the displayed timeago phrase will be off by the timezone offset of the submitter.

PR #1889, opened 2022-03-23
bugfix: display correct submittedAt time ago

howebenhowe:
addresses #1887

Now when you submit a pub, it shows as having just submitted:

Screenshot from 2022-03-23 14-03-51

## _Test plan_

1. create a spub
1. submit it
2. check the submitted at line


idreyn (changes requested):
Fix looks good! (which means that you should be able to revert your changes to `getLocalDateMatchingUtcCalendarDate`)

PR #1890, opened 2022-03-23
feat: add pub word count button to formatting bar

### LINKED ISSUE #1769: Word Count
Hey, small thing: I can't find in the documentation whether or not a draft pub can display a word count, and if it can, how to see it. I like pubpub as a writing environment as well as a platform, but this is a feature found in Ulysses etc that is sort of needed. Sorry if I am being obtuse and just missing how to see this! Thanks in advance for any help anyone can give.

3mcd:
Closes #1769 

This PR adds a button to the `PubHeaderFormatting` component that can be used to perform a one-off calculation of the total word and character count of the Pub. A line chart icon was chosen for this button because it represents statistical... stuff. But I'm not totally happy with it so I'm open to other ideas.

Screen Shot 2022-03-24 at 2 05 18 PM

When clicked, a popover containing word and character count should appear immediately below the button.

Screen Shot 2022-03-24 at 2 05 52 PM

These counts are computed only once when the popover component mounts. Clicking on the page or changing focus from the button immediately hides the popover.

### Test Plan

1. Make a new Pub.
2. Type some words and add some figures.
3. Things that *should not* be included in the word or character counts:
   - Whitespace and punctuation (except punctuation surrounded by spaces, e.g. "  ...  ")
   - Citations
   - Figure captions
   - Equation contents
   - Anything that's not in a heading, paragraph, list, or blockquote
4. The popover should hide if you click the screen or give focus to another element using the tab key.

### DIFF
--- a/client/containers/Pub/PubDocument/PubWordCountButton.tsx
+++ b/client/containers/Pub/PubDocument/PubWordCountButton.tsx
@@ -12,9 +12,7 @@ type Props = {
 const getWordAndCharacterCountsFromDoc = (node: Node) => {
 	const text = node.textBetween(0, node.content.size, ' ', ' ');
 	const words = text.split(' ').filter((word) => word !== '');
-	const wordCount = words.length;
-	const characterCount = words.reduce((a, x) => a + x.length, 0);
-	return [wordCount, characterCount];
+	return [words.length, text.length];
 };
 
 const PubHeaderFormattingWordCountButton = (props: Props) => {
--- a/client/containers/Pub/PubDocument/PubWordCountButton.tsx
+++ b/client/containers/Pub/PubDocument/PubWordCountButton.tsx
@@ -17,18 +17,23 @@ const getWordAndCharacterCountsFromDoc = (node: Node) => {
 
 const PubHeaderFormattingWordCountButton = (props: Props) => {
 	const { doc } = props;
-	const [wordCount, characterCount] = useMemo(() => getWordAndCharacterCountsFromDoc(doc), []);
 	const [open, setOpen] = useState(false);
-	const content = (
+	const counts = useMemo(() => {
+		if (open) {
+			return getWordAndCharacterCountsFromDoc(doc);
+		}
+		return null;
+	}, [open, doc]);
+	const content = counts ? (
 		<div className="pub-word-count-button-popover-content">
 			<dl>
 				<dt>Words</dt>
-				<dd>{wordCount.toLocaleString()}</dd>
+				<dd>{counts[0]}</dd>
 				<dt>Characters</dt>
-				<dd>{characterCount.toLocaleString()}</dd>
+				<dd>{counts[1]}</dd>
 			</dl>
 		</div>
-	);
+	) : undefined;
 
 	return (
 		<Popover
--- a/client/containers/Pub/PubDocument/PubWordCountButton.tsx
+++ b/client/containers/Pub/PubDocument/PubWordCountButton.tsx
@@ -24,16 +24,16 @@ const PubHeaderFormattingWordCountButton = (props: Props) => {
 		}
 		return null;
 	}, [open, doc]);
-	const content = counts ? (
+	const content = (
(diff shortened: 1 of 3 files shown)


client/containers/Pub/PubDocument/PubWordCountButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type Props = {
+	doc: Node;
+};
+
+const getWordAndCharacterCountsFromDoc = (node: Node) => {
+	const text = node.textBetween(0, node.content.size, ' ', ' ');
+	const words = text.split(' ').filter((word) => word !== '');
+	const wordCount = words.length;
+	const characterCount = words.reduce((a, x) => a + x.length, 0);
```
howebenhowe, line 16: Should spaces count towards character limit? They are omitted here, yes?

howebenhowe (approved):
LGTM
Such a neat little PR. Giving me hygge.

Only comment is: should spaces count as characters or no?

client/containers/Pub/PubDocument/PubWordCountButton.tsx
gabestein: Good catch. Should definitely count spaces in char count.

howebenhowe: in which case `const characterCount = node.content.size;`?

client/containers/Pub/PubDocument/PubWordCountButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+	return [wordCount, characterCount];
+};
+
+const PubHeaderFormattingWordCountButton = (props: Props) => {
+	const { doc } = props;
+	const [wordCount, characterCount] = useMemo(() => getWordAndCharacterCountsFromDoc(doc), []);
```
idreyn, line 20: This should have `[doc]` as its deps array, right? For performances' sake, I recommend doing something like this:

```ts
const counts = useMemo(() => {
    if (open) {
        return getWordAndCharacterCountsFromDoc(doc);
    }
    return null;
}, [open, doc]);
```

that way we're really only computing this when the popover is open, which seems most ideal.

client/containers/Pub/PubDocument/PubWordCountButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+	const [open, setOpen] = useState(false);
+	const content = (
+		<div className="pub-word-count-button-popover-content">
+			<dl>
```
idreyn, line 24: ah, a rare one!

image

idreyn (changes requested):
Looks great, requesting changes to make sure `useMemo()` recomputes (but I really recommend making the calculation conditional on `open`)

client/containers/Pub/PubDocument/PubWordCountButton.tsx
3mcd: Yep, great suggestion. I'll get this in now.

client/containers/Pub/PubDocument/PubWordCountButton.tsx
3mcd: I wanna be the very best
Like no one ever was
learning tags is my real test
semantics is my cause

client/containers/Pub/PubDocument/PubWordCountButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+const PubHeaderFormattingWordCountButton = (props: Props) => {
+	const { doc } = props;
+	const [open, setOpen] = useState(false);
+	const counts = useMemo(() => {
+		if (open) {
+			return getWordAndCharacterCountsFromDoc(doc);
+		}
+		return null;
+	}, [open, doc]);
+	const content = (
```
idreyn, line 27: fwiw you could also write `const content = counts && <div> ... </div>` since the `null` fallback is fine for `Popover` content. A little slippery.

idreyn (approved):
LGTM!

Issue #1891, opened 2022-03-24
Choose colorblind-friendly status colors


howebenhowe:
Test for contrast within various color blindness constraints on coolors.co

Candidate color pallets for the three displayed statuses (pending, accepted, declined. "incomplete" status never displays as I understand it)

+ [orange - medium green - purple](https://coolors.co/fbb360-43b177-5c415d)
+ [orange - green - brick](https://coolors.co/fbb360-72ca9b-5e0825)
+ [orange - light green - purple](https://coolors.co/fbb360-72ca9b-5c415d)

PR #1871, opened 2022-03-16
Bh/spubheader restructure

howebenhowe:
Implements #1853

+ Consolidates and narrows CSS
+ Performs preliminary wire-up of #1852, to be finalized by that PR's merge

## _Test plan_
+ open an incomplete spub
+ confirm look + behavior of `SpubHeaderToolbar` at top
+ click `Begin Submission` at bottom of `SpubHeader`, it should take you to the `Submission` tab

PR #1892, opened 2022-03-24
Bh/colorblind friendly colors

howebenhowe:
implements #1891

(please dismiss this until #1871 is finished up -- should make most of the commit noise go away)

One trio of colorblind colors has been selected, take a peek if you like.

We will need to reproduce these color changes in the label pairs configured in `getSubmissionStatusLabel` and rendered in `renderLabelPairs`, both within `client/containers/DashboardOverview/overviewRows/labels.tsx`

Seems like we could damn near use the label pair from the `overviewRow` in our new `SpubHeaderToolBar`.


howebenhowe:
mangled git, decided to just start fresh

PR #1903, opened 2022-03-24
Add colorblind status colors for submission status

howebenhowe:
implements #1891

(One trio of colorblind)[https://coolors.co/fbb360-72ca9b-5c415d] colors has been selected, take a peek in the issue if you'd like to see the other similar candidates I put forward.

We will need to reproduce these color changes in the label pairs configured in `getSubmissionStatusLabel` and rendered in `renderLabelPairs`, both within `client/containers/DashboardOverview/overviewRows/labels.tsx`

Seems like we could damn near use the label pair from the `overviewRow` in our new `SpubHeaderToolBar`.


idreyn:
I can only speak for myself as someone with moderate deuteranopia, but I think we should be using red-yellow-green here:

- We don't have to abandon red entirely as a `declined` color, we just have to find one that has a different brightness value than the green. So a darker red and a lighter green, or vice-versa.
- I think we should move away from the orange because that's exactly the kind of color that looks lime green to me at a glance 😭

From the [Blueprint colors](https://blueprintjs.com/docs/#core/colors), I would suggest `red2`, `green4`, and `gold4`.

Separately, we also need to make sure these get used in the `SubmissionRow`, which implies a slightly larger refactor to make sure `IconLabelPair` includes a color instead of an intent (or some equivalent change)

3mcd (approved):
I don't have any color blindness (colorblindedness?), but your justification with Ian's feedback all checks out to me, and the code LGTM too!

client/containers/DashboardOverview/overviewRows/overviewRowSkeleton.scss
```diff
@@ -1,4 +1,11 @@
 @import '../overviewStyles.scss';
+@import '~@blueprintjs/core/lib/scss/variables';
+
+$status-colors: (
+	'pending': $gold4, //#F0B726
+	'accepted': $green2, //#1C6E42
+	'declined': $red2, //#AC2F33
+);
```
howebenhowe, line 8: @idreyn doing this twice (here and also over in `spubHeaderToolBar.scss`) feels icky -- I'm sure there's composability in scss capabilities, but I haven't seen anything like that in old PubPub.

idreyn: I would `@import` one file from the other and then just reuse the variable!

howebenhowe: converted it to a mixin, stashed that in `overviewStyles.scss`, imported from there. Buh-duh-bing.

Issue #1907, opened 2022-03-28
fix: `updateSubmissionState()` after submitting


howebenhowe:
Currently shows `status: pending` immediately after submitting before refresh.

use `updateSubmissionState()` to bring the submission's `pending` status up to speed

PR #1908, opened 2022-03-28
Update submissionsState after submitting

howebenhowe:
Addresses #1907

_Test plan_

create a spub
submit it
immediately after the submission goes through, where there used to be a 'submit' button it should now display `status: pending`


client/containers/Pub/SpubHeader/SpubHeaderToolBar/SubmitDialog.tsx
```diff
@@ -24,7 +27,15 @@ const SubmitDialog = (props: Props) => {
 					id: props.submission.id,
 					status: 'pending' as SubmissionStatus,
 				})
-				.then((submissionRes) => setUpdatedSubmission(submissionRes))
+				.then((submissionRes) => {
+					setUpdatedSubmission(submissionRes);
+					updateSubmissionState(({ submission }) => ({
+						submission: {
```
qweliant, line 33: This is interesting.

We can destructure an object and update the values that may come from the restructured object as well

qweliant (approved):

PR #1914, opened 2022-03-29
Improved design for `SubmissionWorkflowEditor`

idreyn:
The `SubmissionWorkflowEditor` was getting to be just a massive list of fields, so I broke it up into steps so that it's easier to understand the overall process. I also added fields to enable/disable collecting an abstract and a description, which I will have to wire into the header at some point soon.

_Test plan:_
Verify that:
- You can create a new Submission Workflow
- You can also update an existing Submission Workflow

image


3mcd (approved):
Code looks great and works wonderfully! 🐦

qweliant:
are there fixes for the abstract editor not here? Or is this how its suppose to be?

image

client/containers/DashboardSubmissionWorkflow/SubmissionWorkflowStepPicker.tsx
```diff
@@ -0,0 +1,76 @@
...
+			{...tabs}
+			className="submission-workflow-step-picker-component"
+			aria-label="Workflow configuration steps"
+		>
+			{Object.entries(submissionWorkflowConfigStepLabels).map(([step, label], index) => {
+				const reached = selectedStepIndex >= index;
+				const passed = selectedStepIndex > index;
+				const incomplete = passed && !stepCompletions[step];
+				return (
+					<>
```
howebenhowe, line 44: Fragment wrapper looks like it does nothing here.

client/containers/DashboardSubmissionWorkflow/SubmissionWorkflowStepPicker.tsx
```diff
@@ -0,0 +1,76 @@
...
+		<TabList
+			{...tabs}
+			className="submission-workflow-step-picker-component"
+			aria-label="Workflow configuration steps"
+		>
+			{Object.entries(submissionWorkflowConfigStepLabels).map(([step, label], index) => {
+				const reached = selectedStepIndex >= index;
+				const passed = selectedStepIndex > index;
+				const incomplete = passed && !stepCompletions[step];
+				return (
```
howebenhowe, line 43: whether this returns just the `Tab` or a wrapped anything, it'll need a `key={step}` property or similar per React's `Warning: Each child in a list should have a unique "key" prop.`

client/containers/DashboardSubmissionWorkflow/types.ts
```diff
@@ -1,5 +1,19 @@
 import { SubmissionWorkflow } from 'types';
 
+export const submissionWorkflowConfigSteps = [
+	'instructions-requirements',
+	'response-emails',
+	'layout-banner',
+] as const;
```
howebenhowe, line 7: This is weird to see, but only because it is exported and used elsewhere (namely, in `submissionWorkflowStepPicker`.)
Just strange to see something like `import { variableNotAType } from './types';`

howebenhowe (approved):
Just the one React bug, otherwise looks great and tested clean clean clean!

client/containers/DashboardSubmissionWorkflow/types.ts
idreyn: I agree it's a bit weird but there's some precedent for it:

image

idreyn: Probably going to keep it here, but I agree it's a big hmmm

PR #1922, opened 2022-03-30
A few final tweaks to `SpubHeader` style

idreyn:
Ben and I combed over the SpubHeader yesterday and cleaned up a few things stylistically:

- More consistent style for several input components, and an accent-colored focus style
- Using the Pub column width instead of the site column width. This visually scopes the `SpubHeaderToolbar` to the Pub, and matches up much better with the Pub header in the Preview tab. For your consideration @djagdish 
- Big bad inline `<style>` hack to get background colors looking okay during SSR

_Test plan:_
- `SpubHeaderToolbar` should be in same column as Pub
- Inputs should be underlined with `communityAccentDark` when focused
- Instructions tab should have light grey background during SSR

image


3mcd (approved):
Looks great! My only feedback is that the font size of the prompts on the tabs after the Instructions tab could be increased by a point or two. Maybe to 16px?

Screen Shot 2022-03-30 at 9 31 19 AM

Screen Shot 2022-03-30 at 9 31 34 AM

howebenhowe (approved):
looks good tummy

client/containers/Pub/SpubHeader/SpubHeader.tsx
```diff
@@ -80,6 +80,10 @@ const SpubHeader = () => {
 
 	return (
 		<div className="spub-header-component" ref={headerRef}>
+			<style>{`
+				body { background-color: #f8f8f8; }
+				.pub-document-component { background-color: white; }
+			`}</style>
```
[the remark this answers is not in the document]
idreyn: 🤣

client/containers/Pub/SpubHeader/SpubHeaderField.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+	const content = (
+		<>
+			<h2 className="title">{title}</h2>
+			{instructions && <p className="instructions">{instructions}</p>}
+			{children}
+		</>
+	);
+
+	return React.createElement(
```
howebenhowe, line 27: Exotic.

client/containers/Pub/SpubHeader/SpubHeaderField.tsx
```diff
@@ -0,0 +1,40 @@
...
+	asLabel?: boolean;
+	children?: React.ReactNode;
+	title: React.ReactNode;
+	instructions?: React.ReactNode;
+	fullWidth?: boolean;
+	hasMargin?: boolean;
+};
+
+const SpubHeaderField = (props: Props) => {
+	const { asLabel, children, title, instructions, fullWidth = false, hasMargin = true } = props;
```
howebenhowe, line 17: `asLabel` is optional, if unspecified it'll be undefined which is falsy. It isn't given a default.
`fullWidth` is optional, same setup, but it is assigned a default. Is there a subtly I am not seeing, or are these two ways of going about the same thing?

howebenhowe (approved):
Looks great. Nice call on the `SpubHeaderField` component.

client/containers/Pub/SpubHeader/SpubHeaderField.tsx
idreyn: The behavior is identical but I agree that I should add an explicit fallback for `asLabel` for consistency.

PR #1923, opened 2022-03-30
Submission validation in `SpubHeader`

idreyn:
This PR makes some changes related to submission validation to the `SpubHeader`. The test plan doubles as a changelog!

Screen Shot 2022-03-30 at 3 31 50 PM

_Test plan:_

Visit the `SpubHeader` of an unsubmitted Spub and check that it:
- Shows or hides the Description and Abstract fields based on your workflow settings
- Shows warning messages on fields when they have invalid content (only after they've been focused and blurred)
- Disables and adds a warning tooltip to the Submit button if there are invalid fields
- Shows the text _Saving_ inside the button while changes are persisting
- Lets you submit a Pub otherwise

After submission:

- The Abstract field should no longer be editable and should reflect the abstract in the Pub body
- The status indicator should say "Received"
- The status indicator is fully replaced by the text _Saving_ while changes are persisting


client/containers/Pub/SpubHeader/SpubHeader.tsx
```diff
@@ -31,6 +32,21 @@ const SpubHeader = () => {
 	const { selectedTab, submission } = submissionState!;
 	const headerRef = useRef<HTMLDivElement>(null);
 
+	const [validator] = useState(() => createSubmissionValidator(submission.submissionWorkflow));
```
3mcd, line 35: Is this our preference for creating one-off values (as opposed to `useMemo(() => value, [])`?

client/containers/Pub/SpubHeader/validate.ts
```diff
@@ -0,0 +1,36 @@
...
+		abstract: requireAbstract ? isTruthyAnd(isNonEmptyDocJson) : isAlwaysValid,
+	};
+};
+
+export const validateSubmission = (
+	parts: ValidatedSubmission,
+	validator: RecordValidator<ValidatedSubmission>,
+) => {
+	return validate(parts, validator);
+};
```
3mcd, line 36: I feel like these utilities could live somewhere like `utils/submissions` since it could technically be used on the server too, although I see the benefit of placing them near spub header also. Feel free to ignore.

3mcd (approved):

client/containers/Pub/SpubHeader/SpubHeader.tsx
idreyn: I use them both kinda interchangeably but the nice thing about `useState` here is that the linter doesn't complain about `submission` missing from the dependencies array.

client/containers/Pub/SpubHeader/validate.ts
idreyn: Good idea!

PR #1925, opened 2022-03-30
Submissions are Open Banner

qweliant:
This adds a banner to collections that have submissions open 

Test plan:
Navigate to Collections that do not have any submissions open. 

Make sure the banner is not present

Navigate to collections with submissions open

Make sure the banner is present

(order not relevant)

image

### DIFF
--- a/client/containers/DashboardOverview/CollectionOverview/DashboardCollectionOverview.tsx
+++ b/client/containers/DashboardOverview/CollectionOverview/DashboardCollectionOverview.tsx
@@ -246,13 +246,13 @@ const DashboardCollectionOverview = (props: Props) => {
 		);
 	};
 
-	const renderSubmissionBanner = hasSubmissionWorkflow
+	const submissionBanner = hasSubmissionWorkflow
 		? renderBanner('Submissions are now open for this collection!')
 		: null;
 
 	return (
 		<DashboardFrame
-			banner={renderSubmissionBanner}
+			banner={submissionBanner}
 			icon={getSchemaForKind(collection.kind)?.bpDisplayIcon}
 			title="Overview"
 			className="dashboard-collection-overview-container"
--- a/client/containers/DashboardOverview/CollectionOverview/DashboardCollectionOverview.tsx
+++ b/client/containers/DashboardOverview/CollectionOverview/DashboardCollectionOverview.tsx
@@ -2,6 +2,7 @@ import React, { useState, useMemo } from 'react';
 import classNames from 'classnames';
 import { NonIdealState } from '@blueprintjs/core';
 import { DragDropContext, DraggableProvidedDragHandleProps, DropResult } from 'react-beautiful-dnd';
+import Color from 'color';
 
 import { DashboardFrame, DragDropListing, DragHandle } from 'components';
 import { useManyPubs } from 'client/utils/useManyPubs';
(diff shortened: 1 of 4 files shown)


client/containers/DashboardOverview/CollectionOverview/DashboardCollectionOverview.tsx
```diff
@@ -230,8 +238,21 @@ const DashboardCollectionOverview = (props: Props) => {
 		);
 	};
 
+	const renderBanner = (bannerText: String) => {
+		return (
+			<div style={{ background: lighterAccentColor }} className="banner">
+				<div className="banner-text">{bannerText}</div>
+			</div>
```
3mcd, line 245: It may be out of scope now since we're wrapping things up, but I think this is a good opportunity to create a reusable `<Banner>` component. It could optionally accept a `button` or `right` prop that would render something on the far right side of the banner. In this case, we could add a button that would navigate you to the Submissions tab with pressed.

server/routes/dashboardCollectionOverview.tsx
```diff
@@ -9,6 +9,12 @@ import { hostIsValid } from 'server/utils/routes';
 import { generateMetaComponents, renderToNodeStream } from 'server/utils/ssr';
 import { getCollectionOverview } from 'server/utils/queryHelpers';
 import { createUserScopeVisit } from 'server/userScopeVisit/queries';
+import { SubmissionWorkflow } from 'server/models';
+
+const getSubmissionWorkflow = async (collectionId: string): Promise<Number> => {
```
3mcd, line 14: Since the client just need to know whether or not a submission workflow exists for this collection, we could have this function return a boolean:

```ts
const collectionHasSubmissionWorkflow = async (collectionId: string) =>
  (await SubmissionWorkflow.count({ where: { collectionId } })) > 0;
```

server/routes/dashboardCollectionOverview.tsx
```diff
@@ -38,20 +44,27 @@ app.get('/dash/collection/:collectionSlug/overview', async (req, res, next) => {
 		}
 
 		const overviewData = await getCollectionOverview(initialData);
+		const submissionWorkflowPresent = await Promise.all([
+			getSubmissionWorkflow(overviewData.collection.id),
+		]);
```
3mcd, line 49: With the above proposal we could change this statement to:

```ts
const hasSubmissionWorkflow = await collectionHasSubmissionWorkflow(overviewData.collection.id);
```

(and remove the below `submissionWorkflows = submissionWorkflowPresent[0];`)

server/routes/dashboardCollectionOverview.tsx
3mcd: If you choose not to go that route, you can simplify this function by removing `async` and `await`, and make it a one-liner:

```ts
return SubmissionWorkflow.count({ where: { collectionId } }) as Promise<number>
```

server/routes/dashboardCollectionOverview.tsx
3mcd: Otherwise, you can drop the `Promise.all` here:

```ts
const submissionWorkflowPresent = await getSubmissionWorkflow(overviewData.collection.id);
```

3mcd (changes requested):

3mcd (approved):

PR #1927, opened 2022-03-31
fix: use hello@mg.pubpub.org in digest messages to prevent spam alert

3mcd:
Quick fix to remove the spam header on activity digest emails.

Before:
Screen Shot 2022-03-31 at 8 25 53 AM

After:
Screen Shot 2022-03-31 at 8 26 07 AM


idreyn (approved):
Cool!

isTravis (approved):
My goodness, the speed! From bug to PR in negative 6 minutes it feels like! 🙌

PR #1930, opened 2022-03-31
Banner and Button

qweliant:
Digivolve the open submissions banner to be more generic and take ina right positioned ReactNode

image

Test:
Logout and be sure you cant see a banner

Try changing your personal profiles permissions, as admin, and be sure you can't see that banner when logged in, as not the admin

Then log in as admin again, and take a look at that banner...mmmm good ole visual feedback


howebenhowe (approved):
Clean code, test looked good.

PR #1931, opened 2022-03-31
Fix Activity Items Issues

qweliant:
activity renderers will no longer error if `payload.status` key is not present

issue with `from` and to `to` being on `payload.status`

image

fix correctly render messages as follows
image

test plan:
create two new submissions
submit both
delete one and accept the other
you should see submission activity items generated for all actions similar to the ones above


client/utils/activity/render.ts
```diff
@@ -9,6 +9,13 @@ export const renderActivityItem = (
...
 	// This cast avoids a TypeScript error:
 	// Expression produces a union type that is too complex to represent. ts(2590)
+	const keyIsPresent = item.kind in activityItemRenderers;
+	if (keyIsPresent === false) {
+		return null as unknown as RenderedActivityItem;
+	}
 	const renderer = activityItemRenderers[item.kind] as ActivityItemRenderer<any>;
-	return renderer(item, context) as RenderedActivityItem;
+
+	return renderer(item, context).message !== null
```
idreyn, line 18: See my comments below about null error messages, but also — please prefer storing the result of a call like `renderer(item, context)` in a variable rather than calling it twice to retrieve the same value.

types/activity/submission.ts
```diff
@@ -19,7 +19,7 @@ export type SubmissionCreatedActivityItem = SubmissionActivityItemBase & {
 };
 
 export type SubmissionUpdatedActivityItem = SubmissionActivityItemBase & {
-	kind: 'submission-updated';
+	kind: 'submission-status-updated';
 	payload: {
 		status?: Diff<Submission['status']>;
```
idreyn, line 24: Since this item is now exclusively used for status updates, we can make the `status` in the payload non-optional:

```ts
status: Diff<Submission['status']>;
```

This allows us to remove the `if (payload.status)` check in the renderer, and also to get rid of the `message !== null` check in `renderActivityItem`.

client/utils/activity/render.ts
```diff
@@ -9,6 +9,13 @@ export const renderActivityItem = (
 ): RenderedActivityItem => {
 	// This cast avoids a TypeScript error:
 	// Expression produces a union type that is too complex to represent. ts(2590)
+	const keyIsPresent = item.kind in activityItemRenderers;
+	if (keyIsPresent === false) {
+		return null as unknown as RenderedActivityItem;
```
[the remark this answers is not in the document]
qweliant: 😅😅😅😅

client/utils/activity/renderers/submission.tsx
```diff
@@ -16,33 +16,18 @@ export const renderSubmissionUpdated = itemRenderer<SubmissionUpdatedActivityIte
...
-			}
-			if (payload.status.to === 'declined') {
-				return (
-					<>
-						{actor} denied {pub} for submission to {collection}
-					</>
-				);
-			}
+		const { from, to } = payload.status;
+		if (payload.status.to === 'pending') {
```
idreyn, line 20: I thought we changed this to `received`?

client/utils/activity/render.ts
```diff
@@ -7,8 +7,7 @@ export const renderActivityItem = (
 	item: ActivityItem,
 	context: ActivityRenderContext,
 ): RenderedActivityItem => {
-	// This cast avoids a TypeScript error:
-	// Expression produces a union type that is too complex to represent. ts(2590)
 	const renderer = activityItemRenderers[item.kind] as ActivityItemRenderer<any>;
+
```
idreyn, line 11: Is this added space intentional?

server/activityItem/queries.ts
```diff
@@ -583,13 +583,14 @@ export const createSubmissionUpdatedActivityItem = async (
 		communityId: submission.pub.communityId,
 		collectionId: submission.submissionWorkflow.collectionId,
 		pubId: submission.pub.id,
-		kind: 'submission-updated' as const,
+		kind: 'submission-status-updated' as const,
```
idreyn, line 586: This function still needs to be changed so that it only calls `createActivityItem` when the submission status changes (i.e. when `oldSubmission.status !== newSubmission.status`).

idreyn (changes requested):
Sorry, forgot to mention we only want to create an activity item if the status has actually changed. Otherwise every time a `Submission` row is updated (e.g. when updating the abstract) we'll display a row like:

```
{actor} updated the submission status from incomplete to incomplete
```

client/utils/activity/renderers/submission.tsx
qweliant: this doesn't have the branch included from earlier

idreyn (approved):
One comment inline but otherwise lgtm!

PR #1950, opened 2022-04-14
Fix: show only native submissions in submissions overview

howebenhowe:
For some collection X, when viewing the Submissions tab of X's overview, only show spubs submitted directly to X, filtering out spubs that were added to the collection from elsewhere.

### DIFF
--- a/client/containers/DashboardSubmissions/SubmissionItems.tsx
+++ b/client/containers/DashboardSubmissions/SubmissionItems.tsx
@@ -58,7 +58,7 @@ const SubmissionItems = (props: Props) => {
 	const isSearchingOrFiltering = !!filter || !!searchTerm;
 
 	const {
-		currentQuery: { pubs, isLoading, hasLoadedAllPubs, loadMorePubs },
+		currentQuery: { pubs: foundPubs, isLoading, hasLoadedAllPubs, loadMorePubs },
 	} = useManyPubs<PubWithSubmission>({
 		isEager: isSearchingOrFiltering,
 		initialPubs,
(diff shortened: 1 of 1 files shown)


howebenhowe:
This is a rather dirty way of limiting what spubs are displayed, and I don't know about how it will interact with paginated query results for large numbers of spubs. I would prefer to limit the results queried by `useManyPubs` to include only the set we intend to display here, but `useManyPubs` is a bit of a hairy monster and I'm fearful of modifying it, especially if it will be more receptive to augmentation after #1867.

client/containers/DashboardSubmissions/SubmissionItems.tsx
```diff
@@ -73,6 +73,11 @@ const SubmissionItems = (props: Props) => {
 		},
 	});
 
+	const pubs = foundPubs.filter((pub) => {
+		const submittedCollectionId = pub.submission?.submissionWorkflow?.collectionId;
+		return submittedCollectionId && submittedCollectionId === collection.id;
+	});
```
3mcd, line 79: (feel free to ignore) You might be able to simplify this to:
```ts
foundPubs.filter((pub) => pub.submission?.submissionWorkflow?.collectionId === collection.id)
```

3mcd (approved):

3mcd:
I think this is a fine fix for now. We may eventually want to create a new endpoint specifically for loading submission pubs of a specific Collection or SubmissionWorkflow, and somehow hook it up to `useManyPubs`.

PR #1954, opened 2022-04-19
fix: extra height on sticky banner in Pub release view

### LINKED ISSUE #1952: Pub sticky header too long in some cases
So we keep track of it for next week...

1. Scroll down on a pub, see the sticky header is too long

Sticky header is the length of line of text only.

https://notes.knowledgefutures.org/pub/9p4jxiws/release/1

OS X Latest / Chrome Latest

===

Screen Shot 2022-04-14 at 16 36 42

3mcd:
Resolves #1952

Test Plan
1. Load a Pub release
2. Scroll the page so the sticky banner appears and sticks to the top of the viewport.
3. The banner should not have extra "padding" at the top.

### DIFF
--- a/client/containers/Pub/PubHeader/ResponsiveHeaderButton.tsx
+++ b/client/containers/Pub/PubHeader/ResponsiveHeaderButton.tsx
@@ -4,9 +4,9 @@ import { useViewport } from 'client/utils/useViewport';
 
 import { mobileViewportCutoff } from './constants';
 import LargeHeaderButton from './LargeHeaderButton';
-import SmallHeaderButton from './SmallHeaderButton';
+import SmallHeaderButton, { Props as SmallHeaderButtonProps } from './SmallHeaderButton';
 
-type Props = {
+type Props = SmallHeaderButtonProps & {
 	simpleLabel?: React.ReactNode;
 	showCaret?: boolean;
 	outerLabel?: any;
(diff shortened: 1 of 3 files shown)


client/containers/Pub/PubHeader/ResponsiveHeaderButton.tsx
```diff
@@ -17,13 +17,9 @@ const ResponsiveHeaderButton = React.forwardRef((props: Props, ref) => {
...
 	const largeOnlyProps = { outerLabel, showCaret };
 	const smallOnlyProps = { labelPosition };
-
 	const { viewportWidth } = useViewport();
 
-	if (viewportWidth === null) {
-		return null;
-	}
-	if (viewportWidth > mobileViewportCutoff) {
+	if (viewportWidth === null || viewportWidth > mobileViewportCutoff) {
```
3mcd, line 22: This is the only relevant change. The width/height values returned by `useViewport` are `null` during SSR. Previously, we returned `null` from `ResponsiveHeaderButton` during SSR. With these changes, we'll render the `LargeHeaderButton` by default. On mobile, the buttons will appear large and then shrink down once the document hydrates. I've confirmed that the sticky banner works on both mobile and desktop viewport widths.

howebenhowe (approved):

howebenhowe:
LGTM, tested properly for both mobile and desktop viewport widths.

PR #1958, opened 2022-04-20
fix: use pandoc tmp file output instead of reading from stdout

### LINKED ISSUE #1943: Cannot open exported docx or epub

3mcd:
Resolves #1943

Test Plan
1. Create a Pub. Add a structured citation and an unstructured citation.
2. Download the Pub as a Word document.
3. Open the file using a word processor like Pages (mac)/Word
4. Repeat steps 2-3 for epub, open with an epub viewer like Books (mac).
5. Repeat steps 2-3 for a text format like markdown and an XML format like JATS.
6. The JATS export should have one `mixed-citation` and one `element-citation`.

### DIFF
--- a/workers/tasks/export/notes.ts
+++ b/workers/tasks/export/notes.ts
@@ -6,7 +6,6 @@ import { jsonToNode, getNotes, Note } from 'components/Editor';
 import { NoteManager } from 'client/utils/notes';
 import { getStructuredCitations } from 'server/utils/citations';
 import { RenderedStructuredValues } from 'utils/notesCore';
-import { PandocTarget } from 'utils/export/formats';
 
 import { NotesData, NoteWithStructuredHtml, PubMetadata } from './types';
 
--- a/workers/tasks/export/notes.ts
+++ b/workers/tasks/export/notes.ts
@@ -127,42 +127,35 @@ const emptyElementCitation =
 
 export const modifyJatsContentToIncludeUnstructuredNotes = (
 	documentContent: string,
-	target: PandocTarget,
 	notes: PandocNotes,
 ) => {
-	if (target === 'jats_archiving') {
-		return documentContent.replace(
-			emptyElementCitation,
-			(match, id, spaceBefore, spaceAfter) => {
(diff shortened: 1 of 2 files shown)


workers/tasks/export/notes.ts
```diff
@@ -127,42 +127,35 @@ const emptyElementCitation =
...
-	return documentContent;
+						};
+					},
+				},
+			});
+			const content = `<mixed-citation>${unstructuredContentAsJats}</mixed-citation>`;
+			return `<ref id="ref-${id}">${spaceBefore}${content}${spaceAfter}</ref>`;
+		}
+		return match;
+	});
```
3mcd, line 160: This massive diff was just me removing the `if (target === "jats_archiving")` condition.

workers/tasks/export/pandoc.ts
```diff
@@ -186,21 +184,31 @@ export const exportWithPandoc = async (options: ExportWithPandocOptions) => {
...
+	// At this point, pandoc has written the document to our temp file. Here
+	// we do some additional post-processing.
+	if (pandocTarget === 'jats_archiving') {
+		const pandocOutput = fs.readFileSync(tmpFile.path).toString();
+		const transformedOutput = modifyJatsContentToIncludeUnstructuredNotes(
+			pandocOutput,
+			pandocNotes,
+		);
+		fs.writeFileSync(tmpFile.path, transformedOutput);
+	}
```
3mcd, line 213: I moved the check here because overwriting the contents of docx/epub files was corrupting the contents too (even when it should have been a no-op). Still not totally sure why.

workers/tasks/export/notes.ts
```diff
@@ -127,42 +127,35 @@ const emptyElementCitation =
 
 export const modifyJatsContentToIncludeUnstructuredNotes = (
 	documentContent: string,
-	target: PandocTarget,
```
qweliant, line 130: Remove unused export `PandocTarget`

qweliant (approved):

Issue #1953, opened 2022-04-19
Pub route rate limiting


gabestein:
Web crawler/scrapers on larger communities may rapidly request Pub information, putting a strain on PubPub's web server and databases (including Firebase).

We should look into using a library like [`express-rate-limit`](https://www.npmjs.com/package/express-rate-limit) to throttle incoming requests to our public, Pub-specific endpoints from crawlers or malicious actors. These routes can be found in [server/routes/pubDocument.tsx](https://github.com/pubpub/pubpub/blob/master/server/routes/pubDocument.tsx).

We could also investigate [`express-slow-down`](https://www.npmjs.com/package/express-slow-down) (by the same author) to process incoming requests after a configurable delay during "hot" periods (i.e. someone is crawling a large community).

PR #1959, opened 2022-04-20
enhancement: Slowdown Route Request 

qweliant:
This addresses #1953 
webscraping is intensive on the pubDocument API routes(because scraping pubs). We need to enable rate limiting to throttle requests and add resistance to potential DOS attacks. 

Rate limiting is used to limit requests from IP addresses or users or both. We can use two libraries
- express-rate-limit 
    - To alleviate request load through request rejection
   
- Since the node version we use is 12.13.1 we can't use express-rate-limit 
- we will instead use express-slow-down 

- express-slow-down
    - Will slow down the request and add them to a queue to be completed at some time
    - adds a new request field called `req.slowdown`

## Test Plan:
Set API limiting to something to something this for the routes in `server/routes/pubDocument.tsx` :
```
const speedLimiter = slowDown({
	windowMs: 60000, // 1 minute for requests to be kept in memory. value of 60000ms is default but expressed here for clarity
	delayAfter: 10, // allow 10 requests per minute, then...
	delayMs: 100, // 100th request has a 100ms delay, 7th has a 200ms delay, 8th gets 300ms, etc.
	maxDelay: 20000, // max time of request delay will be 20secs
});
```
### Option 1
do `console.log(req.slowDown)` in each app.get function

you should see something like this:

  image

I used different settings to verify it worked under normal circumstances, but using the settings above will allow you to see how the response time slows until it hits 20secs. 

### Option 2
download postman and create a new collection under your default workspace
image

simply requesting any valid API route with test settings should trigger a response slow down

image

image

image

### DIFF
--- a/server/routes/pubDocument.tsx
+++ b/server/routes/pubDocument.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import slowDown from 'express-slow-down';
 
 import { getPubPageContextTitle } from 'utils/pubPageTitle';
 import { getPdfDownloadUrl, getTextAbstract, getGoogleScholarNotes } from 'utils/pub/metadata';
@@ -112,6 +113,15 @@ const getEnrichedPubData = async ({
 	};
 };
 
+const speedLimiter = slowDown({
+	windowMs: 15 * 60 * 1000, // 15 minutes
+	delayAfter: 3, // allow 10 requests per 15 minutes, then...
+	delayMs: 500,
+});
+
+//  apply to all requests
+app.use(speedLimiter);
+
 app.get('/pub/:pubSlug/release/:releaseNumber', async (req, res, next) => {
 	if (!hostIsValid(req, 'community')) {
 		return next();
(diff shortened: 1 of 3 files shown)


server/routes/pubDocument.tsx
```diff
@@ -112,7 +113,14 @@ const getEnrichedPubData = async ({
 	};
 };
 
-app.get('/pub/:pubSlug/release/:releaseNumber', async (req, res, next) => {
+const speedLimiter = slowDown({
+	windowMs: 60000, // 1 minute for requests to be kept in memory. value of 60000ms is default but expressed here for clarity
+	delayAfter: 240, // allow 240 requests per minute, then...
```
3mcd, line 118: I think 240 may be a little high. Maybe 60 requests per minute/one request per second?

server/routes/pubDocument.tsx
```diff
@@ -156,7 +164,7 @@ app.get('/pub/:pubSlug/release-id/:releaseId', async (req, res, next) => {
 	}
 });
 
-app.get('/pub/:pubSlug/discussion-id/:discussionId', async (req, res, next) => {
+app.get('/pub/:pubSlug/discussion-id/:discussionId', speedLimiter, async (req, res, next) => {
```
3mcd, line 167: We may not want to rate-limit the discussions endpoint.

3mcd (approved):

PR #1968, opened 2022-04-25
fix: hide submit button on submitted spubs

howebenhowe:
Nothing fancy, just hiding the submission button when the spub is not 'incomplete'.

when bugged:
Screenshot from 2022-04-25 10-43-53


qweliant (approved):
Totally right. We should probs rethink how this entire flow is supposed to work. I think after submitting the pub shouldn't be editable.

PR #1976, opened 2022-04-27
Bh/add discussion tooltip

howebenhowe:
Tiny improvements, adding tooltip on hover to inline discussion button and ctrl + alt + [number] mapping to setting heading levels
Screenshot from 2022-04-26 22-08-13

_Test plan_

- (tooltip) go to any released pub, select some pub body text, hover over discussion icon
- (heading keys) Go to any pub draft, select some pub body text, and press `ctrl-alt-1` (or `2` or `3`), text should format to heading of the specified number. Behavior should be consistent with previously existing `shift-alt-[number]`


howebenhowe:
@3mcd @gabestein as this is a change to the editor behavior, I'd love sign-off on the function of the heading shortcut keys, as specified in the test plan above

Issue #1917, opened 2022-03-29
spike: math editor experience


gabestein:
See: #1869 

Desire:

Block: $$ \math $$
Inline: $\math

e.g. Is something like this https://github.com/benrbray/prosemirror-math doable?

howebenhowe:
see [writeup on the question](https://notes.knowledgefutures.org/pub/wy4480jp/draft)

PR #1977, opened 2022-04-27
enhancement: inline math editor

howebenhowe:
Implements #1917

See #1869 for further reference.
Uses [prosemirror-math](https://github.com/benrbray/prosemirror-math)

## Features

+ clicking the old equation formatting bar button will insert an empty `math_inline` node at the cursor's position
+ Typing `$stuff$` will result in an `math_inline` node.
+ Typing `$$` then space will result in a new `math_display` node.
+ Old `equation` nodes are converted at document load to new `math` nodes.

## Known places for improvement
+ after clicking the equation formatting bar button to insert an empty `math_inline` node, the cursor cursor should be placed inside the newly created node
+ after beginning a line of prose, typing `$$ ` causes the preceding text, which is in the same block as the newly-written double $, to be converted to a `math_display` node. Preferably, it would insert a `math_display` node _after_ the prose.

## Testing

Try typing with it, see how entering and exiting the math nodes feels. For copying in old `equation` nodes to test the node translation plugin, you can draw from [this reference pub on duqduq](https://demo.duqduq.org/pub/pxbv0zyj/draft?collectionSlug=math)

Try out the following exercises w/ new editing

+ create an empty testing pub
+ write an inline math node
+ write a display math node
+ copy/paste equation nodes from an old pub (try a mix of inline node, display node, and a selection of text that includes both as well as other node types)
+ try clicking into/out of math nodes
+ try using arrow keys to navigate into and out of math nodes
+ export math-laden pub to PDF


client/components/Editor/plugins/equationsToMath.ts
```diff
@@ -0,0 +1,58 @@
...
+		return transaction;
+	}
+	return null;
+};
+
+export default (_, props) => {
+	if (props.isReadOnly) {
+		return [];
+	}
+	return new Plugin({
```
idreyn, line 47: To reduce the overhead of this plugin iterating through all the nodes on every transaction, it might be worth trying to make sure it only runs once (it shouldn't need to again!). A sketch of how this would work follows.

The first step is to instantiate a "plugin key". This is basically a unique, never-colliding identifier for this plugin to use when it stores and retrieves information from shared PM state (in `EditorState` and `Transaction` objects):

```ts
const equationsToMathPluginKey = new PluginKey("equations-to-math");
```

Then modify the `Plugin` constructor to hold a state value called `hasRun`. This starts `false` and is updated every transaction — it becomes `true` if the `equationsToMathPluginKey` has added anything to the [transaction metadata](https://prosemirror.net/docs/guide/#state.plugins):

```ts
new Plugin({
    view: ... // as-is,
    appendTransaction: ... // as-is,
    key: equationsToMathPluginKey,
    state: {
        init: () => {
            return { hasRun: false }
        },
        apply: (transaction: Transaction, { hasRun }) => {
            return { hasRun: transaction.getMeta(equationsToMathPluginKey) || hasRun };
        },
    }
})
```

Finally, modify `getEquationTranslationTransactionForState` to (1) signal that it has run and (2) short-circuit if it turns out that it's already run:

```ts
const getEquationTranslationTransactionForState = (editorState: EditorState) => {
    // Return early if the plugin state says we already did the translation.
    const pluginState = equationsToMathPluginKey.getState(editorState);
    if (pluginState.hasRun) {
         return null;
    }
    // ... some time later ...
    if (mustReturnTransaction) {
        // Signal back to the plugin state that this transaction handled the node translations
        transaction.setMeta(equationsToMathPluginKey, true);
    }
};
```

idreyn (commented):
This looks so good to me! Two minor comments and a potentially larger one:

1. It would be neat if we could retain the toolbar button for equations.
2. Interestingly, the editor thinks RTL is active when a `math` node is selected. Might be worth trying to figure out why: 
image
3. Before we ship this, we need migrate the equation-numbering capabilities of the `block_equation` node. This mostly looks like copying over the `reactive` schema entry and the `equation-label` bits from `toDOM`. There's also a checkbox in the toolbar for `block_equation` to skip numbering that we need to find a solution for. 
image

gabestein:
> It would be neat if we could retain the toolbar button for equations.
> There's also a checkbox in the toolbar for block_equation to skip numbering that we need to find a solution for.

Agreed on both. For the latter, I was thinking maybe we could use an inline menu similar to the link or table menu to both toggle that box and cycle between inline and block?

client/components/Editor/plugins/paste/decorations.ts
```diff
@@ -5,7 +5,8 @@ import { pastePluginKey } from './key';
 import { Meta, PluginState } from './types';
 import { createPlaceholderWidgetElement } from './widget';
 
-type PlaceholderDecoration = Decoration<{ key: string; id: string }>;
+type PlaceholderDecoration = Decoration;
+// type PlaceholderDecoration = Decoration<{ key: string; id: string }>;
```
idreyn, line 9: Spare comment to remove

client/components/FormattingBar/controlComponents/ControlsMath.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+	const mathButtons: FormattingBarButtonData[][] = [[swapMathTypeButton]];
+	if (isDisplay) mathButtons.push([toggleMathLabelButton]);
+
+	return (
+		<FormattingBar
+			editorChangeObject={editorChangeObject}
+			buttons={mathButtons}
+			showBlockTypes={false}
+			isSmall
```
idreyn, line 45: I think we should remove `isSmall` here and let the buttons be the standard size. The toolbar gets kind of beefy, but I have a set of CSS changes lying around in my stash that might help with that.

client/components/FormattingBar/controlComponents/ControlsMath.tsx
```diff
@@ -0,0 +1,50 @@
...
+		title: 'Change display position',
+		label: `Make ${otherPosition}`,
+		icon: 'publish-function',
+		command: mathToggleKind,
+	};
+	const toggleMathLabelButton: FormattingBarButtonData = {
+		key: 'toggle-math-display-label',
+		title: 'Toggle label',
+		label: `${otherVisibility} label`,
+		icon: 'function',
```
idreyn, line 33: I'd suggest `eye-off` and `eye-open` for this.

idreyn (approved):
Looks great to me (and thank you for cleaning up so many of those random TypeScript annoyances along the way!)

I have some small suggestions for the `FormattingBar` layout (and I have some bits of CSS I might try to add to the mix, if that helps) but this looks ready to me!

Issue #1882, opened 2022-03-22
Unpublished pubs w/ custom pub date don't say "unpublished" on thumbnail cards


metasj:
As an author, after setting a custom publication date in my pub's settings, the text on the settings page is also ambiguous about what happens after that field is set.  when viewing a collection that includes it, the pub shows up with "Published on DATE" below it (unlike vanilla drafts that say "Unpublished").  This makes one think no further action is needed for it to be published / submitted for publication. 

An author in one of my communities thought his essay was published for a month before realizing this was not so:

<img width="1213" alt="Screen Shot 2022-03-22 at 12 23 53" src="https://user-images.githubusercontent.com/1615548/159580914-c0d053c5-d03b-4291-a9c2-d0200a3ddf3b.png">

What I expected:
* Unpublished pubs all say "_Unpublished_" on their overview card. 
* The full text might be "_Unpublished (will be: DATE)_ "

PR #1993, opened 2022-05-03
Bh/unpublish custom date thumbnails

howebenhowe:
addresses #1882

I'm not 100% clear on the distinction we draw between `published` and `released`.
For example, in the label for a pub's `OverviewRow`, we show either 'Unreleased' or else 'Published: [date]'. which suggests they are comparable, but in the `getPubPublishedDate` function the distinction is explicitly one of the parameters. It is important to represent the distinction to the user, particularly when they have specified a custom published date for their pub but haven't yet released it. See images below for my crack at representing the distinction.

_Test plan_
Confirm the visual representation of the four combinations of un/released + with/out custom date:
+ unreleased
+ released
+ unreleased w/ custom date
+ released w/ custom date

(see images below for what it ought to look like)

I have prepped four pubs in the collection `published/released/custom` to represent the four pub states

### overview rows
Screenshot from 2022-05-04 10-28-40

### pubs block
Screenshot from 2022-05-04 10-31-46

### pub details (expanded)
Screenshot from 2022-05-04 10-36-26
Screenshot from 2022-05-04 10-36-16
Screenshot from 2022-05-04 10-36-09
Screenshot from 2022-05-04 10-36-01

### pub header title content
Screenshot from 2022-05-03 14-32-01
Screenshot from 2022-05-03 14-31-53
Screenshot from 2022-05-03 14-31-47

### DIFF
--- a/utils/pub/pubDates.ts
+++ b/utils/pub/pubDates.ts
@@ -1,7 +1,8 @@
 import dateFormat from 'dateformat';
 
+import { formatDate, getLocalDateMatchingUtcCalendarDate, isValidDate } from 'utils/dates';
 import { getPrimaryCollection } from 'utils/collections/primary';
-import { getLocalDateMatchingUtcCalendarDate, isValidDate } from 'utils/dates';
+
 import { DefinitelyHas, Maybe, CollectionPub, Pub } from 'types';
 
 export const getPubLatestReleasedDate = (pub: Pub) => {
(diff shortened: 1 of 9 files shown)


howebenhowe:
@idreyn yeah. I had shied from that approach b/c the slight differences here and there on the treatment of the text itself (sometimes in `<i>` tags, sometimes with a colon) but I think it can still be reused plenty.

utils/pub/pubDates.ts
```diff
@@ -31,6 +32,15 @@ export const getPubPublishedDate = (pub: Pub, includeCustomPublishedAt = true) =
 	return null;
 };
 
+export const getPubPublishedDateString = (pub: Pub): string => {
```
idreyn, line 35: The logic here is subtly different than `getPubPublishedDate`: it calls `getLocalDateMatchingUtcCalendarDate` on either a Release date of a `customPublishedAt`, but `getPubPublishedDate` will only do the former.

I suggest calling `getPubPublishedDate(pub, true)` from this function, and also returning `null` instead of the empty string when there's no publication date available.

idreyn (changes requested):
Thanks for unifying all that. Just a quick note about a difference in behavior between this function and `getPubPublishedDate`, which it should mirror (or, ideally, call).

PR #1994, opened 2022-05-03
fix: Update Pub-body citation editor contents when new node is selected

### LINKED ISSUE #1973: Citation rich text field does not update when switching between cites
1. Click on a cite with structured text
2. Click on a different cite with rich test
3. Click on a third cite with structured text
4. Rich text field does not clear

Rich text field clears

===

idreyn:
Resolves #1973

We use the `usePendingAttrs()` hook in the formatting bar to keep track of pending changes to a node (like an image or citation) before they are committed to the ProseMirror editor. This hook lives in the `FormattingBar` and not the individual node editor components (like `ControlsFootnoteCitation`) so that it can commit pending changes even after those controls are closed and unmounted when the editor selection changes.

Currently, when we click from one citation to another, `usePendingAttrs` commits changes to the old node before starting to provide attrs for the new node. But it does this in a `useEffect` hook, and this only fires _after_ a new instance of `ControlsFootnoteCitation` and its "Rich Text" editor have mounted. Since the editor is an uncontrolled `SimpleEditor`, it receives an out-of-date `initialContent` from `usePendingAttrs`.

Basically, I've rewritten `usePendingAttrs` so that it always reports pending attrs for the currently selected node (by indexing into an object that keys these pending attrs by node ID). Thankfully, the logic gets a little simpler since some of the existing implementation was concocted back before nodes had stable IDs, at a time when addressing a specific node in the document was almost impossible.

_Test plan:_
- Create two citation nodes with distinct content in the editor
- Verify that clicking directly from one to the other updates the "Rich Text" editor in the formatting bar
- Verify that you can commit changes to a citation by closing the editor (e.g. by unfocusing from the node)
- Also make sure that other node editors, like the one for image nodes, still work correctly.


3mcd (approved):
LGTM! Awesome explanation, and I agree that the new `usePendingAttrs` is a lot simpler and easier to understand.

howebenhowe (approved):
Everything agreeable, nothing objectionable. Ship, ship away.

PR #1995, opened 2022-05-04
Mobile UX improvements

djagdish:
- new and improved navigation within Dashboard for mobile using BottomMenu
- updated header for mobile 
- overall css changes to support mobile view for existing Dashboard views at all scope levels

### DIFF
--- a/client/components/AttributionEditor/attributionEditor.scss
+++ b/client/components/AttributionEditor/attributionEditor.scss
@@ -1,3 +1,11 @@
+$mobile-viewport-cutoff: 750px;
+
+@mixin mobile {
+	@media screen and (max-width: $mobile-viewport-cutoff) {
+		@content;
+	}
+}
+
 .attribution-editor-component {
 	width: 100%;
 	display: flex;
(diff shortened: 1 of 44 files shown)


gabestein:
Two small things I'm noticing in testing:
- When logged out, login or signup button conflicts with new top right nav bar (still text, needs icon)
- Submissions tab is not in mobile footer menu

idreyn (approved):
To me, this looks about ready to go. I've made some systematic changes to this PR that fall into four categories:

1. Using the `MobileAware` component from #2014 instead of a `useViewport()` check to make decisions in JavaScript about which component to render. This ensures we always server-side render something that looks good to the client.
2. Moving the dashboard-menu rendering logic to `usePageContext()` in #2034 so it's not duplicated across `SideMenu` and `BottomMenu`.
3. Using a single `mobile` SCSS mixin from `styles/mixins.scss` instead of redeclaring it in every file that needs it.
4. Adding a new `--community-accent-dark-faded` CSS variable to represent the background color of the sidebar, which allows us to move some styles declared in JSX `<style>` tags into our SCSS files.

Aside from that, I'm mostly eyeballing that the actually CSS changes look good and trusting that since the vast majority of them are scoped to `@include mobile` blocks, they won't cause any visual regressions for desktop users. All told, this is a fantastic improvement, and I can't wait to see it live!

gabestein:
Thanks for the huge effort here to get this across the finish line. The MobileAware component is super cool and will no doubt be very useful in the future. After a quick look on the phone, it looks great to me and I think we're ready to merge. Leaving a few notes here that are non-blocking for me to come back to and do cards for:

- Activity filters don't wrap
- Need more padding for bottom component so bottoms of some lists (e.g. reviews) clear the bottom bar.
- Scrolling of impact a bit wonky on ios due to iframe-in-frame
- Collection layout/page block settings don't wrap

idreyn:
> Need more padding for bottom component so bottoms of some lists (e.g. reviews) clear the bottom bar.

That's a good callout, I think I'll try to at least get that done before merging.

PR #2001, opened 2022-05-05
fix: don't clear attrs of target node when splitting block

### LINKED ISSUE #1978: heading nodes changing level on paragraph node insertion above
1. created an h2 (or h3 or h4) on the first line of an empty pub
2. moved cursor to start of heading
3. hit 'enter' to create new empty line above heading
4. heading level jumped to h1 regardless of previous level
...

I expected an new empty line to be inserted above but for the heading to maintain its level.

[Any editable pub](https://notes.knowledgefutures.org/pub/wy4480jp/draft)

Browser: Chrome Version 100.0.4896.127 (Official Build) (64-bit)
OS: Ubuntu 20.0.4 LTS

===

3mcd:
Resolves #1978 

Simple fix for a tricky bug!

*Test Plan*
- Create a non-`h1` heading
- Move your cursor to the beginning of the heading node/line
- Hit 'Enter'
- The heading should not be converted into an `h1`

### DIFF
--- a/client/components/Editor/commands/textAlign.ts
+++ b/client/components/Editor/commands/textAlign.ts
@@ -101,9 +101,13 @@ export const splitBlockPreservingTextAlign = (state: EditorState, dispatch?: Dis
 	return splitBlock(state, (tr) => {
 		if (dispatch) {
 			const targetNodePosition = tr.selection.$from.before();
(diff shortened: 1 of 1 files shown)


idreyn:
(also I'm going to merge this since you're out 😄)

Issue #1998, opened 2022-05-05
Community-level 'view' member cannot see private collection in dash or layout


gabestein:
### What went wrong, step-by-step?
1. Added a member as view-level to a Community
2. Browsed to the community dash
3. Could not see any private collections in community dash
4. Could visit and view private collection dashboards via url
5. Could not view private collection layouts (404)
6. Restricted does not make a difference

### What did you expect to happen?
As a community-level view member, should be able to:

- See private collections in community overview dash
- See private collection layouts

This should also apply to editors at the community level. They should have inherited edit-level permissions for all collections.

### What browser/operating system were you using?

===

### Who reported this? (If not you)
Arcadia

PR #2008, opened 2022-05-10
View Level Permissions

qweliant:
addresses issue #1998 

Community scoped members with view permissions can view private collections that are descendants of the community scope

image

image

test:
login as a user with view permissions
from the community dashboard be sure you can see collections that are private
navigate to a collection with a collection layout that has private collections: [here](http://localhost:9876/erics-submission-collection)
hopefully, you can see those

### DIFF
--- a/server/utils/queryHelpers/communitySanitize.ts
+++ b/server/utils/queryHelpers/communitySanitize.ts
@@ -7,7 +7,7 @@ export default (communityData, locationData, loginData, scopeData) => {
 		.filter((item) => {
 			/* Collection access is granted when */
 			/* 1. it is public collection, or */
-			/* 2. the user can view a community or */
+			/* 2. the user has view permission in a community or */
 			/* 3. the user has explicit permissions on the collection */
 			const hasCollectionMemberAccess = item.members.find((member) => {
 				return member.userId === loginData.id;
--- a/server/utils/queryHelpers/communitySanitize.ts
+++ b/server/utils/queryHelpers/communitySanitize.ts
@@ -7,7 +7,7 @@ export default (communityData, locationData, loginData, scopeData) => {
 		.filter((item) => {
 			/* Collection access is granted when */
 			/* 1. it is public collection, or */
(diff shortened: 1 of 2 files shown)


server/utils/queryHelpers/communitySanitize.ts
```diff
@@ -1,18 +1,18 @@
...
-	const { canManageCommunity } = scopeData.activePermissions;
+	const { canManageCommunity, canViewCommunity } = scopeData.activePermissions;
 	const availablePages = {};
 
 	cleanedData.collections = cleanedData.collections
 		.filter((item) => {
 			/* Collection access is granted when */
 			/* 1. it is public collection, or */
-			/* 2. the user is a community manager, or */
+			/* 2. the user can view a community or */
```
idreyn, line 10: I might word this as `the user has view permissions in a community` since "can view" could also mean "able to load the community homepage".

server/utils/queryHelpers/scopeGet.ts
```diff
@@ -328,6 +328,19 @@ getActivePermissions = async (
 		canAdminCommunity ||
 		scopeMemberData.find((member) => member.communityId && member.permissions === 'manage');
 
+	const canEditCommunity =
+		isSuperAdmin ||
```
idreyn, line 332: These checks are redundant — if `isSuperAdmin || canAdminCommutity,` then `canManageCommunity` will be true.

server/utils/queryHelpers/scopeGet.ts
```diff
@@ -328,6 +328,19 @@ getActivePermissions = async (
...
+		canAdminCommunity ||
+		canManageCommunity ||
+		scopeMemberData.find((member) => member.communityId && member.permissions === 'edit');
+
+	const canViewCommunity =
+		isSuperAdmin ||
```
idreyn, line 338: Same here with these three!

idreyn (approved):
LGTM, a few suggestions inline!

PR #2010, opened 2022-05-11
enhancement: Az byline for Arcadia

### LINKED ISSUE #1999: Show external child connections in pub blocks
Currently, pub blocks show only internal (e.g. Pub) connections when the show connections setting is set. It would be great if they showed external connections as well.

Show Pubs as part of a larger network of artifacts.

---

gabestein:
New and improved refactoring!

When Arcadia's community is loaded: 

1. Overrides the byline with an alphabetically sorted list and an `Az` icon with a tooltip. Not implemented in exports (lower priority since it doesn't have the word by, and needs a more major refactor), search (we send algolia a string containing the word 'by'. Will need more of a refactor), or `PubMenuItems` (really tough to enforce type, will take suggestions if you have them).
2. Per request, uses the title of connections instead of authors to display all connections (since they care less about authorship).
3. Fixes #1999 (a bug I'm pretty sure, because there's code for it that was never being accessed) that made it so external child connections were never displayed in pub blocks, also requested by Arcadia.

One todo I'd like your help with: proper type guarding in https://github.com/pubpub/pubpub/pull/2010/files#diff-5bcb9ab7ea16fa0f1d43c3da38cb07845081ba970557f079c5e0f223ddcc9939R103

Screen Shot 2022-05-11 at 11 38 02

_To test_
- Load the `arcadia-research` community locally
- Verify that the new tooltip appears in: pub header, pub previews, overview dashboards, internal pub connections
- Verify that it does not appear in external pub connections
- Load a different community, make sure by still appears.

### DIFF
--- a/client/components/Byline/Byline.tsx
+++ b/client/components/Byline/Byline.tsx
@@ -4,11 +4,13 @@ import ensureUserForAttribution from 'utils/ensureUserForAttribution';
 import { joinOxford, naivePluralize } from 'utils/strings';
 import { AttributableUser, User } from 'types';
 
+require('./byline.scss');
+
 const isFullUser = (u: AttributableUser): u is User => 'slug' in u;
 
 export type BylineProps = {
 	ampersand?: boolean;
-	bylinePrefix?: null | string;
+	bylinePrefix?: React.ReactNode;
 	contributors: (string | {})[] | string;
 	linkToUsers?: boolean;
 	renderEmptyState?: () => React.ReactNode;
--- a/client/components/Layout/LayoutCollectionHeader.tsx
+++ b/client/components/Layout/LayoutCollectionHeader.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Button } from 'reakit/Button';
 
-import { Byline, GridWrapper, ContributorAvatars, Icon, ClickToCopyButton } from 'components';
+import { GridWrapper, ContributorAvatars, Icon, ClickToCopyButton } from 'components';
 import { Collection } from 'types';
 import { LayoutBlockCollectionHeader } from 'utils/layout/types';
 import { getAllCollectionContributors } from 'utils/contributors';
@@ -13,6 +13,7 @@ import {
 	getOrderedCollectionMetadataFields,
 	formattedMetadata,
 } from 'utils/collections/getMetadata';
+import WithinCommunityByline from '../WithinCommunityByline/WithinCommunityByline';
 
 require('./layoutCollectionHeader.scss');
 
@@ -99,7 +100,7 @@ const LayoutCollectionHeader = (props: Props) => {
 			<GridWrapper columnClassName="inner">
 				<h1>{collection.title}</h1>
 				{bylineContributors && !hideByline && (
-					<Byline
+					<WithinCommunityByline
 						contributors={bylineContributors}
 						renderSuffix={() => {
 							return (
--- a/client/components/PubByline/PubByline.tsx
+++ b/client/components/PubByline/PubByline.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 
-import Byline, { BylineProps } from 'components/Byline/Byline';
+import { BylineProps } from 'components/Byline/Byline';
 import { getAllPubContributors } from 'utils/contributors';
 import { Pub } from 'types';
+import WithinCommunityByline from '../WithinCommunityByline/WithinCommunityByline';
 
 type OwnProps = {
 	pubData: Pub;
@@ -22,7 +23,7 @@ type Props = OwnPubBylineProps & typeof defaultProps;
 const PubByline = (props: Props) => {
 	const { pubData, hideAuthors = false, hideContributors = false } = props;
 	const authors = getAllPubContributors(pubData, 'contributors', hideAuthors, hideContributors);
-	return <Byline {...props} contributors={authors} />;
+	return <WithinCommunityByline {...props} contributors={authors} />;
 };
 PubByline.defaultProps = defaultProps;
 export default PubByline;
--- a/client/components/PubEdge/PubEdge.tsx
+++ b/client/components/PubEdge/PubEdge.tsx
@@ -5,8 +5,10 @@ import { Byline } from 'components';
 import { usePageContext } from 'utils/hooks';
 import { getHostnameForUrl, getValuesFromPubEdge } from 'utils/pubEdge';
 
+import { AttributionWithUser } from 'types';
 import PubEdgeLayout from './PubEdgeLayout';
 import PubEdgeDescriptionButton from './PubEdgeDescriptionButton';
+import WithinCommunityByline from '../WithinCommunityByline/WithinCommunityByline';
 
 require('./pubEdge.scss');
 
@@ -32,7 +34,7 @@ const PubEdge = (props: PubEdgeProps) => {
 		communityData,
 		viewingFromTarget,
 	);
-
+	const { externalPublication } = pubEdge;
 	const detailsElementId = `edge-details-${pubEdge.id}`;
 	const detailsElement = description && (
 		<details open={open} id={detailsElementId}>
@@ -93,7 +95,14 @@ const PubEdge = (props: PubEdgeProps) => {
 			})}
 			titleElement={maybeLink(title)}
 			bylineElement={
-				contributors && contributors.length > 0 && <Byline contributors={contributors} />
+				contributors &&
+				contributors.length > 0 &&
+				(externalPublication ? (
+					<Byline contributors={contributors} />
+				) : (
+					// I know there's a better way to do this with a typeguard.
+					<WithinCommunityByline contributors={contributors as AttributionWithUser[]} />
+				))
 			}
 			metadataElements={[
 				description && (
--- a/client/components/PubPreview/ManyAuthorsByline.tsx
+++ b/client/components/PubPreview/ManyAuthorsByline.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
 
-import Byline, { BylineProps } from 'components/Byline/Byline';
-import { naivePluralize } from 'utils/strings';
+import { BylineProps } from 'components/Byline/Byline';
 import { getAllPubContributors } from 'utils/contributors';
 import { Pub } from 'types';
+import WithinCommunityByline from '../WithinCommunityByline/WithinCommunityByline';
 
 require('./manyAuthorsByline.scss');
 
@@ -25,20 +25,12 @@ const ManyAuthorsByline = (props: Props) => {
 	if (isExpanded && isTruncating) {
 		return (
 			<>
-				<Byline
-					contributors={authors}
-					bylinePrefix={`by ${authors.length} ${naivePluralize(
-						'author',
-						authors.length,
-					)}: `}
-					renderUserLabel={(user, index) => `(${index + 1}) ${user.fullName}`}
-					linkToUsers={false}
-				/>
+				<WithinCommunityByline contributors={authors} linkToUsers={false} />
 			</>
 		);
 	}
 	return (
-		<Byline
+		<WithinCommunityByline
 			contributors={authors}
 			renderTruncation={(n) => `${n} more`}
 			truncateAt={truncateAt}
--- a/client/components/PubPreview/PubPreviewEdges.tsx
+++ b/client/components/PubPreview/PubPreviewEdges.tsx
@@ -28,9 +28,7 @@ const getChildEdges = (pubData: Pub) => {
 			}
 			return false;
 		}),
-		...outboundEdges.filter(
-			(edge) => edge.pubIsParent && (edge.targetPub || edge.externalPublication),
-		),
+		...outboundEdges.filter((edge) => edge.pubIsParent),
 	];
 };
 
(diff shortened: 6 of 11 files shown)


client/components/Byline/Byline.tsx
```diff
@@ -72,6 +119,23 @@ const Byline = (props: PrivateBylineProps) => {
 		if (typeof contributors === 'string') {
 			return contributors;
 		}
+		if (bylinePrefix === BylineOption.Az || BylineOption.Za) {
```
3mcd, line 122:
```suggestion
		if (bylinePrefix === BylineOption.Az || bylinePrefix === BylineOption.Za) {
```

client/components/Byline/Byline.tsx
```diff
@@ -72,6 +119,23 @@ const Byline = (props: PrivateBylineProps) => {
...
+						return 1;
+					}
+					return -1;
+				}
+				if (bylinePrefix === BylineOption.Az) {
+					return -1;
+				}
+				return 1;
+			});
+		}
```
3mcd, line 138: Would something this be a bit simpler to read (although not as elegant)?
```ts
return bylinePrefix === BylineOption.Az
  ? lastUser.lastName - nextUser.lastName
  : nextUser.lastName - lastUser.lastName;
```

client/components/Byline/Byline.tsx
```diff
@@ -50,6 +58,45 @@ const Byline = (props: PrivateBylineProps) => {
...
+					children={
+						<Icon iconSize={14} icon="sort-alphabetical-desc" className="byline-icon" />
+					}
+				/>
+			);
+			break;
+		default:
+			bylinePrefixObject = 'by';
+			break;
+	}
```
3mcd, line 99: This might break any existing `bylinePrefix`es used today that aren't "by", since this overwrites the `string` variant of the `bylinePrefix` prop with "by" if it doesn't match one of the `BylineOption` values.

client/components/Layout/LayoutCollectionHeader.tsx
```diff
@@ -94,13 +95,20 @@ const LayoutCollectionHeader = (props: Props) => {
 		<MetadataDetails collection={collection} content={content} key={3} />,
 	].filter((x) => x);
 
+	// If arcadia, use A-Z byline option
+	const bylineOption =
+		collection.communityId === '1a71ef4d-f6fe-40d3-8379-42fa2141db58'
+			? BylineOption.Az
+			: BylineOption.By;
```
3mcd, line 102: I know it's obvious, but could we add a `TODO: implement byline config in scopes` or something along those lines? Same goes for all of the following places we set the byline option for Arcadia.

3mcd (approved):

client/components/PubPreview/pubPreview.scss
```diff
@@ -220,6 +220,13 @@
 		display: inline;
 		font-weight: 500;
 	}
+	.byline-icon {
+		margin-bottom: 3px;
```
3mcd, line 224: Super-duper minor, I prefer using `position: relative` for minor offset tweaks like this. When I read margin my brain thinks "pushing other stuff away".

3mcd (approved):

client/components/Byline/Byline.tsx
```diff
@@ -3,12 +3,20 @@ import React from 'react';
 import ensureUserForAttribution from 'utils/ensureUserForAttribution';
 import { joinOxford, naivePluralize } from 'utils/strings';
 import { AttributableUser, User } from 'types';
+import { Tooltip, Icon } from '@blueprintjs/core';
 
 const isFullUser = (u: AttributableUser): u is User => 'slug' in u;
 
+export enum BylineOption {
```
idreyn, line 10: I think the name `BylineOrdering` would better express what this is.

client/components/Byline/Byline.tsx
```diff
@@ -3,12 +3,20 @@ import React from 'react';
...
+	By = 'by',
+	Contributors = 'Contributors (No Particular Order)',
+	Az = 'Contibutors (A-Z)',
+	Za = 'Contributors (Z-A)',
+}
+
 export type BylineProps = {
 	ampersand?: boolean;
-	bylinePrefix?: null | string;
+	bylinePrefix?: null | string | BylineOption;
```
idreyn, line 19: This is expressing something a bit muddled — the `bylinePrefix` is supposed to be "the thing before the byline" but it's now "the way we sort byline contributors". Even though these things are related, I think they need to remain as two separable options — `bylinePrefix` and `bylineOrdering` — because there are places in the codebase such as Pub Connections cards where we set the `bylinePrefix` to `null`, and, best I can tell, this would override that.

client/components/Layout/LayoutCollectionHeader.tsx
idreyn: My vote is for moving all of the Arcadia logic inside of `Byline` (likewise, until we systematize this some more).

client/components/PubPreview/ManyAuthorsByline.tsx
```diff
@@ -18,27 +17,26 @@ const ManyAuthorsByline = (props: Props) => {
...
 	if (authors.length === 0) {
 		return null;
 	}
 
 	if (isExpanded && isTruncating) {
 		return (
 			<>
-				<Byline
-					contributors={authors}
-					bylinePrefix={`by ${authors.length} ${naivePluralize(
```
idreyn, line 30: What are the consequences of removing this custom `bylinePrefix` logic and the `renderUserLabel` prop?

client/components/PubPreview/pubPreview.scss
idreyn: `tranform: translateY(-3px)` is also a good option here

client/containers/DashboardOverview/overviewRows/overviewRowSkeleton.scss
```diff
@@ -67,6 +67,13 @@ $mobile-viewport-cutoff: 750px;
 		}
 		.byline {
 			font-style: italic;
+			.byline-icon {
```
idreyn, line 70: Let's move the repeated CSS related to the `.byline-icon` into the purview of the `Byline` component. I think this'll mean creating a new `byline.scss` file and importing it from `Byline.tsx`.

client/components/Byline/Byline.tsx
```diff
@@ -72,6 +119,23 @@ const Byline = (props: PrivateBylineProps) => {
 		if (typeof contributors === 'string') {
 			return contributors;
 		}
+		if (bylinePrefix === BylineOption.Az || BylineOption.Za) {
+			// ReviewNote: Do we need to hande the case where a contributor is a string?
```
idreyn, line 123: Yes! We have contributors coming into this component from e.g. the `PubEdgePreview`, where we probably don't have any structured information on their name.

client/components/Byline/Byline.tsx
```diff
@@ -72,6 +119,23 @@ const Byline = (props: PrivateBylineProps) => {
...
+			contributors.sort((last: any, next: any) => {
```
idreyn, line 124: We should avoid sorting arrays in place during React renders. The safe way to do this is by making a copy of the array (`array.concat().sort()`) but that means the logic here will need to be tweaked a bit. I think this calls for a helper function:

```ts
getOrderedBylineContributors(contributors: Contributor[], ordering: BylineOption) => Contributor[]
```
where `Contributor` is (mumble mumble) a `string | Attribution`.

client/components/Byline/Byline.tsx
```diff
@@ -50,6 +58,45 @@ const Byline = (props: PrivateBylineProps) => {
 		truncateAt = null,
 	} = props;
 
+	let bylinePrefixObject;
+
+	switch (bylinePrefix) {
+		case BylineOption.By:
+			bylinePrefixObject = BylineOption.By;
+			break;
+		case BylineOption.Contributors:
```
idreyn, line 67: If the `BylineOption -> IconName` mapping were stored in an object somewhere nearby:

```ts
const iconsForBylineOptions: Record<BylineOption, IconName> = {
    [BylineOption.Contributors]: 'people',
    // ...
}
```
...then we could collapse these three cases into the same piece of code, since they otherwise do the same thing.

client/components/PubByline/PubByline.tsx
```diff
@@ -21,8 +21,13 @@ type Props = OwnPubBylineProps & typeof defaultProps;
 
 const PubByline = (props: Props) => {
 	const { pubData, hideAuthors = false, hideContributors = false } = props;
+	// If arcadia, use A-Z byline option
```
idreyn, line 24: Since we actually do want this logic in place everywhere we use a `Byline`, I suggest moving it into that component.

idreyn (changes requested):
First off I think the outcome here is lovely: I really like how cleanly this icon does its job here (and I wouldn't be surprised if a lot of other Communities opt to use this in the future as well).

I do think we need to rethink some of the logic in this component at a high level. We need to sort out the relationship between two different properties:

- The byline ordering
- The byline prefix (text or icon)

Which are related, but not the same thing. I think we basically need to accept these two props from callers and then choose the "right" default values within the component with an Arcadia-specific override. Ideally we won't need to modify any `Byline` callsites to do this.

As a next step, maybe you can let me bloviate about this some more on a Meet?

client/components/PubPreview/PubPreviewEdges.tsx
```diff
@@ -27,7 +28,9 @@ const getChildEdges = (pubData: Pub) => {
 			}
 			return false;
 		}),
-		...outboundEdges.filter((edge) => edge.pubIsParent && edge.targetPub),
+		...outboundEdges.filter(
+			(edge) => edge.pubIsParent && (edge.targetPub || edge.externalPublication),
+		),
 	];
```
gabestein, line 34: remove (edge.targetPub || )

client/components/PubPreview/PubPreviewEdges.tsx
```diff
@@ -41,34 +44,44 @@ const getEdgesByRelationType = (edges: PubEdge[]): Record<string, PubEdge[]> =>
 	return res;
 };
 
-const renderEdgeLink = (edge: PubEdge) => {
+const renderEdgeLink = (edge: PubEdge, renderTitle: Boolean = false) => {
```
gabestein, line 47: no need for Boolean

client/components/WithinCommunityByline/WithinCommunityByline.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import Byline, { BylineProps } from 'components/Byline/Byline';
+import { AttributionWithUser } from 'types';
+import { usePageContext } from 'utils/hooks';
+import { Tooltip, Icon } from '@blueprintjs/core';
+
+// Review Note; I bet there's a better way to do this?
+type OwnProps = Omit<BylineProps, 'contributors'> & {
+	contributors: AttributionWithUser[];
+};
```
3mcd, line 11: Can just call this `Props`

client/components/PubEdge/PubEdge.tsx
```diff
@@ -93,7 +95,14 @@ const PubEdge = (props: PubEdgeProps) => {
...
 			bylineElement={
-				contributors && contributors.length > 0 && <Byline contributors={contributors} />
+				contributors &&
+				contributors.length > 0 &&
+				(externalPublication ? (
+					<Byline contributors={contributors} />
+				) : (
+					// I know there's a better way to do this with a typeguard.
+					<WithinCommunityByline contributors={contributors as AttributionWithUser[]} />
+				))
```
3mcd, line 105: I think the "ideal" way to solve this issue would be in the type definition of PubEdge. Basically PubEdge would have two variants:
```ts
type BasePubEdge = { /* shared attributes */ }
type ExternalPubEdge = BasePubEdge & { contributors: string[], externalPublication: ExternalPublication }
type InternalPubEdge = BasePubEdge & { contributors: AttributionWithUser[], pub: Pub }
type PubEdge = ExternalPublication | InternalPublication
```

Then, just checking `"externalPublication" in pubEdge` would automatically narrow the type of `pubEdge` from `ExternalPubEdge | InternalPubEdge` to `ExternalPubEdge` or `InternalPubEdge`:

```ts
"externalPublication" in pubEdge ?
  // narrowed to `ExternalPubEdge`
  ? <Byline contributors={pubEdge.contributors} />
  // narrowed to `InternalPubEdge`
  : <WithinCommunityByline contributors={pubEdge.contributors} />
```

3mcd (approved):

client/components/WithinCommunityByline/WithinCommunityByline.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const WithinCommunityByline = (props: OwnProps) => {
+	const { communityData } = usePageContext();
+	const { contributors } = props;
+	let { bylinePrefix } = props;
+	// TODO: Add function to get the sort option from community settings, and add other sort options
+	if (communityData.id === '1a71ef4d-f6fe-40d3-8379-42fa2141db58') {
+		contributors.sort((last, next) => {
```
idreyn, line 19: Don't forget to avoid mutating this array! I'd suggest:

```ts
const sortedContributors = communityId === "..." ? sortContributorsByLastName(contributors) : contributors;
```
and passing that into `Byline`.

idreyn (approved):
This is super clean, looks great!

PR #2014, opened 2022-05-12
Mobile-aware global header buttons

idreyn:
This PR re-implements the changes to the global nav from #1995 with SSR-readiness:

image

Server-side rendering responsive UI is a hard problem. While client-side React can read the viewport width and respond by adding or removing different elements, the server has no such capability and must respond with HTML that looks right without knowing anything about the client's screen size. There are a few strategies in our toolkit to respond to this problem:

1. Parsing the `user-agent` string. In theory (and usually in practice) this is able to tell us whether a device is a phone, tablet, or computer, and allows us to infer something about the viewport width. But there are limitations to this strategy; these strings are becoming gradually more obfuscated and hard to parse, they may be intentionally masked, and they don't account for not-so-edge viewport edge cases like a narrow window open on a wide desktop monitor.
1. The [Client Hints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Client_hints) API, which lets browsers provide the server with "hints" about the nature of the device, notably the viewport size. This would honestly solve all of our problems here, but there are privacy concerns about the scope of the API, and it's [currently not implemented](https://caniuse.com/#search=client-hints) by Firefox or Safari.
1. The "render two, hide one" approach. If we're not sure whether to show the mobile or desktop version, we can just render both, and use CSS media queries to hide the "wrong" one! This approach has limitations and comes with a certain amount of waste, but it works well when applied judiciously.

This PR takes the third approach and systematizes it with a `MobileAware` component that takes `desktop` and `mobile` render props (either a function or a React element). It works like this:

1. The `mobile` and `desktop` elements each receive `{ className, ref }` props from `MobileAware`
2. During SSR, it renders both elements, and uses the `className` passed to its children to hide one of them. (We take this approach to avoid adding extra markup).
3. When the client loads, before React mounts, we read the viewport size using a new `withEarlyMeasurement` option in the `useViewport` hook.
4. When `MobileAware` mounts on the client, it knows the viewport size during its first render! This allows it to forward a `ref` only to the correct child, which becomes important when working with the Reakit API. But we still have to render both elements, because React insists that the server-side render and the first client-side one match exactly.
5. After mounting, `MobileAware` renders again, this time only rendering the correct child.
6. `MobileAware` continues to monitor the viewport size and swap out its rendered child as appropriate.

Now we are able to server-side render two (drastically) different versions of the same thing. Importantly, this lets us build mobile styling _with component APIs_ rather than by inspecting the HTML they render and manipulating them with media queries. For instance, we can manipulate the `large`, `text`, and `icon` props on `Button` instead of targeting its sub-elements' `font-size` and `display` properties in CSS, which will lead to more maintainable code.

_Test plan:_
- Visit [the storybook fixture](http://localhost:9001/?path=/story/components-globalcontrols--default) and check that the right buttons are in the right place. Change the viewport size and verify that all the dropdowns continue to work.
- Then visit the devserver and check the same thing. Try switching from an initial desktop viewport to mobile and vice-versa.

Open question: do we have a better icon for "create Pub"?


client/components/MobileAware/MobileAware.tsx
```diff
@@ -0,0 +1,61 @@
...
+		return null;
+	}
+	const renderProps: RenderProps = {
+		className: isMobileVersion ? `${componentPrefix}__mobile` : `${componentPrefix}__desktop`,
+		...(matchesViewportSize ? { ref } : null),
+	};
+	if (typeof renderable === 'function') {
+		return renderable(renderProps);
+	}
+	return React.cloneElement(renderable, renderProps);
```
howebenhowe, line 43: why not `return <renderable {...renderProps} />;`?

client/components/GlobalControls/UserMenu.tsx
```diff
@@ -0,0 +1,82 @@
...
+	loginData: LoginData;
+};
+
+const handleLogout = () => {
+	apiFetch('/api/logout').then(() => {
+		window.location.href = '/';
+	});
+};
+
+const renderDisclosureButton = (
```
howebenhowe, line 19: First time I've encountered the word 'disclosure' used this way -- took me a minute to end up on the Reakit and WAI-ARIA breakdowns of the disclosure pattern. Struck me as semantically non-obvious, but hey, that's most variable names.

howebenhowe (approved):
Looks great, works great.

I think I'm due for my umpteenth review of how `refs` work.

client/components/MobileAware/MobileAware.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+	useEffect(() => setIsFirstMount(false), []);
+
+	return (
+		<>
+			{renderMobileOrDesktop(mobile, ref, isFirstMount, isMobile, true)}
+			{renderMobileOrDesktop(desktop, ref, isFirstMount, isMobile, false)}
+		</>
+	);
+});
```
3mcd, line 59: Very nice. I feel inspired... I'll apply this to the Pub headers' `ResponsiveHeaderButtons` next!

3mcd (approved):
LGTM!

client/components/MobileAware/MobileAware.tsx
idreyn: This is `React.cloneElement()` not `React.createElement()`!

howebenhowe: I see your `cloneElement` and I raise you (the question) `return <renderable.type {...renderProps} />;`

per [react top-level api documentation](https://reactjs.org/docs/react-api.html#cloneelement) these things should be nearly equivalent. The difference is the [ben]-ne-sais-quoi of preserving refs.

idreyn: It never occurred to me that you could do `<node.type .../>`! `React.cloneElement(node)` is the way I learned how to do this. I'm kind of getting lost in their explanation around refs there, but we do definitely need React to pass along the `ref` from `renderProps` in this case.

Editorializing: I think that example is just there to build intuition and that you should treat React elements (the return value of `React.createElement(X)` or `<X/>`) as black boxes, and avoid inspecting their properties (e.g. `type`) at all costs! I don't think those are part of the public React API.

3mcd: My two cents: disregarding refs, I feel like `cloneElement` is a bit more elegant/easier on the eyes than

```tsx
<renderable.type {...renderable.props} {...renderProps} />
```

client/components/MobileAware/MobileAware.tsx
idreyn: Oh my lord I forgot all about those 😰

PR #2015, opened 2022-05-12
Bump pandoc and support JATS footnotes and Word citations from Mendeley, Endnote, Zotero

gabestein:
Bumps pandoc to 2.18 and fixes a tif-to-png conversion bug.

This gets us:
-  Importing Word citations from Mendeley, Endnote, and Zotero (uses the code we built for processing jats-extracted references to do the same for Word docs). There's probably a universe where we handle that in pandoc-prosemirror instead, but doesn't seem worth it for now.
- Importing JATS footnotes!!!
- Importing LaTeX footnotes!!
- Some other cool stuff I'm sure...

_Test plan_
I tested import and export of complex docs heavily, using [this collection](http://localhost:9876/pandoc-importexport-tests), but you should also take it for a spin. The easiest way to do this is locally, making sure you have pandoc 2.18 installed via brew.

We should also do some group smoke testing on dev before release, particularly of the png-to-tif conversion stuff.

### DIFF
--- a/Aptfile
+++ b/Aptfile
@@ -1,12 +1,16 @@
 # Pandoc is central to the Pub import and export features
-https://github.com/jgm/pandoc/releases/download/2.14.2/pandoc-2.14.2-1-amd64.deb
+
+https://github.com/jgm/pandoc/releases/download/2.18/pandoc-2.18-1-amd64.deb
 
 # xsltproc is ued to extract bibliographies from XML (JATS) files on import
+
 xsltproc
 
 # poppler-utils contains pdftoppm, which we use to import PDF graphics as web-friendy images
+
 :repo:deb http://cz.archive.ubuntu.com/ubuntu bionic main universe
 poppler-utils
 
 # imagemagick is used to convert other images to web-friendly formats
+
 imagemagick
--- a/client/components/PubPreview/PubPreview.tsx
+++ b/client/components/PubPreview/PubPreview.tsx
@@ -1,6 +1,5 @@
 import React, { useState, useRef, useEffect } from 'react';
 import classNames from 'classnames';
-import dateFormat from 'dateformat';
 
 import { Pub, Community } from 'types';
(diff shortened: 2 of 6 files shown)


idreyn (approved):
LGTM and excited for this new functionality!

One caveat: I'm a little worried about using `docx+citations` as the canonical name for the Pandoc format here. As we add more flags to these strings we'll have to remember to update them everywhere or things will randomly break, which is not a good situation to be in.

The way we solved this on the exporter side is with a utility called `getPandocFlags` which computes the format flags separately. On the export side it looks like this:

```ts
const getPandocFlags = (options: ExportWithPandocOptions): PandocFlag[] => {
	const { pandocTarget } = options;
	if (pandocTarget === 'jats_archiving') {
		return [{ name: 'element_citations', enabled: true }];
	}
	return [];
};
```

And here I think it might look like this:

```ts
type PandocFlag = {
    name: string;
    enabled: boolean;
};

const getPandocFlags = (pandocTarget: string): PandocFlag[] => {
    if (pandocTarget === 'docx') {
        return [{name: 'citations', enabled: true}];
    }
    return [];
};
```

(Sadly there's a bunch of work on strongly typing the export pipeline that hasn't made it over to the import side, even though there are obvious symmetries)

For the exporter, those flags are passed as an argument to `getPandocArgs`, which does this to render them to Pandoc's format string:

```ts
const targetPlusFlags = pandocTarget + pandocFlags.map((f) => `${f.enabled ? '+' : '-'}${f.name}`).join('');
```

The importer could do the same, or call `getPandocFlags()` from `getPandocArgs()`. Either way, the result is that we don't "care" about the `+citations` change until the moment it matters, when we're constructing the Pandoc invocation.

I don't want you to get unnecessarily bogged down in this — if you like, feel free to merge as-is and I'll follow up with an implementation of this change.

gabestein:
Thanks! I'm gonna push now and create a new ticket for that follow-up.

Issue #2016, opened 2022-05-12
Double-digit numbers in ordered lists in tables getting cutoff


gabestein:
### What went wrong, step-by-step?
1. Create a table with a list with 10+ items in a column.
2. First digit of double-digits get cut off.

### What did you expect to happen?
See all the digits!

### What URL can we use to see the issue?
See screenshot below.

### What browser/operating system were you using?
Chrome

===

### Have any screenshots? (Attach them by dragging the image here)
![D0eSn19uIrlYJUDl8fV95I4AK99r7AIINg](https://user-images.githubusercontent.com/639110/168167547-c464bf10-02b0-450e-9155-4839724f60fc.png)

### Who reported this? (If not you)
Amara / HDSR

gabestein:
sorry to re-open this one, but a small bug emerged as a result of this. The word-break: break-all is causing weird breaking at the end of some lists, e.g.:

<img width="707" alt="Screen Shot 2022-05-20 at 16 00 40" src="https://user-images.githubusercontent.com/639110/169602495-ac8f40c1-ba7c-4f48-b555-a12196a340a5.png">

I did some playing around and it seems like switching to break-word is fine, but wanted to get your 2c on it in case break-all was key to your solution.

gabestein:
I think `break-word` is basically the expected behavior, and preferable to hyphenating for both the reasons mentioned and the fact that it can sometimes over-hyphenate. But maybe time to tag in @idreyn or @3mcd for a quick peek?

PR #2017, opened 2022-05-12
bug: spacing for list items inside tables

howebenhowe:
addresses #2016 

The price paid is not having the '.' between the number marker and the list item be vertically aligned, as increasing numbers of digits will bump the '.' over. Upshot is, this seems to be standard behavior, at least in common with google docs.

_test plan_

+ check on a pub with > 100 item length un/ordered lists inside a table, oh, for example, [this one](http://localhost:9876/pub/rv25lfjn/draft)
+ spacing should be visually consistent with [duqduq representation of same pub](https://demo.duqduq.org/pub/rv25lfjn/draft)

Screenshot from 2022-05-12 18-39-25


client/containers/Pub/PubDocument/pubBody.scss
```diff
@@ -89,6 +89,7 @@ $pub-body-font-size: if-is-export(14px, 20px);
 		margin: 0;
 		> p {
 			margin: 0;
+			display: inline-block;
```
gabestein, line 92: I think this should be `inline` rather than `inline-block`. Otherwise, when the p within the list item overflows, it will go to the next line. But there's also just something a little weird about making a p inline, so I wonder if there's another approach as well.

Screen Shot 2022-05-13 at 09 05 52

gabestein: Playing more...I think this positioning should be `inline`, but targeted to `p:first-child`. Our editor enforces that there is always a p as the first child of the list-item, which is maybe not great, since it means you can never get, say, an image to display just to the left of the bullet. And it's hard to create multiple `p` elements within a list, but there are a few scenarios where this can happen and cause unpredictable results if every p is set as inline (or inline-block for that matter).

gabestein (approved):
Thanks for grabbing this! Small change needed to the approach to account for ps that span multiple lines, and making sure it doesn't cause weird behavior for multi-child lis.

gabestein (changes requested):
Thanks for grabbing this! Small change needed to the approach to account for ps that span multiple lines, and making sure it doesn't cause weird behavior for multi-child lis.

gabestein (approved):
Thanks for digging in on a wild one! LGTM!

PR #2020, opened 2022-05-15
fix: include authors with non-editor/illustrator/translator/chair roles in citation

gabestein:
Our citation improvements introduced a bug where a byline contributor with a role that was not one of our known ones would not show up as an author in the pub citation. This fixes that by checking the contributor against known roles and, if they don't have one as their primary role, adds them as an author.

It also makes a slight change to behavior with collection-level non-byline editors and chairs by always adding them to citations. This was a regression from prior behavior, but is what we want, as it allows people to credit editors on collections without forcing them to be in the byline.

_To test_
1. Create a pub with two byline authors, one with a custom role; one pub-level editor with byline; one illustrator with byline; one collection-level editor with byline; one collection-level editor without byline.
2. Verify:
Byline includes: two byline authors, pub-level editor with byline, illustrator with byline, collection-level editor with byline
Citation includes:
- Authors: two byline authors only
- Editors (some styles): pub-level editor, collection-level editor with byline, collection-level editor without byline
- Illustrators (some styles): illustrator with byline

e.g.: http://localhost:9876/pub/zo3qc6kw/draft

### DIFF
--- a/utils/contributors.ts
+++ b/utils/contributors.ts
@@ -16,6 +16,8 @@ const orderedContributors = (maybeContributors: Attribution[] | undefined | null
 
 const editorRoles = ['Editor', 'Writing – Review & Editing'];
 const illustratorRoles = ['Illustrator', 'Visualization'];
+const otherKnownRoles = ['Translator', 'Series Editor', 'Chair'];
+
 const getPrimaryRole = (contributor: AttributionWithUser) =>
 	contributor.roles ? contributor.roles[0] : '';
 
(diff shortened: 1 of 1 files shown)


idreyn (changes requested):

idreyn (approved):
Sorry for the RC there — I briefly totally misunderstood the context where this function was operating and now I understand that it's doing the right thing! 😅

qweliant (approved):

Issue #2023, opened 2022-05-16
Collection-level editors and managers cannot view all collections


gabestein:
### What went wrong, step-by-step?
In pushing https://github.com/pubpub/pubpub/pull/2008, we appear to have added a regression where now, edit and manage users on a community-level cannot see or interact with collections they do not have explicit access to.

### What did you expect to happen?
Community-level editors and managers should have appropriate access to 

### What browser/operating system were you using?

===

PR #2025, opened 2022-05-16
Fix View Permissions

qweliant:
solves issue #2023 

I added the can view/manage/edit permissions explicitly

test:
edit your personal user to have view roles from the team pub pub admin
login as that user
be sure you can see all collections at the community level

### DIFF
--- a/server/utils/queryHelpers/scopeGet.ts
+++ b/server/utils/queryHelpers/scopeGet.ts
@@ -321,7 +321,7 @@ getActivePermissions = async (
 
 	const scopeMember = (scopedMember: any, permissionLevel: string) => {
 		return (
-			!permissionLevels.includes(permissionLevel) ||
+			permissionLevels.includes(permissionLevel) &&
 			scopedMember.find(
 				(member) => member.communityId && member.permissions === permissionLevel,
 			)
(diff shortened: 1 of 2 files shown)


gabestein:
Will defer to experts on the code changes, but working as expected from a QA perspective!

server/utils/queryHelpers/scopeGet.ts
```diff
@@ -319,24 +319,22 @@ getActivePermissions = async (
...
+	const scopeMember = (scopedMember: any, permissionLevel: string) => {
+		return (
+			permissionLevels.includes(permissionLevel) &&
+			scopedMember.find(
+				(member) => member.communityId && member.permissions === permissionLevel,
+			)
+		);
+	};
+
+	const member = scopeMember.bind(this, scopeMemberData);
```
idreyn, line 331: A note on `this`: If you're in an arrow function, and most of our functions are arrow functions, then `this` is basically useless — it can't be re-bound to anything! In this specific instance:

1. `this` is not defined for the containing arrow function `getActivePermissions`
2.  The enclosed arrow function `scopeMember` won't care about what you pass as the first argument to `bind`

It's completely harmless to reference `this` here, but it won't let you do anything. I would write:

```ts
scopeMember.bind(null, scopeMemberData)
```

this signals to the reader of this code that you don't expect `scopeMember` to do anything with its `this` value. Which makes sense, because it doesn't reference it! This is good practice even when working with non-arrow functions — if you're not going to use `this`, don't bind it forward!

idreyn (changes requested):
Sorry for the crosstalk here vs. on slack — basically, the permissions gotta cascade! And I have some other suggestions, too.

idreyn (approved):
LGTM!

PR #2026, opened 2022-05-16
fix: About link

gabestein:
So........nesting ULs in P tags, which I guess react doesn't like, created some kind of client/server mismatch that resulted in a single link to our roadmap ("on GitHub") inexplicably being overwritten with a link to our membership page? ¯\\_(ツ)_/¯

_To test_
1. Visit about.
1. Click on all the links and make sure they


idreyn (approved):
Love this stuff. LGTM!

Issue #2027, opened 2022-05-16
Deposit preview crashes page in some instances


gabestein:
Seems to be a regression related to #1972, but I am having trouble recreating the conditions on demo or duqduq.

### What went wrong, step-by-step?
1. Try to preview by (re)-depositing this article: https://jbt.pubpub.org/dash/pub/rvmunuw3/settings
2. Blank screen of death!
3. Error:
``` 
TypeError: Cannot read properties of undefined (reading 'person_name')
    at AssignDoiPreview.tsx:104:59
    at b (AssignDoiPreview.tsx:120:26)
    at v (AssignDoiPreview.tsx:220:76)
```

### What browser/operating system were you using?

===

### Who reported this? (If not you)
Dawit

gabestein:
Ah, figured it out. It's when a collection has no contributors. I'll take care of it.

PR #2029, opened 2022-05-16
bug: Check person_name

qweliant:
fixes issue #2027 

I goofed by removing a check

I added it back.

test:
make a deposit for any collection and remove all contributors. be sure that preview DOI still renders.


gabestein (approved):
Thanks for the quick fix. LGTM, and works as expected in qa.

PR #2032, opened 2022-05-18
chore: bump prosemirror-pandoc

3mcd:
Fixes an small issue with a specific Pub not being exportable due to empty `<tr>` elements.


idreyn (approved):
cool and good

PR #2034, opened 2022-05-19
Move dashboard menu creation to `usePageContext()`

idreyn:
This PR takes the code that renders the Dashboard side menu and moves it into `usePageContext()`, where it is computed once when the page loads. This looks trivial, but it includes things like the list of visible links, active state, counts, etc. The resulting object is of the shape:

```ts
Array<{
    title: string;
    icon: IconName;
    href: string;
    active: boolean;
    count: null | number;
}>;
```

and it can be trivially rendered by any component that might want this information. Like the side menu:

image

Okay. Why is he doing this. The answer is that we're about to render _two_ of these menus, a `SideMenu` and a `BottomMenu`, in our dashboard mobile refactor. So this is a good way for them to share as much code as possible!

_Test plan:_

Visit a (Community, Collection, Pub) and check:
- All the links lead to the appropriate dashboard pages
- The current dashboard tab is selected
- If there are reviews/submissions open, those counts are reflected in the side menu
- If a Pub dashboard tab has a `collectionSlug` URL param, that's carried into the dashboard tab hrefs.
Now log out and do the same thing, checking to make sure only the appropriate tabs are shown.


client/utils/icons.ts
```diff
@@ -0,0 +1,32 @@
...
+	| 'submission'
+	| 'connection'
+	| 'impact'
+	| 'member'
+	| 'settings'
+	| 'collection'
+	| 'pub'
+	| 'community';
+
+export const pubPubIcons: Record<PubPubIconName, IconName> = {
```
idreyn, line 18: This is my first crack at providing PubPub names for our mostly-blueprint set of icons. We should gradually introduce this mapping into our codebase so we don't always need to remember that e.g. the Submissions icon is `manually-entered-data`.

client/containers/DashboardOverview/overviewRows/PubOverviewRow.tsx
```diff
@@ -19,7 +19,7 @@ type Props = {
 	labels?: IconLabelPair[];
 	className?: string;
 	pub: Pub;
-	inCollection?: boolean;
+	inCollection?: Collection;
```
idreyn, line 22: The changes in this file fix a small(?) bug where Pub links from Collection overviews don't have a `collectionSlug`, and we lose the Collection information from the Pub dashboard breadcrumbs.

client/utils/navigation/dashboardNavigation.ts
```diff
@@ -0,0 +1,120 @@
...
+	mode: DashboardMode;
+	subMode: string | null;
+};
+
+export type DashboardMenuState = {
+	menuItems: RenderedDashboardMenuItem[];
+	activeMode: null | ActiveDashboardMode;
+};
+
+const inferDashboardModeFromPath = (path: string): DashboardMode => {
```
idreyn, line 28: It would be really cool if we were just able to grab this from `locationData`...

3mcd (approved):
LGTM!

PR #2035, opened 2022-05-19
Implement multiple emails for submissions

howebenhowe:
implements #1991
Allows adding multiple email correspondence addresses in workflow settings, so more than one person might catch the CC on status-advancement emails.

for big fat lists of emails, you end up here:

Screenshot from 2022-05-19 18-03-04

_test plan_

1. edit a submission workflow
2. add a buncha emails
3. submit a submission
4. go back to submission workflow editor
5. try to find ways to mis-enter email addresses

_potential augmentation_
accommodate the entry of email addresses in  the `Firstname Lastname <email@address.ccom>` format, like how many email clients do, so that copy-pasting from those clients is absorbed with no trouble.


utils/email.ts
```diff
@@ -4,3 +4,5 @@ const emailRegex =
 	/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
 
 export const isValidEmail = (str: string) => emailRegex.test(str);
+export const isValidEmailList = (emailList: string[]) =>
```
idreyn, line 7: In my opinion, it would be better to put this function in `utils/validate`, since its usefulness is pretty closely tied to this specific task of form validation (whereas I could see `isValidEmail` being called in all sorts of places).

tools/migrations/2022_5_19_renameWorkflowTargetAddresses.js
```diff
@@ -0,0 +1,24 @@
...
+		await SubmissionWorkflow.findAll(),
+		(submissionWorkflow) => {
+			submissionWorkflow.targetEmailAddresses = [submissionWorkflow.targetEmailAddress];
+			return submissionWorkflow.save();
+		},
+		{ concurrency: 100 },
+	);
+};
+
+export const down = async ({ sequelize }) => {
```
idreyn, line 22: Potentially handy tip: if you wanted to, you could also add another function here to remove the old column and then run it by passing the `--fn <name>` flag to the migration script instead of `--up` or `--down`.

idreyn (approved):
LGTM, and cool!

PR #2039, opened 2022-05-23
Dev: add laval doi prefix

gabestein:
What it says!

_Test plan_
¯\_(ツ)_/¯


idreyn (approved):
LGTM! Obligatory "we should put this in the database" comment.

PR #2040, opened 2022-05-23
feature: `PubTitle` component

idreyn:
Test plan:

(1) Check that Pub titles render correctly in these places:
- Community/Collection overview rows
- The Pub Overview dashboard tab
- Pub previews on Layouts
- `PubMenuItem` instances (like in the Connections tab Pub search box)
- The Collection browser at the top of the Pub header
- The "Read Next" section at the bottom a Pub

(2) Check that we're still nicely psuedorandomly generating gradients for Pubs on layouts and the Search page.


3mcd:
LGTM!!

PR #2041, opened 2022-05-23
feat: Pub title decorations

3mcd:
This PR adds **bold** and _italics_ decorations to Pub titles!

These decorations can be added through conventional keyboard shortcuts (Meta+B, Meta+I), or through copy/pasting of existing HTML. All HTML other than `b`, `i`, `strong`, and `em` will be stripped out of pasted content. Undo and redo of formatting commands is supported.

https://user-images.githubusercontent.com/6402908/169891770-62656def-48b7-4987-9f7a-0c547bbf8607.mov

_Test Plan_

1. Follow the test steps from #2040 
2. Ensure the title saves on blur. Attempting to navigate away from the draft page should trigger a warning prompt.
3. Ensure Meta+B and Meta+I will bold/italics selections within the Pub title editor.
4. Ensure the Pub title editor is read-only on the release view.
5. Ensure undo/redo work as expected.
6. Attempt to paste HTML content containing tags other than the supported decoration tags. Ensure those tags (including attributes) are stripped.
7. Confirm that Pub titles render with decorations in the Activity tab.


client/components/TitleEditor/TitleEditor.tsx
```diff
@@ -0,0 +1,185 @@
...
+			onChange?.(initialValue);
+		}
+	}, [initialValue, onChange]);
+
+	useBeforeUnload(
+		() => !isReadOnly && focused && node.current?.innerHTML !== initialValue,
+		'You have unsaved changes to this Pub. Are you sure you want to navigate away?',
+	);
+
+	if (typeof window === 'undefined' || isReadOnly) {
```
idreyn, line 161: FYI we have a `ClientOnly` component that abstracts over this kind of `window` checking logic. Not sure if it's worth using here because this looks pretty clean.

client/components/TitleEditor/TitleEditor.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+const commonProps = {
+	role: 'textbox',
+	'aria-label': 'Edit Pub title',
+	tabIndex: 0,
+};
+
+export default function TitleEditor(props: Props) {
+	const { initialValue = '', onChange, isReadOnly = false, ...restProps } = props;
+	const node = useRef<HTMLDivElement>(null);
```
idreyn, line 73: It's blowing my mind that TypeScript doesn't complain about this...I would have thought you'd have to put `null | HTMLDivElement` in the brackets in order to set the initial value to `null`.

client/components/TitleEditor/TitleEditor.tsx
```diff
@@ -0,0 +1,185 @@
...
+	return template;
+}
+
+const commonProps = {
+	role: 'textbox',
+	'aria-label': 'Edit Pub title',
+	tabIndex: 0,
+};
+
+export default function TitleEditor(props: Props) {
```
idreyn, line 71: Cool, I like this way of writing this out.

client/components/TitleEditor/TitleEditor.tsx
```diff
@@ -0,0 +1,185 @@
+import classNames from 'classnames';
+import { useBeforeUnload } from 'react-use';
+import React, { ClipboardEvent, useCallback, useEffect, useRef, useState } from 'react';
+
+require('./titleEditor.scss');
+
+type Props = {
+	isReadOnly?: boolean;
+	initialValue?: string;
+	onChange?: (value: string) => void;
```
idreyn, line 10: Thoughts on doing some tag stripping within this component, so that `onChange` has this signature instead?

```ts
onChange: (update: Pick<Pub, 'title' | 'htmlTitle'>) => unknown;
```

This way the Pub page gets to know its updated `title` in real time, which might be good. We'll probably still want to do the normalization on the server, too, which is annoying. (But this is good: we're starting to see bits of annoying synchronization logic that Facets can abstract over).

idreyn (approved):
This looks pretty great to me. A couple of thoughts, maybe for future PRs:

1. If we want to permit editing a rich-text title from the Pub Settings page, and I think we do, then the `TitleEditor` might need to be broken into presentational and functional layers to deal with styling and behavioral differences — for instance the `useBeforeUnload` behavior may not be applicable there.

2. It would be nice to retain some of the visual cues from the existing title editor: when you hover over it, there's a gentle grey outline that suggests editability, and there's a thick border around it while editing is in progress. (It also turns white regardless of the theme, but I think what you have here is an improvement).

gabestein (approved):
Looks really great!

Two thoughts from play-testing, both non-blocking but maybe worth thinking through:

1) If you delete the title and blur so it shows the placeholder text, it then takes two clicks to get the cursor back and make an edit.

2) You can't paste italics/bold from the pub body into the title, because the body uses em/strong whereas the title uses i/b. I know why, and it's totally fine, but it's a little awkward and some day we might want to map those two, or something?

Overall, huge thumbs up.

PR #2042, opened 2022-05-25
fix: break-word for <ol> in tables

howebenhowe:
The final word.

swap in `break-word` for `break-all`


gabestein (approved):
🙏 setting records for tiny, impactful commits.

PR #2043, opened 2022-05-26
fix: copy text title to html title when updating from dashboard

3mcd:
*Test Plan*
* Change a Pub's title via the Pub header
* Change a Pub's title via the Pub dashboard page
* Go back to the draft/release view. The Pub's title should be updated with the value entered via the Pub dashboard page.


idreyn (approved):
LGTM!

Issue #1794, opened 2022-02-09
Add support for RTL editing


gabestein:
### What's your enhancement request?
Add a button that switches any given text block to RTL mode, like [TinyEditor's Directionality button](https://www.tiny.cloud/docs/plugins/opensource/directionality/). (See example in the ... menu here: https://www.tiny.cloud/docs/demo/full-featured/).

### What would this enhancement help you accomplish?
We currently support left, center, and right alignment, but even in right alignment mode, the cursor is on the right side of the document, making writing in RTL languages difficult to impossible. Adding a button that would switch a block to RTL using CSS would make it much easier to edit in RTL languages.

Prosemirror [seems to support](https://discuss.prosemirror.net/t/does-prosemirror-support-rtl-languages-editing/1318/6) simply using CSS to accomplish this, so I'm hoping it's a relatively straightforward task that will vastly improve our RTL capability.

---

### Any attachments or reference images? (Drag them here)
<img width="762" alt="Screen Shot 2022-02-09 at 10 26 10" src="https://user-images.githubusercontent.com/639110/153232783-1140d17c-ebe0-480d-9caf-ab97c7ddda63.png">

### Who requested this? (If not you)
PEYSSARD Jean-Christophe / Syria Lexicon

gabestein:
- Button in formatting bar that shows on any block where you can set left/right/center alignment
- Ability to set on a block level
- Toggles left or right

PR #2047, opened 2022-05-31
RTL 

qweliant:
adresses issue #1794 

Users can now orient their pubs RTL


client/components/Editor/commands/rtl.ts
```diff
@@ -0,0 +1,63 @@
...
+	attr: string | null,
+) => {
+	const { tr } = state;
+	nodes.forEach(({ pos, node }) =>
+		tr.setNodeMarkup(pos, undefined, { ...node.attrs, [directionAttr]: attr }),
+	);
+	dispatch(tr);
+};
+
+const createDirectionCommandSpec = (direction: string) => {
```
idreyn, line 50: You can probably just get rid of this outer factory function, since we only ever call it once, and write:

```ts
const rtlToggle = createCommandSpec(...);
```

client/components/Editor/commands/rtl.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const isRtlTarget = (node: Node) => {
+	const { spec } = node.type;
+	return spec.attrs && directionAttr in spec.attrs;
+};
+
+// nodes could be in ltr or rtl
+const getDirectionNodes = (state: EditorState): NodePos[] => {
+	const { selection, doc } = state;
+	const alignableNodes: NodePos[] = [];
```
idreyn, line 25: I'd maybe rename this to `rtlTargetNodes` or something, since this no longer has anything to do with alignment.

client/components/Editor/commands/rtl.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// what nodes able to be rtl, not sure, how to check?
+
+const isRtlTarget = (node: Node) => {
+	const { spec } = node.type;
+	return spec.attrs && directionAttr in spec.attrs;
+};
+
+// nodes could be in ltr or rtl
+const getDirectionNodes = (state: EditorState): NodePos[] => {
```
idreyn, line 23: Since you introduced the nicely precise "RTL target" terminology in the function above, I'd stick with it, and name this function `getRtlTargetNodes` or something.

client/components/Editor/commands/rtl.ts
```diff
@@ -0,0 +1,63 @@
...
+	nodes.forEach(({ pos, node }) =>
+		tr.setNodeMarkup(pos, undefined, { ...node.attrs, [directionAttr]: attr }),
+	);
+	dispatch(tr);
+};
+
+const createDirectionCommandSpec = (direction: string) => {
+	return createCommandSpec((dispatch: Dispatch, state: EditorState) => {
+		const nodes = getDirectionNodes(state);
+		const canRun = true; // not sure ?
```
idreyn, line 53: This is true iff `nodes.length > 0` — in other words, if there are any nodes to run the command on.

client/components/Editor/commands/rtl.ts
```diff
@@ -0,0 +1,63 @@
...
+		const directionAttrValue = getDirectionValue(direction, nodes);
+		return {
+			isActive: false,
```
idreyn, line 56: This should be set to `true` if every node in `nodes` has RTL alignment. It highlight the button to indicate that RTL is active on those nodes.

client/components/Editor/commands/rtl.ts
```diff
@@ -0,0 +1,63 @@
...
+			canRun,
+			run: () => orientNodes(nodes, state, dispatch, directionAttrValue),
+		};
+	});
+};
+
+export const rtlToggle = createDirectionCommandSpec('true');
```
idreyn, line 63: I want to mention that there are be better ways to represent the RTL direction than the string `'true'`. For instance, you could store the RTL-ness using `true | null`, and then write:

```ts
{ 'data-rtl': String(node.attrs.rtl) } 
```

when actually specifying DOM nodes in `base.ts`. It feels cleaner to use the least powerful data type required (in this case, a boolean).

Regardless, there should be a TypeScript type in this function somewhere that helps a reader understand what values the `rtl` attribute can take on.

idreyn (changes requested):
This is working pretty well! Mostly, `getDirectionValue` needs a behavioral tweak, and there is a bit more state to provide to the button with `{ isActive, canRun }`, but you're almost there.

One thing I didn't mention is that we'll need to modify `splitBlockPreservingTextAlign` in `textAlign.ts` to also preserve text-direction when splitting a block from another block — so that when you're in RTL mode and you press Enter, you'll continue to be in RTL mode in the next paragraph. That function should probably also be renamed and moved to a different file since it will no longer strictly relate to text alignment. Happy to chat about that some more once you start implementing it.

Overall, this looks really good and I'm excited to see it land soon!

qweliant:
A question I do have is how many other `blockNode`s also require rtl direction.
codeblocks, quotes, etc

gabestein:
How complex is it to expand? Quotes and [code](https://en.wikipedia.org/wiki/Qalb_(programming_language) should have it if possible. Captions, too, but they don't even have alignment yet so we can hold off on that (although we should do that).

I don't think math blocks need it.

If it's a lot, we can push with just text and iterate.

idreyn:
- I think we're covered on blockquotes because those contain paragraph elements which themselves get the RTL attribute. - Code blocks should be easy to update in the schema (although the value there is diminished since so much code is written LTR worldwide).
- I experimented a bit with HTML and CSS and I think we can add it to `bullet_list` and `ordered_list` to make the list bullets/numbers right-aligned, too.

gabestein:
Cool. I do think we should do code eventually. It's a nice gesture if nothing else, and who knows!

client/components/Editor/commands/rtl.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const createRtlCommandSpec = (direction: RtlDirection) => {
+	return createCommandSpec((dispatch: Dispatch, state: EditorState) => {
+		const rtlTargetNodes = getRtlTargetNodes(state);
+		const sharedDirection = doAllNodesHaveRtlDirection(rtlTargetNodes);
+		const sharedDirectionAttr = sharedDirection ? true : null;
+		const rtlAttrValue = swapRtlAttr(sharedDirectionAttr);
+		const canRun = rtlTargetNodes.length > 0;
+		return {
+			isActive: sharedDirectionAttr === direction,
```
idreyn, line 60: I'm really just nitpicking now but it might be more a e s t h e t i c to write out `canRun` and `isActive` in the same way: either create a variable or specify them inside the object.

idreyn (approved):
Looking incredibly clean! Fantastic work!

client/components/Editor/commands/rtl.ts
```diff
@@ -0,0 +1,80 @@
...
+type NodePos = { node: Node; pos: number };
+
+type RtlDirection = null | true;
+type AlignableNodesResult = {
+	rtlTargetNodes: NodePos[];
+	sharedDirectionType: RtlDirection;
+};
+const rtlAttr = 'rtl';
+
+const setRtlType = (rtl: RtlDirection): RtlDirection => (rtl === null ? true : null);
```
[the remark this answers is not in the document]
qweliant: yeah it's more of a swap from null to true

3mcd (approved):

qweliant:
> How complex is it to expand? Quotes and [code](https://en.wikipedia.org/wiki/Qalb_(programming_language)) should have it if possible. Captions, too, but they don't even have alignment yet so we can hold off on that (although we should do that).
> 
> I don't think math blocks need it.
> 
> If it's a lot, we can push with just text and iterate.

Issue #1633, opened 2021-10-21
Autogenerated PDF misteps (affiliation order, no link to pub)


Dawit-Tegbaru:
_Before filing your bug report, please search existing issues to make sure a similar one doesn't already exist. If one does, please leave a comment or reaction in support of the issue rather than creating a new one._

_To fill out the report, please leave the below headers (**) in tact, and fill out each section. If a section isn't applicable, leave it blank or write n/a under it. Please do not delete any template sections._

**PubPub displays author byline with one or multiple affiliations per author.** 

### What went wrong, step-by-step?
1. Autogenerated PDF incorrectly numbers affiliations when an author has multiple affiliations and another author shares at least one affiliation.  

For example: 
Martin Ingram1, Reuben Moreton2, Benjamin Gancz2, Frank Pollick3

1School of Psychology, University of Glasgow; Qumodo, London,
2Qumodo, London,
3School of Psychology, University of Glasgow

3. Autogenerated Pub displays DOI that isn't hyperlinked. Moreover, if a Pub doesn't have a DOI there is no URL that points readers to the VoR (Pub). 

...

### What did you expect to happen?
Martin Ingram1,2, Reuben Moreton2, Benjamin Gancz2, Frank Pollick1

1School of Psychology, University of Glasgow; Qumodo, London,
2Qumodo, London,
3School of Psychology, University of Glasgow

*Note in this example, Martin Ingram is affiliated w/University of Glasgow and Qumodo. 

DOI should be hyperlinked and point to Pub URL. If there is no DOI the Pub URL should be listed. 

### What URL can we use to see the issue?
https://tmb.apaopen.org/pub/calibrating-trust  (this issue is not specific to this pub or community - it's platform-wide)

### What browser/operating system were you using?
Chrome

===

### Have any screenshots? (Attach them by dragging the image here)
![image](https://user-images.githubusercontent.com/72410195/138190075-ddd04cfa-d61f-4288-8cff-51d5caa7fcd4.png)
![image](https://user-images.githubusercontent.com/72410195/138190338-77d21124-3593-4659-b900-2188d0c9d37c.png)

PR #2051, opened 2022-06-03
fix: handle multiple affiliations properly

fernandolguevara:
context #1633

### DIFF
--- a/workers/tasks/export/html.tsx
+++ b/workers/tasks/export/html.tsx
@@ -227,10 +227,14 @@ const renderFrontMatter = ({
 								<span className="name" key={index}>
 									{fullName}
 									{affs?.length > 0 &&
-										affs.map((affiliation) => (
-											<sup>{1 + affiliations.indexOf(affiliation)}</sup>
+										affs.map((affiliation, affIndex) => (
+											<sup key={affIndex}>
+												{1 + affiliations.indexOf(affiliation)}
+												{affs.length > 1 &&
+													affIndex < affs.length - 1 &&
+													','}
+											</sup>
 										))}
-									{index < attributions.length - 1 && ', '}
 								</span>
 							);
 						})}


gabestein:
Thanks so much for this quick fix @fernandolguevara! Looks great, there was a check for affiliation length I added so it wouldn't print a 0 by names with no affiliations. Adding one of our engineers now to take a final look.

If you're looking to make contributions in the future, please feel free to reach out to hello@pubpub.org or comment on an issue you'd like to take on. We don't have great contribution guidelines, but it's something we'd like to support more of. In general, we like to make sure PRs include:

- a high-level description of what the PR accomplishes
- screenshots/recordings of the differences if applicable/possible
- a test plan for other devs to follow

See, e.g.:  https://github.com/pubpub/pubpub/pull/2041

fernandolguevara:
@gabestein 

Can u change the Author 2 to be affiliated to `Knowledge Futures; Wassamatu U`?  and regenerate the PDF, please?

gabestein:
Ah, I see. Sorry I didn't test that. We'll have to have a chat about semicolons vs. commas as the separator. Ultimately, we plan to tokenize the field to support multiple institutions, so we'll want something that's fairly distinct to do a migration.

workers/tasks/export/html.tsx
```diff
@@ -206,17 +221,20 @@ const renderFrontMatter = ({
...
-								affiliation && affiliations.includes(affiliation)
-									? 1 + affiliations.indexOf(affiliation)
-									: null;
+							const affs = getAffiliations(attr);
 							return (
 								<span className="name" key={index}>
 									{fullName}
-									{affiliationNumber && <sup>{affiliationNumber}</sup>}
-									{index < attributions.length - 1 && ', '}
+									{affs?.length > 0 &&
```
idreyn, line 229: You could skip this check since if `affs.length === 0` the `map()` will just not render anything.

idreyn (approved):
Looks good to me, thanks for this change!

PR #2054, opened 2022-06-06
Enhancement: PDF and HTML links

gabestein:
In PDF and HTML exports: if no DOI is issued, adds a URL field to the export's cover page. If DOI issued, adds a link to the DOI. 

Resolves part of #1633 

Screen Shot 2022-06-06 at 15 47 18

Screen Shot 2022-06-06 at 15 47 30

_Test Plan_
1. Export PDF of pub without DOI.
2. Make sure link to Pub is in PDF cover page
3. Mint test DOI for pub
4. Make sure URL is replaced with DOI, with proper link (https://doi.org...)

### DIFF
--- a/utils/pubEdge/helpers.ts
+++ b/utils/pubEdge/helpers.ts
@@ -13,7 +13,7 @@ export const getHostnameForUrl = (url: string) => {
 	}
 };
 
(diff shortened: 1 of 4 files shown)


3mcd (approved):
LGTM!

Issue #2058, opened 2022-06-07
Add RTL to footnote, citaton, caption, page/layout editor rich text fields


gabestein:
### What would this enhancement help you accomplish?

---

idreyn:
Resolved by #2047

PR #2065, opened 2022-06-13
enhancement: Add RTL to Additional Editor Fields

qweliant:
addresses issue #2058

We recently added RTL to the main editor. This adds RTL directionality to other editors. 

test:
Create text in the layout editor, submission workflow, submission abstract, discussion, citation, and footnotes. Observe RTL persist when rendered

*also fixes some padding added to comments to the right caused by being a child of a prosemirror node

### DIFF
--- a/client/components/Editor/views/notePopover.scss
+++ b/client/components/Editor/views/notePopover.scss
@@ -8,8 +8,8 @@
 	p:last-child {
 		margin-bottom: 0;
 	}
-}
 
-[data-rtl='true'] {
-	direction: rtl;
+	[data-rtl='true'] {
+		direction: rtl;
+	}
 }
--- a/client/components/Editor/views/notePopover.scss
+++ b/client/components/Editor/views/notePopover.scss
@@ -9,3 +9,7 @@
 		margin-bottom: 0;
 	}
 }
+
+[data-rtl='true'] {
+	direction: rtl;
+}
--- a/client/containers/Pub/PubDocument/PubDiscussions/Discussion/ThreadComment.tsx
+++ b/client/containers/Pub/PubDocument/PubDiscussions/Discussion/ThreadComment.tsx
@@ -68,22 +68,18 @@ const ThreadComment = (props: Props) => {
 	};
 
 	const isAuthor = loginData.id === threadCommentData.userId;
-	const handleChangeEditorObject = (editorChangeObject: EditorChangeObject) => {
-		if (isEditing) {
-			setChangeObject(editorChangeObject);
-		}
-	};
-	const renderText = (
-		key: string, // key={`${isEditing}-${threadCommentData.text}`}
-		isReadOnly: boolean, // isReadOnly={!isEditing}
-		onChange?: any,
-	) => {
+
+	const renderText = () => {
 		return (
 			<Editor
-				key={key}
-				isReadOnly={isReadOnly}
+				key={`${isEditing}-${threadCommentData.text}`}
+				isReadOnly={!isEditing}
 				initialContent={threadCommentData.content}
-				onChange={onChange}
+				onChange={(editorChangeObject) => {
+					if (isEditing) {
+						setChangeObject(editorChangeObject);
+					}
+				}}
 			/>
 		);
 	};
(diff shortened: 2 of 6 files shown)


client/containers/Pub/PubDocument/PubDiscussions/Discussion/ThreadComment.tsx
```diff
@@ -68,6 +68,17 @@ const ThreadComment = (props: Props) => {
 	};
 
 	const isAuthor = loginData.id === threadCommentData.userId;
+	const renderText = (key: string, isReadOnly: boolean, onChange?: any) => {
```
idreyn, line 71: I believe `onChange` is of type `Callback<EditorChangeObject>`

client/components/Editor/views/notePopover.scss
```diff
@@ -9,3 +9,7 @@
 		margin-bottom: 0;
 	}
 }
+
+[data-rtl='true'] {
```
idreyn, line 13: ⚠️ This should be nested inside `.note-popover-component` so that it doesn't affect everything else in the webapp. Note that we _do_ put the `data-rtl` selector at the top level of `textAlign.scss` because that file is actually `@imported` into a selector in `base.scss`!

client/containers/Pub/PubDocument/PubDiscussions/Discussion/ThreadComment.tsx
```diff
@@ -84,7 +95,11 @@ const ThreadComment = (props: Props) => {
 						{isPreview ? ': ' : ''}
 					</span>
 
-					{isPreview && <span className="preview-text">{threadCommentData.text}</span>}
+					{isPreview && (
```
idreyn, line 98: ⚠️ It's a cool idea to try to make sure the previews have RTL but sadly I don't think it's going to work without a way to truncate the contents of the editor a little bit more.

image

Otherwise we lose the "preview" nature of these boxes and they just contain the entire comment text, which could potentially be very long. This looks okay with one comment but will become unreasonable once there are many. There are some strategies we could try to limit their height:

1. Using `-webkit-line-clamp` to set a maximum number of lines to show in the element — [see our `clamp-lines` mixin](https://github.com/pubpub/pubpub/blob/cf76b61f7decc94fb78e68483cb2256c746dc4f5/client/components/UserNotifications/userNotifications.scss#L13). Didn't work for me when I tried it just now but it could be worth messing with a little more.
2. Some kind of reasonably short `max-height` on the editor, plus a nice transparent-to-opaque gradient over the remainder of the text so it fades gently into that good night. I feel like I see this all over the internet but I can't find a decent example right now, sorry.

But honestly, my vote is to revert to using `threadComment.text` and accept that we'll have left-aligned RTL text in the comment previews for now.

idreyn (changes requested):
Thanks for handling this and for thinking carefully about a good RTL experience for comments. I think we're going to need to rethink using the entire comment text in the `ThreadComment` preview state — see my comment inline.

client/containers/Pub/PubDocument/PubDiscussions/Discussion/ThreadComment.tsx
idreyn: This looks great! Except, the username and the preview text are now misaligned by 1px:

image

but if you change `align-items` of the wrapping `.item-header` from `center` to `flex-start` it looks like that takes care of the problem:

image

client/containers/Pub/PubDocument/PubDiscussions/Discussion/ThreadComment.tsx
```diff
@@ -134,16 +154,15 @@ const ThreadComment = (props: Props) => {
...
 							/>
 						)}
-						<Editor
-							key={`${isEditing}-${threadCommentData.text}`}
-							isReadOnly={!isEditing}
-							initialContent={threadCommentData.content}
-							onChange={(editorChangeObject) => {
+						{renderText(
+							`${isEditing}-${threadCommentData.text}`,
+							true,
```
idreyn, line 159: I'm finding that I can't edit comments anymore and I'm wondering if you might want to pass `!isEditing` in here?

idreyn (changes requested):
Looks good —

PR #2066, opened 2022-06-14
Update: add helper body css class for user status

### LINKED ISSUE #2062: Add css classes for logged in vs. logged out, member or not
Add css classes to the body for when a user is logged in, logged out, or a community member.

Target custom CSS to community members, logged out, logged in.

---

### Who requested this? (If not you)
Arcadia

gabestein:
resolves #2062

Adds classes to the site body based on the user, to help people target custom CSS changes to different types of users.

If logged out: no `user-*` classes, can target with :not()
If logged in but not a member at a given scope: `user-logged-in`
If a member at a given scope: `user-logged-in` and `user-member` and `user-permission-[permission]`

_To test_
1. Logout and visit. Should see no `user-*` body classes.
2. Login as a member and visit. Should see `user-member` and `user-permission-[permission]` as a body class.
3. Login as a non-member and visit. Should see `user-logged-in` as a body class.
4. Add user as a member to a non-community-wide scope.
6. Visit that scope.
7. Should see `user-member` and `user-permission-[permission]` as a body class.
8. Visit another scope they're not a member of.
9. Should see `user-logged-in` as a body class.

### DIFF
--- a/server/Html.tsx
+++ b/server/Html.tsx
@@ -49,6 +49,16 @@ const getActiveSlugClassName = (initialData: InitialData, viewData?: any) => {
 	return '';
 };
 
+const getUserClassName = (initialData: InitialData) => {
+	if (initialData.scopeData.activePermissions.activePermission) {
+		return 'user-member';
+	}
+	if (initialData.loginData.id) {
+		return 'user-logged-in';
+	}
+	return 'user-anonymous';
+};
+
 const Html = (props: Props) => {
 	const { customScripts } = props;
 	const getPath = (chunkName, extension) => {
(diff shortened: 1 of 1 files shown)


3mcd (approved):
LGTM, just a couple quick questions.

Should we apply the `user-logged-in` and `user-member` classes progressively? Might be confusing to folks if `.user-logged-in` styles are no longer applied when a user happens to be a member at that scope. So if a user is logged in, both `user-logged-in` and `user-member` classes are applied.

If someone wants to target logged in users but _not_ members, they could use the `:not` pseudo-selector like `.user-logged-in:not(.user-member)`.

I also wonder if we should add classes for each membership role like `user-admin`, `user-edit`, etc.

3mcd (approved):
@gabestein Ahh looks great!

PR #2067, opened 2022-06-14
Update: deposit ORCID IDs to crossref

gabestein:
Very simple enhancement that will help a lot with Crossref support. Crossref will now automatically ask authors permission to add pubpub-deposited documents to their profiles.

_To test_
1. Create a pub with four users: a registered user with an ORCID in their profile, a registered user without one, a shadow author without an ORCID, and a shadow author with an ORCID.
2. Release & generate a deposit for the pub
3. Check the XML for the deposit: registered user with ORCID and shadow user with ORCID should have `<ORCID>https://orcid.org/ORCID-ID-NUMBER</ORCID>` as an attribute in the `<person>` record, per [Crossref](https://www.crossref.org/documentation/content-registration/descriptive-metadata/contributors/) docs. The users without ORCID should not have an `<ORCID>` attribute, as an empty one will cause a Crossref error.
4. Deposit to crossref test and login to https://test.crossref.org to verify it works.

Note that Crossref will error if the ORCID ID is not valid. This is expected.

### DIFF
--- a/utils/crossref/schema/contributors.js
+++ b/utils/crossref/schema/contributors.js
@@ -29,13 +29,16 @@ export default (attributions) => {
 						? attribution.user.lastName
 						: attribution.user.firstName,
 					affiliation: attribution.affiliation,
-					ORCID: attribution.user.orcid && `https://orcid.org/${attribution.user.orcid}`,
+					ORCID: `https://orcid.org/${attribution.user.orcid}`,
 				};
 				if (!personNameOutput.affiliation) {
 					delete personNameOutput.affiliation;
 				}
 				if (!personNameOutput.given_name) {
 					delete personNameOutput.given_name;
+				}
+				if (!attribution.user.orcid) {
+					delete personNameOutput.ORCID;
 				}
 				return personNameOutput;
 			}),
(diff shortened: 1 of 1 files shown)


server/Html.tsx
```diff
@@ -49,6 +49,16 @@ const getActiveSlugClassName = (initialData: InitialData, viewData?: any) => {
...
 
+const getUserClassName = (initialData: InitialData) => {
+	if (initialData.scopeData.activePermissions.activePermission) {
+		return 'user-member';
+	}
+	if (initialData.loginData.id) {
+		return 'user-logged-in';
+	}
+	return 'user-anonymous';
+};
```
3mcd, line 60: Looks like this may have snuck in accidentally.

3mcd (changes requested):

3mcd (approved):

3mcd:
Tested, works as expected. I got a strange error message that I hadn't seen before, though:

```ts
Error: cvc-pattern-valid: Value '/pub/lbzexuhq' is not facet-valid with respect to pattern '([hH][tT][tT][pP]|[hH][tT][tT][pP][sS]|[fF][tT][pP])://.*' for type 'resource_t'.
```

But the generated XML looked right w.r.t the `<ORCID>` elements.

PR #2072, opened 2022-06-16
feature?: "My Pubs" tab in overview

idreyn:
I've gotten sick of not having a place to find my own Pubs in a Community when I visit the dashboard, so I made one. It shows you all the Pubs in the Community that you're a Member of or a contributor to:

image

If we like this, do we like "Mine" or "My Pubs" better?

_Test plan:_
```
npm run test-dev server/pub
```
- Visit the Community overview and verify that the "Mine" tabs shows you only Pubs that you are a Member of or contributor to
- Do the same with a Collection overview
- Log out and check that neither the "Drafts" or "Mine" tabs are visible.


gabestein (commented):
Unable to test due to errors, but I think this is great. Prefer "My Pubs" I think, but don't feel strongly about it.

3mcd (approved):
LGTM!!

PR #2083, opened 2022-06-21
fix: Memoize `usePubContext()` contents while typing

idreyn:
Recall two quick facts about state management on the Pub page:

- We use the `PubSuspendWhileTyping` component to prevent major re-renders while the Pub receives text input
- We (increasingly) use the `usePubContext()` hook to access Pub state deep in the render tree

The problem is that `usePubContext()` basically makes `PubSuspendWhileTyping` because React re-renders whenever a Context value changes. So I've made these changes:

- Replace the `PubSuspendWhileTyping` wrapper component with a `usePubSuspendWhileTyping()` hook that memoizes a value it receives.
- Call `usePubSuspendWhileTyping()` from `usePubContext()` so that, by default, it passes a memoized value while the Pub is receiving input. This prevents a lot of unnecessary re-renders.

Since we do sometimes want to immediately receive the latest state, I added a second context provider that stores the most recent pub state, rather than the memoized one. You can access it by calling `usePubContext({ immediate: true })`. We do this in two places where immediate visual feedback from the editor is important:

- The Pub formatting bar
- The Pub table of contents

As a result of these changes, a typical render that occurs after a keypress now takes about half as long in my testing:

_Before:_
image

_After:_
image

_Test plan:_
Visit a Pub and make sure that:
- Broadly speaking, text input still works
- The `PubHeaderFormatting` immediately receives changes from the editor (try using Cmd+B to switch to bold text quickly as you type and verify that the Bold button in the toolbar is highlighted)
- The `PubToc` also receives changes immediately (try using the `#` shortcut to create a heading and verify that the Contents menu in the Pub header immediately appears)

### DIFF
--- a/client/containers/Pub/PubSuspendWhileTyping.tsx
+++ b/client/containers/Pub/PubSuspendWhileTyping.tsx
@@ -14,8 +14,6 @@ export const PubSuspendWhileTypingContext = React.createContext<PubSuspendWhileT
 	getTimeRemainingToUpdate: () => 0,
 });
 
-type ChildrenFn = () => React.ReactElement;
-
 type PubSuspendWhileTypingProviderProps = {
 	children: React.ReactNode;
 };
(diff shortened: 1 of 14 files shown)


client/containers/Pub/pubHooks.ts
```diff
@@ -1,22 +1,23 @@
 import { useContext } from 'react';
 
-import { PubContext } from './PubContextProvider';
+import { ImmediatePubContext, SuspendedPubContext } from './PubContextProvider';
 
-export const usePubContext = () => {
-	return useContext(PubContext);
+export const usePubContext = (options?: { immediate: boolean }) => {
+	const { immediate = false } = options ?? {};
```
qweliant, line 6: Is this written this way to always have a default negative value?

client/containers/Pub/PubSuspendWhileTyping.tsx
```diff
@@ -49,30 +47,14 @@ export const PubSuspendWhileTypingProvider = (props: PubSuspendWhileTypingProvid
...
-);
-
-type PubSuspendWhileTypingProps = {
-	children: ChildrenFn;
-	delay: number;
-};
-
-export const PubSuspendWhileTyping = (props: PubSuspendWhileTypingProps) => {
-	const { children, delay } = props;
+export const usePubSuspendWhileTyping = <T extends any>(value: T, delay = 1000): T => {
```
qweliant, line 50: Why use <T extends any> instead of <T extends PubContextType>?

idreyn: I'd say it's for separation of concerns: this hook works fine regardless of what kind of value you pass it, so there's no reason to more tightly bind it to `usePubContext()`.

The reason I wrote `<T extends any>` instead of `<T>` is much sillier. In a .tsx file, I guess the parser can't tell the difference between `<T>` the generic parameter and `<T>` as an opening JSX tag, so explicitly writing `<T extends any>` is just to help it out!

qweliant (approved):

3mcd (approved):
10/10 refactor. Looks great.

PR #2085, opened 2022-06-22
fix: Don't de-dupe two `AttributionWithUsers` with different roles

### LINKED ISSUE #2078: Editor not listed on "show details"

1. When clicking on "show details"
2. Not all of the editors set for the colection are displayed under "Contributors" -- Carmelo Ignaccolo is missing
...

All the attributions to be shown. 

https://projections.pubpub.org/pub/xusawg1f

===

Screen Shot 2022-06-20 at 4 56 28 PM

Projections editor: Carmelo Ignaccolo

idreyn:
Resolves #2078

Sometimes a user appears in both Collection and Pub-level attributions for a Pub. We tend to de-dupe these so the user only appears once, but in the case where e.g. a user is an editor at the Collection level and an author at the Pub level, we actually want them to appear twice, once for each role. So this change updates the de-duping logic to only filter duplicates that have the same set of roles.

Test plan:
- Create a Pub and add several attributions: a user `A` and several other attributions unaffiliated with a user.
- Check that all of these users appear on the byline, i.e. we are not de-duping unaffiliated attributions.
- Now add that Pub to a Primary Collection (public, non-Tag).
- Visit the Settings tab for that Collection and add `A` as Collection attribution.
- Return to the Pub. Click "Show Details" in the header and verify that `A` only appears once in the Contributors list.
- Now add a role to `A`'s Collection-level citation.
- Return to the Pub and verify that `A` is now shown twice in the Pub's Contributors list.

### DIFF
--- a/utils/contributors.ts
+++ b/utils/contributors.ts
@@ -1,7 +1,8 @@
+import { Attribution, AttributionWithUser, Collection, Pub } from 'types';
 import ensureUserForAttribution from 'utils/ensureUserForAttribution';
 import { joinOxford } from 'utils/strings';
-import { Attribution, AttributionWithUser, Collection, Pub } from 'types';
-import { getPrimaryCollection } from './collections/primary';
+import { unique } from 'utils/arrays';
+import { getPrimaryCollection } from 'utils/collections/primary';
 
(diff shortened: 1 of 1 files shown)


3mcd (approved):
LGTM (and thanks for the comment)

PR #2086, opened 2022-06-22
fix/update: Use our own CSL files and bump citation.js

### LINKED ISSUE #2077: Harvard-style citation

it seems that the Harvard-style citation isn't quite right, it looks like this:
 Porta, S. et al., 2022. Urban MorphoMetrics + Earth Observation: An integrated approach to rich/extra-large-scale taxonomies of urban form C. Ahn, C. Ignaccolo, & A. S. Miranda, eds. _Projections_, 16. Measuring the City: The Power of Urban Metrics. Available at: https://projections.pubpub.org/pub/vfwe0fdj.

the article title should be in quotes, the full issue name in italics, and the publisher be shown as well. at least this is what the editors say.

 Porta, S. et al., 2022. ‘Urban MorphoMetrics + Earth Observation: An integrated approach to rich/extra-large-scale taxonomies of urban form’, Ahn C., Ignaccolo C., & Salazar Miranda A., eds. _Projections, 16. Measuring the City: The Power of Urban Metrics._ MIT Press. Available at: https://projections.pubpub.org/pub/vfwe0fdj.

https://projections.pubpub.org/pub/vfwe0fdj

===

Projections editors, Carmelo Ignaccolo

gabestein:
Closes #2077.

This does a number of things to incrementally improve our citation handling. It:

- Updates citation.js to the latest version, which includes a bunch of bug fixes
- Adds our own versions of APA 6th Edition, Vancouver, and Harvard, so styles included in citation.js from the CSL repo that are modified upstream don't affect us
- "Fixes" the new imported Harvard style to get around [an issue with locales](https://github.com/citation-style-language/styles/issues/5956).
- Switches to using APA 7th Edition as the default for new pubs and for generating inline citations.
- Removes duplicate APA 6th/7th in the pub cite modal if an APA style is being used.

_Test plan_
- Visit an existing pub that uses APA 6th (basically any old Pub) and make sure it loads, that the style is set to APA 6th, and that it is in fact using APA 6th for the Pub cite menu and for the references list (telltale is it will add ellipses for authors > 6, e.g. `Dent, A., Beeblebrox, Z., Trillian, Prefect, F., McMillan, T., Prolonged, W. the I., … Stein, G. (2022). Hmm ok. Journal of KFG. https://doi.org/10.21428/0417b0c0.2ef9383d`).
- Switch the pub to APA 7th and make sure it updates as expected (will list all authors instead of using ellipses).
- Switch the pub to Harvard and make sure it works as expected in the cite menu and the references list (et al. once it reaches 4 authors, single quotes around titles, old-school doi prefix, e.g. `Dent, A. et al. (2022) 'Hmm ok', Journal of KFG [Preprint]. Edited by I. Reynolds. doi:10.21428/0417b0c0.2ef9383d.`
- Try another few styles just to be sure.
- Add a few new citations to the body and make sure they load with the set style.
- Flip the inline citation style to author-year or author and make sure it works as expected.
- Export the pub as a PDF and as a Word doc and make sure the expected citation styles work.

_A note for the future_
We should really overhaul the way we do inline styles to use citation.js's citation functionality rather than making our own, as styles have different inline style guidelines that citation.js supports.


gabestein:
ok I think this is ready to go

idreyn (approved):
LGTM!

PR #2088, opened 2022-06-23
fix: Alphabetize citations in HTML/PDF exports, where appropriate

### LINKED ISSUE #2082: Citation export not Alphabetical
1. When the Pub is exported (in PDF), the citations are numbered and not in alphabetical order, even though they are alphabetical in the citation footnote.
2. When the Pub is exported (in Word), the citations don't show up.
...

1. To have the exported citations match the footer (ideally, any repeated references are not duplicated -- this is the editor's request)
2. To be able to read the in-line citations.

https://projections.pubpub.org/pub/vfwe0fdj/release/1
--> or any in the community

===

For the Word references:
Screen Shot 2022-06-21 at 12 54 22 PM

Projections Editor, Carmelo Ignaccolo

idreyn:
Resolves #2082

Pubs have an "inline citation style" that dictates how in-text citation nodes should be displayed. Two common values are `count`, which looks like `[4]` and `author-year`, which looks like `(Deng et al., 2031)`. For styles besides `count`, readers expect to see citations in alphabetical order rather than by order of appearance. We implemented this behavior in #1349, but never added it to the separate codepaths that render citations for HTML/PDF exports. Until now!

Some notes-to-future-self: This PR is reminding me how much complexity there is around citation rendering in PubPub, and how much it needs to be streamlined. Basically there are four steps that happen in places all over the codebase:

1. Grabbing a set of `Note` objects with structured values (like a DOI) from somewhere, probably the Pub body with `getNotes()`.
2. Feeding those into citation-js to get `RenderedStructuredValues`, which maps these e.g. DOIs to rendered citations in HTML, CSL-JSON, etc. This must be done on the server. On the client, we have `NotesManager` to help manage the round-trip.
3. Merging the unstructured `Note[]` with `RenderedStructuredValues` to get `RenderedNote[]`.
4. De-duping, sorting, and numbering these rendered notes as appropriate.

For now I've introduced a new function called `renderAndSortNotes()` that handles these last two. But it's becoming clearer to me that we should probably have a helper class of some kind that manages this whole process. I think it could compose `NoteManager` (which should be renamed, ahem, `RenderedStructuredValuesManager`) and also manage steps 1, 3, and 4.

 _Test plan:_
Add two citations to a new Pub in reverse-alphabetical order by first author's surname. You can use [this one](https://doi.org/10.1073/pnas.2102466119) and [that one](https://doi.org/10.1016/j.cstp.2017.08.007)

- They should appear in the Pub-bottom Citations section in that order, with numbering.
- The backlinks on the citations should work.
- Add a _footnote_ with structured data. It should appear in the footnotes section.
- Adding unstructured (rich text) data to citations and footnotes should work.
- Refreshing the page should not crash the Pub.

Now visit Pub Settings and change the inline citation style to `author-year`.

- The citations should now be rendered in author-year form.
- The Pub-bottom Citations section should list them alphabetically.
- They should not be numbered.

Now create a JATS XML export of the Pub.

- Verify that we're still linking citations correctly. Each inline citation should create an `xref` element with an ID that corresponds to a richly-populated `ref` element in `ref-list`.

Now create a PDF (or, hell, HTML) export of the Pub.

- Check that the citations at the bottom appear in a bullet list, in alphabetical order.
- Restore the `Count` inline citation style. Check that they appear in an ordered list, as usual.

_Cool, huh:_

### DIFF
--- a/utils/notes.ts
+++ b/utils/notes.ts
@@ -1,0 +1,88 @@
+import { CitationInlineStyleKind } from './citations';
+
+export type Note = {
+	id: string;
+	structuredValue: string;
+	unstructuredValue: string;
+};
+
+export type RenderedNote = Note & {
+	renderedStructuredValue?: RenderedStructuredValue;
+	number?: number;
+};
+
+export type StructuredValue = string;
+
+export type RenderedStructuredValue = {
+	html: string;
+	inline: string;
+	error?: boolean;
+	json: any;
+};
+
+export type RenderedStructuredValues = Record<StructuredValue, RenderedStructuredValue>;
+
+export type ByNoteKind<T> = {
+	footnotes: T;
+	citations: T;
+};
+
+const getAlphabetizableValue = (note: RenderedNote) => {
+	const hasRenderedStructuredValue = note.renderedStructuredValue?.json?.length;
+	const sortableString = hasRenderedStructuredValue
+		? note.renderedStructuredValue?.html
+		: note.unstructuredValue;
+	return (sortableString || '')
+		.replace(/(<([^>]+)>)/gi, '')
+		.toLowerCase()
+		.trim();
+};
+
+const getRenderedNotes = (
+	notes: Note[],
+	renderedStructuredValues: RenderedStructuredValues,
+	includeNumbers: boolean,
+): RenderedNote[] => {
+	return notes.map((note, index) => {
+		return {
+			...note,
+			...(includeNumbers && { number: index + 1 }),
+			renderedStructuredValue: renderedStructuredValues[note.structuredValue],
+		};
+	});
+};
+
+const alphabetizeCitations = (citations: RenderedNote[]) => {
+	return citations
+		.map((citation) => {
+			return {
+				...citation,
+				sortOn: getAlphabetizableValue(citation),
+			};
+		})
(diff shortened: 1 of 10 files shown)


utils/notes.ts
```diff
@@ -0,0 +1,88 @@
+import { CitationInlineStyleKind } from './citations';
```
idreyn, line 1: The logic in this file should be identical to what I removed from `usePubNotes()`. I factored it out so we can use it in the export pipeline as well.

utils/notes.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export type Note = {
```
idreyn, line 3: I moved this type from `components/Editor` because /shrug

utils/notes.ts
```diff
@@ -0,0 +1,88 @@
...
+	structuredValue: string;
+	unstructuredValue: string;
+};
+
+export type RenderedNote = Note & {
+	renderedStructuredValue?: RenderedStructuredValue;
+	number?: number;
+};
+
+export type StructuredValue = string;
```
idreyn, line 14: I moved the next few types over from next door in `utils/notesCore.ts`

3mcd (approved):
LGTM!

PR #2101, opened 2022-06-28
Fix: deprecated client sass, blueprint sass, bump mudder

howebenhowe:
For those of us tired of getting the `dart-sass` deprecation warning on `/` division and the `mudderjs` warning on bum `package.json` style.

Used `sass-migrator` package against both the pubpub scss and the `@blueprintjs/core` package, then saved the changes to `@blueprintjs/core` using `patch-package` package. Also, optimistically adds `quietDeps: true` to dev webpack config, though it seems like it's not working yet.


howebenhowe:
https://github.com/sass/dart-sass/issues/672

PR #2106, opened 2022-07-01
fix: prevent double doi field in Layout header

tefkah:
The DOI field gets printed twice in the LayoutCollectionHeader, this PR filters out the DOI in the second printing.

There's probably better ways to do this, but I'll leave that up to your discretion!

image

This is related to https://github.com/pubpub/pubpub/discussions/1565#discussioncomment-3063822


idreyn (approved):
Nice, thanks!

PR #2132, opened 2022-07-18
feat?: change citations to references and demote read next

gabestein:
Per Web of Science's review of HDSR, citations should be called "references" (and I agree, this is the correct language -- citations are the actual in-text things), and the "read next" module should be after the license info so as not to confuse people into thinking its related to the metadata for that pub (¯\_(ツ)_/¯).

Anyhow, this does that.

_Test plan_
1. Visit any pub with citations in a collection with multiple pubs
2. Make sure the bottom section says "references" not "citations"
3. Make sure read next appears below the license section
4. Make a change and download a PDF export of the pub
5. Make sure that export says "references" at the bottom of the PDF, not "citations"


gabestein:
Whoops, I fixed it anyway. Just one line related to spacing.

PR #2139, opened 2022-07-20
fix: unbreak pandox math node handling, bump prosemirror-pandoc

howebenhowe:
Fixes bug [pointed out here](https://github.com/pubpub/pubpub/discussions/1869#discussioncomment-3186517) where importing from markdown wasn't working so well.

_test plan_

+ write a pub with both inline and display math nodes
+ export that pub to md
+ import the md
+ check for consistency
+ repeat above steps for LaTeX filetype


idreyn (approved):
Ship it!

PR #2143, opened 2022-07-25
update: Add HTML title editor to Pub Settings

idreyn:
In order to launch rich text titles for all of PubPub, we need to add this capability to Pub Settings. This turns out to be pretty easy! Adding the `.bp3-input` class to the `TitleEditor` component's rendered div styles it perfectly. The only other substantial change I made was to add an `onInput` callback to the `TitleEditor` so we can get changes on keypress instead of blur, and a new check in the Pub PUT handler to convert `htmlTitle` to `title` by stripping tags.

_Test plan:_
- Visit Pub Settings for a Pub and verify that you can use bold and italics in the title, and persist those changes. The title at the top of the dashboard won't change, but that's an existing issue.
- Also make sure the `TitleEditor` in the Pub header continues to work.

### DIFF
--- a/client/components/TitleEditor/TitleEditor.tsx
+++ b/client/components/TitleEditor/TitleEditor.tsx
@@ -10,6 +10,7 @@ type Props = {
 	isReadOnly?: boolean;
 	initialValue?: string;
 	onChange?: (html: string, text: string) => void;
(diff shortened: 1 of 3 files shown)


3mcd (approved):
LGTM!!!

PR #2144, opened 2022-07-26
fix: use right key for issn meta display

gabestein:
Fixes metadata for things that have ISSNs so it actually shows up.

_To test_
1. Visit a pub that's in a journal issue collection that has an ISSN (electronic or print) entered in the settings.
2. Open the inspector and make sure the `<meta name="citation_issn" content="XXXX-XXXX" />` tag displays properly, with the ISSN you entered in step 1.

### DIFF
--- a/server/utils/ssr.tsx
+++ b/server/utils/ssr.tsx
@@ -137,9 +137,9 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 				<meta
 					key="c4"
 					name="citation_issn"
-					content={collection.metadata?.electronic_issn}
+					content={collection.metadata?.electronicIssn}
 				/>,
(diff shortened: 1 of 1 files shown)


qweliant (approved):
i would change it to so you dont have a blank line reading `<meta name="citation_issue">`

```
const issn = collection.metadata?.electronicIssn || collection.metadata?.printIssn;

...
<meta key="c4" name="citation_issn" content={content} />,
```

gabestein:
Thanks! Every meta tag operates like this because of the declarative way they were built, so I think I'm gonna leave it this way for now until we rebuild.

Also, I think it's okay (maybe even good) to have both the electronic and print issn if they both exist.

PR #2149, opened 2022-07-27
fix: De-dupe notes in `renderNotesForListing()` instead of `getNotesByKind()`

### LINKED ISSUE #2146: Error generating DOCX/EPUB/LaTeX/JATS/Plaintext/OpenDocument download in some Pubs

1. Try to download generated DOCX/EPUB/LaTeX/JATS/Plaintext/OpenDocument in some files
2. image

The weird thing is that this doesn't happen all the time, I couldn't narrow down what was causing the issue. See below for working/not working urls.

The Network tab shows requests to a worker (see below), which does not seem to happen for "successful" downloads, which shows the following error (prettified)

```json
{
  "id": "c439234c-424f-4234-9daf-66c5e0880a33",
  "isProcessing": false,
  "error": "Cannot destructure property 'id' of 'resources.note(...)' as it is undefined.",
  "output": null
}
```

I could download the file.

Working:
- https://openpresstiu.pubpub.org/pub/kritisch-wetenschappelijk-1/release/1?readingCollection=19a3c21e
- https://openpresstiu.pubpub.org/pub/kritisch-wetenschappelijk-0/release/1?readingCollection=19a3c21e

Not working:
- https://openpresstiu.pubpub.org/pub/kritisch-wetenschappelijk-4/release/1?readingCollection=19a3c21e
- https://openpresstiu.pubpub.org/pub/kritisch-wetenschappelijk-2/release/1?readingCollection=19a3c21e

-

see above

image

idreyn:
Resolves #2146

This fixes a problem where identical `citation` nodes in a Pub document are de-duped "too early" in `getNotesByKind()`, which prevents them from being processed by `getPandocNotesById()`. We want to de-dupe a list of notes only shortly before displaying it to users, so I've moved that functionality into `renderNotesForListing()`.

_Test plan:_
- Create a Pub with several identical citation nodes (feel free to copy [this one](https://openpresstiu.pubpub.org/pub/kritisch-wetenschappelijk-4/release/1?readingCollection=19a3c21e) from the linked bug report)
- Make sure that both Pandoc (JATS, Markdown, ...) and non-Pandoc (HTML) exports work propertly
- Verify that identical citations are de-duped properly in both the Pub footer and HTML exports (easiest to see when they are alphabetized by turning on Author-Year citations style)

### DIFF
--- a/client/components/Editor/utils/notes.ts
+++ b/client/components/Editor/utils/notes.ts
@@ -5,41 +5,34 @@ import { Note } from 'utils/notes';
 
 export type CitationFingerprintFunction = (node: Node) => string;
 
-const citationFingerprintStripTags = (node: Node) => {
+const fingerprintNote = (node: Node) => {
 	const { unstructuredValue, value } = node.attrs;
 	const strippedUnstructuredValue = unstructuredValue ? striptags(unstructuredValue) : '';
 	return `${value}-${strippedUnstructuredValue}`;
 };
 
-export const getNotes = (
-	doc: Node,
-	citationFingerprintFn: CitationFingerprintFunction = citationFingerprintStripTags,
-) => {
-	const citationFingerprints = new Set<unknown>();
+export const getNotesByKindFromDoc = (doc: Node) => {
 	const footnoteItems: Note[] = [];
 	const citationItems: Note[] = [];
 
 	doc.nodesBetween(0, doc.nodeSize - 2, (node) => {
+		const fingerprint = fingerprintNote(node);
 		if (node.type.name === 'footnote') {
+			const { value, structuredValue } = node.attrs;
 			footnoteItems.push({
 				id: node.attrs.id,
-				structuredValue: node.attrs.structuredValue,
-				unstructuredValue: node.attrs.value,
+				unstructuredValue: value,
+				structuredValue,
+				fingerprint,
 			});
 		}
 		if (node.type.name === 'citation') {
 			const { unstructuredValue, value } = node.attrs;
-			if (citationFingerprintFn) {
-				const citationFingerprint = citationFingerprintFn(node);
-				if (citationFingerprints.has(citationFingerprint)) {
-					return true;
-				}
-				citationFingerprints.add(citationFingerprint);
-			}
 			citationItems.push({
 				id: node.attrs.id,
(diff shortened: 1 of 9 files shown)


client/components/Editor/utils/notes.ts
```diff
@@ -5,41 +5,34 @@ import { Note } from 'utils/notes';
...
 	const strippedUnstructuredValue = unstructuredValue ? striptags(unstructuredValue) : '';
 	return `${value}-${strippedUnstructuredValue}`;
 };
 
-export const getNotes = (
-	doc: Node,
-	citationFingerprintFn: CitationFingerprintFunction = citationFingerprintStripTags,
-) => {
-	const citationFingerprints = new Set<unknown>();
+export const getNotesByKindFromDoc = (doc: Node) => {
```
idreyn, line 14: Renamed this function because the existing name was hopelessly vague.

utils/notes.ts
```diff
@@ -63,14 +65,16 @@ const alphabetizeCitations = (citations: RenderedNote[]) => {
...
-type RenderAndSortOptions = {
+type RenderForListingOptions = {
 	citations: Note[];
 	footnotes: Note[];
 	citationInlineStyle: CitationInlineStyleKind;
 	renderedStructuredValues: RenderedStructuredValues;
 };
 
-export const renderAndSortNotes = (options: RenderAndSortOptions): ByNoteKind<RenderedNote[]> => {
+export const renderNotesForListing = (
```
idreyn, line 75: Also renamed this one to better emphasize what it's for.

workers/tasks/export/notes.ts
```diff
@@ -88,10 +87,6 @@ export const getPandocNotesByHash = (notesData: NotesData): PandocNotes => {
 	return index;
 };
 
-export const getCslJsonForPandocNotes = (notes: PandocNotes) => {
```
idreyn, line 91: I just inlined this at its only callsite because it felt like it wasn't really pulling its weight as a standalone fn.

gabestein (approved):
Thanks for the quick fix here. From a QA perspective, works perfectly.

One small thing we should do later is, in HTML/PDF, include multiple links "back" to the citation from the references section if there are duplicated cites.

PR #2150, opened 2022-07-27
fix: Better inputRules for inline math (and one for code)

### LINKED ISSUE #2141: Math appears when trying to write about US dollars

1. Whenever there are two instances of USD ($) mentioned in a Pub,
2. The text in-between turns into LaTex-looking mathematics.
...

Just the $ sign to appear and leave the text in-between alone. Copying and pasting from other places doesn't seem to cause this, but typing in a $ will cause the issue.

https://help.pubpub.org/pub/08907yo3/draft

Google Chrome/BigSur macOS

===

Screen Shot 2022-07-25 at 12 10 47 PM

idreyn:
Resolves #2141

We caught a few issues with the input rule that automatically converts `$ this $` to inline math:

1. It fires on sentences involving money ("$7.00 to $9.00")
2. It fires inside of inline code marks, which makes it hard to discuss how to use the feature

This PR fixes these issues. I have also added an equivalent rule for turning single backticks into inline code (finally!)

_Test plan:_

Okay, my brain is too fried to write all this out. But here is some stuff you can try to replicate. Let me know you see anything funky. Note that the fourth line down will trigger a new problem because we no longer match `$ this when it has a TeX dollar sign ($) in it $`. I think this is a tractable problem with clever use of regex, but I'm going to punt on it for now. 

image

### DIFF
--- a/client/components/Editor/plugins/inputRules.ts
+++ b/client/components/Editor/plugins/inputRules.ts
@@ -1,3 +1,5 @@
+import { Fragment, MarkType, NodeType } from 'prosemirror-model';
+import { EditorState } from 'prosemirror-state';
 import {
 	inputRules,
 	wrappingInputRule,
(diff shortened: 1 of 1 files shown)


3mcd (approved):
Managing math nodes feels a lot more natural now. Loving the backticks too. LGTM!

howebenhowe (commented):
LGTM! (belatedly)

PR #2158, opened 2022-08-01
dev: Remove giant Google and AWS SDKs in favor of smaller ones

idreyn:
I saw [this article](https://effectivetypescript.com/2022/07/30/treemap-for-source-files/) today and was immediately sniped! Its author noted that the `googleapis` Node package was adding tons of (mostly unused) source files that the TypeScript compiler has to snoop through when compiling, slowing things down. He was able to speed up his TS builds by 20% simply by removing this package and replacing it with a smaller subset of functionality. 

Does PubPub have this problem? Well, I ran the provided command to find out:

```zsh
npx tsc --noEmit --listFiles | xargs stat -f "%z %N" | npx webtreemap-cli
```

And behold, we have _two_ monstrous FAANG (MAGMA?) SDKs that we mostly don't use, `googleapis` and `aws-sdk`:

Screen Shot 2022-08-01 at 12 18 47 PM

So...`googleapis` we only use in some one-off scripts, and can be trivially replaced with `google-auth-library`. `aws-sdk` is only used to work with S3, so I swapped that out for the much smaller `@aws-sdk/s3-client`, which unfortunately has a new API. Updating to use a (unified PubPub wrapper over) this API accounts for the bulk of the changes in this PR.

After swapping out these two packages, the situation is much better:

Screen Shot 2022-08-01 at 1 38 46 PM

In practice, this brings `npx tsc` from 22s to 17s. Hopefully this will add up over time.

_Test plan:_
Our test plan is focused on the S3 API because we don't use the swapped-out Google authentication API in production.

Note that there should be no possible change to client-side AWS uploads (importing media via the editor) because this uses the AWS HTTP API directly.

**Imports**
We use the S3 API in two places when importing files:

1. _To upload media extracted from binary documents (e.g. docx)._ To test this, add an image to a `.docx` or `.odt` file, import it, and verify that the image appears in the Pub.
2. _To download media that the client-side importer UI has already sent to AWS._ To test this, create a `.md` file which references an image and put them both into the uploader. Make sure the image appears in the Pub.

**Exports**
The export pipeline ends by sending the exported file to S3. Just test that a few formats work.

**Sitemaps**
We store sitemaps in S3 as gzipped XML files.
1. Force `shouldGenerateSitemapIndex()` to return `true`
2. Visit `/sitemap.xml` and verify that everything runs to completion. You should be able to follow the links within the returned XML.
3. Undo your override in `shouldGenerateSitemapIndex()` and verify that sitemap re-generation does not occur, because this function was able to read the `LastModified` header from the sitemap index just uploaded to S3.

**Backups**
We're just going to have to see how this one goes :)


3mcd (approved):
Great work, a few seconds improvement is huge!

workers/tasks/import/assetStore.ts
```diff
@@ -19,55 +17,34 @@ export const generateAssetKeyForFile = (filePath) => {
...
+			}
+			// eslint-disable-next-line no-await-in-loop
+			await sleep(1000);
+		}
+		return reject();
 	});
 
 export const downloadFileFromAssetStore = (assetKey: string, filePath: string) =>
 	new Promise(async (resolve, reject) => {
 		await blockForAssetFile(assetKey);
```
howebenhowe, line 36: The refactor looks great to me overall, but I don't understand what `blockForAssetFile()` does. Looks like it deliberately slows things down if the asset file in question doesn't exist or else if the assets client takes a long time to check for the presence of the file. But why would it take a long time? And why would you only want to slow it down, not error out, or some other more complete shut-down?

howebenhowe (approved):
Besides making the build more efficient, there's a good measure of tidying up in here. Nice PR, Ian!

Only question was on what I should make of the `blockForAssetFile` function, but that's for edification, not a blocker at all. Shipit!

Issue #2137, opened 2022-07-19
math editor followups


gabestein:
- make math nodes referenceable/linkable
- adjust css to account for different wrappers between old and new math, new math too big in both editor and export
- some errors when there is only a math node on the page
- math not importing/exporting via pandoc, see: https://github.com/pubpub/pubpub/discussions/1869#discussioncomment-3186517

howebenhowe:
dollar signs give trouble, for example when you click to edit [on this pub](https://notes.knowledgefutures.org/pub/q8gh8lv2/draft)

also, whatever is going on [here](https://help.pubpub.org/pub/08907yo3/draft)

gabestein:
one more: undo/redo don't seem to work within math blocks themselves (only after exiting the block, you can command-z to restore the old one0

PR #2159, opened 2022-08-01
Fix: linkable, referenceable math nodes

howebenhowe:
Restores linking functionality to display math nodes.
Addresses one of the bugs identified in #2137.

_Test plan_
+ go to a [pub with a display math node](http://localhost:9876/pub/m9xhcfz8/draft)
+ copy the link to that node
+ paste the link in new tab
+ should be navigated to that math node


client/components/Editor/schemas/equation.tsx
```diff
@@ -41,16 +41,6 @@ export default {
 		inline: true,
 		group: 'inline',
 		draggable: false,
-
```
idreyn, line 44: thank you!!

idreyn (approved):
LGTM!

PR #2163, opened 2022-08-04
Move DOI prefixes and Crossref credentials to database

3mcd:
This PR moves hard-coded DOI prefixes from [our code](https://github.com/pubpub/pubpub/blob/master/utils/crossref/communities.js)—and Crossref credentials from our environment variables—into a new database model called `DepositTarget`.

**Test Plan**

PubPub prefix
- Create a DOI for a Pub on duqduq. It should have the prefix `10.21428`
- Deposit the Pub. It should work!

Managed Prefix
- Log in to a community that has a custom DOI prefix, like Projections
- Go to a Pub settings dashboard page, e.g. `/dash/pub/2w805oes/settings` (Projections)
- The DOI prefix in the **DOI** section should match that community's custom prefix, e.g. `10.1162` (Projections), and the Pub should be depositable

Unmanaged Prefix
- Visit my test community `mcd` (this community has a custom prefix but no DepositTarget credentials)
- Create a Pub and publish a release
- Go to the Pub settings dashboard Page
- The Pub should not be depositable with the warning `Ths Pub cannot be deposited to Crossref because this Community's DOI prefix is not managed by PubPub.`


gabestein (changes requested):
Mostly looking good! Found a few things:

- Quick copy change on the "unmanaged" warning as noted inline.

- Bug: for managed communities with their own prefix, the "generate suffix" button no longer generates a suffix in the pattern we use. Instead, it seems to just add in the set prefix. Seems to work fine for "unmanaged" communities, though?

- Bug: typing in a new suffix and clicking "update suffix" does not save the custom suffix, instead it seems to save the prefix.

- It looks like unmanaged communities that have a prefix set aren't working as expected. Or perhaps the DB backfill wasn't quite right. E.g. for TMB, which is currently an "unamanged custom prefix" or whatever we want to call it, it uses the PubPub prefix, does not allow suffix customization, and allows depositing. MCD works, though! So probably just in the db?

- For unmanaged communities with custom prefixes, let's remove the "A DOI has been assigned to this Pub, but it has not yet been deposited to Crossref." warning on saving a custom suffix, since it's not particularly useful in that instance where we're not managing deposits.

- Maybe a thing?: On build on local, I saw this error pop up for Projections, stemming from crypto.js: `[0] TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined`

gabestein (changes requested):
Thanks for making those changes, looks good from that perspective.

Testing HDSR on dev (a fully-managed, custom-prefix community), I am getting the following 500 error when attempting to deposit a new Pub or redeposit an existing:

`<pre>AssertionError<br> &nbsp; &nbsp;at assert (/Users/gabe/code/pubpub/pubpub/dist/server/utils/assert.js:13:19)<br> &nbsp; &nbsp;at expect (/Users/gabe/code/pubpub/pubpub/dist/server/utils/assert.js:19:5)<br> &nbsp; &nbsp;at /Users/gabe/code/pubpub/pubpub/dist/server/server/doi/submit.js:29:85<br> &nbsp; &nbsp;at Generator.next (&lt;anonymous&gt;)<br> &nbsp; &nbsp;at fulfilled (/Users/gabe/code/pubpub/pubpub/dist/server/server/doi/submit.js:5:58)</pre>`

No issues on demo (PubPub's prefix).

gabestein (approved):
Nope, it was me not updating config!

LGTM!

idreyn (approved):
Awesome, LGTM!

PR #2180, opened 2022-08-11
hack: Add extra error handling (and patched prosemirror-math) to recover from bad chars in KaTeX

idreyn:
See https://github.com/KaTeX/KaTeX/issues/3680 for a detailed writeup of our problem. I'm likely to propose a patch to KaTeX at some point, but to keep PubPub afloat we make the following changes here:

1. Fork `@benrbray/prosemirror-math` so that it catches all errors that KaTeX could throw (our changes [here](https://github.com/pubpub/prosemirror-math/commit/ccf5737f75485925a25515818a1516ee59e76f90))
2. Write our own wrapper around KaTeX that also catches all errors

_Test plan:_

- Add a suspicious Unicode character like `𝛾` to a math node
- Make sure it doesn't immediately brick the Pub
- Refresh and make sure it doesn't brick the server-side render of the Pub

### DIFF
--- a/client/components/Editor/schemas/math.tsx
+++ b/client/components/Editor/schemas/math.tsx
@@ -1,5 +1,6 @@
 import { DOMOutputSpec, Node } from 'prosemirror-model';
-import katex from 'katex';
+
+import { renderToKatexString } from 'utils/katex';
 
 import { counter } from './reactive/counter';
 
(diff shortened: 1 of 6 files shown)


howebenhowe (approved):
Does what it says on the tin.
Shippit.

PR #2185, opened 2022-08-15
enhancement: show password reset error msg

howebenhowe:
When a user requests a password change for an invalid email (one that doesn't correspond with anything in the database) we should give that information to the user rather than a general 'error' message that might lead them to keep trying the same username.

_Test plan_
try submitting an invalid email, observe error message below


gabestein:
Will leave it to people with more expertise, but I'm not sure I agree we should do this. Some sites have been leaning away from this as it technically allows an attacker to fish for emails that do have accounts, and use that information. Instead, we may want to build a "forgot username" feature, and then default the password response in all cases to something like: "If the email address you entered has a valid PubPub account, you will receive an email with instructions for resetting your password."

PR #2192, opened 2022-08-18
Fix: actual first or primary

gabestein:
Actually grabs the first or primary collection, and in so doing moves us closer to doing it all in ScopeGet.

_To Test_
1. Visit a pub where the primary collection is not the first one in the ordered list and make sure the primary still shows as the collection in the dash menu.
2. Visit a pub where there is no primary collections and all other collections are ordered in non-alpha order and make sure the first ordered collection shows in the dash menu.

### DIFF
--- a/client/components/ScopeDropdown/ScopeDropdown.tsx
+++ b/client/components/ScopeDropdown/ScopeDropdown.tsx
@@ -219,7 +219,9 @@ const ScopeDropdown = (props: Props) => {
 											)}
 											<a
 												href={`/${
-													(scope.slugs && 'pub/' + scope.slugs.pubSlug) ||
+													(scope.slugs &&
+														scope.slugs.pubSlug &&
+														'pub/' + scope.slugs.pubSlug) ||
 													(scope.slugs && scope.slugs.collectionSlug)
 												}`}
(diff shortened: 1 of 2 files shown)


client/components/ScopeDropdown/ScopeDropdown.tsx
```diff
@@ -63,10 +65,7 @@ const ScopeDropdown = (props: Props) => {
 	const { canManageCommunity, canManage } = scopeData.activePermissions;
 	const collectionSlug = locationData.params.collectionSlug || locationData.query.collectionSlug;
 	const pubSlug = locationData.params.pubSlug;
-	const nonActiveDashboardCollectionPub = getPrimaryOrFirstCollectionPub(
-		activePub,
-		communityData,
-	);
+	const nonActiveDashboardCollectionPub = getPrimaryOrFirstCollection(activePub, communityData);
```
idreyn, line 68: Given how much work this is doing, would you mind throwing it into a `useMemo()`?

idreyn (approved):
LGTM but one non-blocking request inline — let's avoid calling `.sort()` and `.find()` on every render.

PR #2199, opened 2022-08-22
fix: add nodetype patch for referencing block equations

howebenhowe:
Makes block equations reference-able.

_Test plan_

1. Go to/create a pub with a math node
2. reference that math node
3. sha-bayum


idreyn (commented):
TFTF, LGTM!

idreyn (approved):

PR #2221, opened 2022-09-01
fix: math in captions

3mcd:
Quick fix for math in captions.

**Test Plan**

- Add math to a caption. It should render correctly in the editor.


idreyn (approved):
LGTM. If it seems like this is too slow on each keystroke we can pass a default `debounceEditsMs` into the `MinimalEditor` that this component renders.

PR #2222, opened 2022-09-01
Facets (1/x): Core functionality and database models

idreyn:
Well, here it is. This PR has the first work related to facets/cascading Pub settings:

### The facets "core" code (in `facets/`)
This is logic for defining, instantiating, validating, and cascading facets and their props. Compared to most of our code, this is _very_ type-heavy. A lot of the utility types are necessary for making sure type information is propagated properly so we get good code completion for facets stuff.

### Sequelize bindings for facets (in `server/facets`)
Some very basic code for automatically creating database models from facet definitions. Each of these models will have a field for every prop, plus a `facetBindingId` field. This associates the model with a new `FacetBinding` model, which associates each facet model to exactly one scope:

```
{ PubHeaderTheme, NodeLabels, ... } -> FacetBinding -> { Pub, Community, Collection }
```

Eventually this model could be extended to hold more information about a single facet instance's relationship to a scope, e.g. whether it has been released or not.

### Some stray code for running PubPub with a local database

This hasn't been fully systematized, but you can now start PubPub locally with the `USE_LOCAL_DB=true` and it will automatically create and bind to a local Postgres database instead of the duqduq one.

### Migrations to move data from the Pub model to facets

My plan is to land this PR on master as soon as it's approved, and then start running these migrations now. The migration writes to a temporary `facetsMigratedAt` field on the Pub model, which we can compare against `updatedAt` to incrementally do the migration without repeating work. In a few weeks when the rest of the code goes live, the migration will be done for all but the most recent/active Pubs.

<br>

Lot more to say here — I'll drop some comments in the code for more context.


facets/core/prop.ts
```diff
@@ -0,0 +1,81 @@
...
+	// The value of this prop at the "root" of PubPub -- in the absence of any other facet instances
+	// that override it, what value does this prop have?
+	rootValue: RootValue;
+	// The default value of this prop, if not explicitly provided to createFacet(). By contrast with
+	// `rootValue`, this value will actually be stored on a facet instance.
+	defaultValue?: DefaultValue;
+	// Explains how this prop should cascade from higher to lower scopes.
+	cascade?: AvailableCascadeStrategyForPropType<PropType>;
+	// A human-readable label for this prop
+	label?: string;
```
3mcd, line 36: Thanks for these comments! Also excellent typings in general all throughout this PR.

facets/core/prop.ts
```diff
@@ -0,0 +1,81 @@
...
+};
+
+export type FacetProp<
+	PropType extends FacetPropType = FacetPropType,
+	RootValue extends NullableTypeOfPropType<PropType> = NullableTypeOfPropType<PropType>,
```
3mcd, line 41: I love how easy this reads. But I do sometimes prefer generic type parameters shortened to a single letter or otherwise distinguished from module-level types.

facets/core/prop.ts
```diff
@@ -0,0 +1,81 @@
...
+	Prop['propType']
+>;
+
+export type CascadedTypeOfFacetProp<Prop extends FacetProp> =
+	// If the root value of the prop is non-null...
+	IsNeverNull<RootValueOfFacetProp<Prop>> extends true
+		? // ...then we'll never get null as a cascaded value for this prop
+		  TypeOfFacetProp<Prop>
+		: // Otherwise, we might.
+		  NullableTypeOfFacetProp<Prop>;
```
3mcd, line 64: Does this mean that intermediate scopes (i.e. collection) won't be able to nullify facet prop values?

init.js
```diff
@@ -2,15 +2,21 @@
 const {
 	argv: { watch },
 } = require('yargs');
-const throng = require('throng');
```
3mcd, line 5: Smell ya later

server/database/postgres.ts
```diff
@@ -0,0 +1,99 @@
...
+	new Promise<void>((resolve) => {
+		const check = () => {
+			if (fs.existsSync(filename)) {
+				resolve();
+			} else {
+				setTimeout(check, timeout);
+			}
+		};
+		check();
+	});
```
3mcd, line 20: FFTI: could be simplified a little?
```ts
const wait = (ms: number) => new Promise(resolve => setTimeout(resolve, ms))
const blockForFile = (filename: string, timeout = 100) => {
  while (!fs.existsSync(filename)) {
    await wait(timeout);
  }
}
```

server/facets/ingest.ts
```diff
@@ -0,0 +1,56 @@
...
+			as: 'facetBinding',
+			onDelete: 'CASCADE',
+		});
+		modelsByName[name] = FacetModel;
+	});
+	return {
+		facetModels: modelsByName as Record<FacetName, any>,
+		FacetBinding,
+	};
+};
```
3mcd, line 56: FFTI: Not sure how I feel about `ingest`, since _modelize_ is taken, maybe `relationalize`, `dbify`, `toSequelize`? All terrible suggestions, so sorry.

3mcd (approved):
Looks good to me! 1/x defeated. Nothing but minor suggestions here. FFTI="feel free to ignore" btw

facets/core/prop.ts
```diff
@@ -0,0 +1,81 @@
...
+
+type IsNeverNull<T> = Extract<T, null> extends never ? true : false;
+
+export type AvailableCascadeStrategyForPropType<PropType extends FacetPropType> =
+	// Any prop type can use the `overwrite` strategy
+	| 'overwrite'
+	// Only props that are objects can object-merge during cascade" {...a, ...b}
+	| (TypeOfFacetPropType<PropType> extends Record<string, any> ? 'merge' : never)
+	// Only props that are arrays can array-merge during cascasde: [...a, ...b]
+	| (TypeOfFacetPropType<PropType> extends any[] ? 'extend' : never);
```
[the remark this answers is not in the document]
idreyn: That's a good suggestion, I will change this. Or maybe remove this/the notion of `strategy` entirely for now? There's no clear timeline for adopting it anywhere but I did do a fair amount of thinking through it here. Curious for your thoughts.

facets/core/prop.ts
idreyn: That's interesting and I agree that the length of the type's name does tend to provide this signal. There are a _lot_ of functions parameterized over these facet types coming up in future PRs, so we should probably be consistent about it.

facets/core/prop.ts
idreyn: Yes — in facets-world, `null` is a clear pane of glass through which values from higher scopes can shine through (I promise I will document all of this). In cases where you don't want this to happen, I imagine there being an attribute on the specific facet instance (i.e. the `FacetBinding`?) that says "don't accept a cascade from higher scopes" or equivalently "make me the root for this cascade", which would accomplish the same thing. We could also invent another special signal value that acts like `initial` in CSS. Probably best to see which fits the exact use cases that arise.

facets/core/prop.ts
```diff
@@ -0,0 +1,81 @@
+import { FacetPropType, NullableTypeOfPropType, TypeOfFacetPropType } from './propType';
+
+type IsNeverNull<T> = Extract<T, null> extends never ? true : false;
```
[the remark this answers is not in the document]
idreyn: They do differ in an interesting way which makes me think yours is better. Check out `Test1` vs `Test2` in [this example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgzgOQgNwgJwQVwDbYIwA8AKgHxQC8UAogB7BoCGAxsMQDRQB2O2ZEdETgBM4XFOigB+KPUzQAXFABmDbHAgBuALAAoUJFiJxGHgCZiZSt1xR+wQSKhEpy1eqiLZm3bv3QAaqpyFFAMnCDaeuDQRBBwwHgh8EioJriEgdhyJJF+TnHApklGqVi45pnZGkA):

```ts
type IsNeverNull1<T> = Extract<T, null> extends never ? true : false;
type IsNeverNull2<T> = null extends T ? false : true;

type Value = any;
type Test1 = IsNeverNull1<Value>; // --> boolean (?!)
type Test2 = IsNeverNull2<Value>; // --> false
```

facets/core/primitives.ts
```diff
@@ -0,0 +1,47 @@
+import { z } from 'zod';
+
+import { FacetPropType, propType } from './propType';
+
+export const string = propType({
+	name: 'varchar',
+	schema: z.string(),
+	postgresType: 'text',
```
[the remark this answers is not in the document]
idreyn: I'll change it — I think at one point this was more tightly bound to Postgres types or something.

facets/core/primitives.ts
```diff
@@ -0,0 +1,47 @@
...
+});
+
+export const choice = <T extends string, U extends [T, ...T[]]>(
+	values: U,
+): FacetPropType<z.ZodEnum<U>, { values: U }> => {
+	return propType({
+		identity: choice,
+		schema: z.enum(values),
+		postgresType: 'varchar',
+		extension: { values },
```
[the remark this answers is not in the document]
idreyn: I think we use it later on to read the specific values allowed by the enum in runtime code. I would much rather do this by introspecting the value of `z.enum(values)` but they're hidden behind underscored properties...

facets/core/prop.ts
3mcd: I think we should keep them long for now, but maybe we could prefix them (yuuuuck) with a `$` or `_` or something? Really ugly, but solves both problems...

facets/core/primitives.ts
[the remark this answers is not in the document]
3mcd: Gotcha, that makes perfect sense!

facets/core/parse.ts
```diff
@@ -0,0 +1,80 @@
...
+	ThrowErrorOnFailure extends boolean,
+> = {
+	valid: ThrowErrorOnFailure extends true ? ParsedRecordType : Partial<ParsedRecordType>;
+	invalid: Partial<Record<keyof ParsedRecordType, true>>;
+};
+
+function parsePartialOrEntireFacetInstance<
+	Def extends FacetDefinition,
+	InstanceKey extends keyof Def['props'],
+	AllowPartialInstance extends boolean,
```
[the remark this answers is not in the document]
idreyn: I believe it does if we want the TS language server to be able to infer the return type from the arguments (which, granted, is a real nice-to-have)

server/facets/ingest.ts
idreyn: How about `createSequelizeModelsFromFacetDefinitions` :P

PR #2223, opened 2022-09-01
fix: User authored list

gabestein:
Fixes user profile to shows authored Pubs in the authored pubs tab. Previously, this had been filtering on login id for some reason? Also yeah we really need to update user profiles. @isTravis was the last one to touch this ... 3 years ago ... so you win the review draw!

_Test plan_
- Visit a user profile who has some authored pubs and some contributed pubs at a community as a logged out user (e.g. `user/gabriel-stein`).
- Make sure you can see authored pubs in the authored list, and they are authors but not contributors on all those pubs.
- Make sure you can see all pubs in the other list, and they are authors or contributors on all those pubs
- Make sure you can't see any non-released pubs
- Profit?

### DIFF
--- a/client/containers/User/User.tsx
+++ b/client/containers/User/User.tsx
@@ -30,12 +30,10 @@ const User = (props: Props) => {
 	});
 	const authoredPubs = communityPubs.filter((pub) => {
 		const collaborators = pub.attributions || [];
-		const isAuthor = collaborators.reduce((prev, curr) => {
-			if (curr.isAuthor) {
-				return true;
-			}
-			return prev;
-		}, false);
+		const isAuthor = collaborators.some((collaborator) => {
+			const isCurrentProfileCollaborator = collaborator.user.id === userData.id;
+			return isCurrentProfileCollaborator && collaborator.isAuthor;
+		});
 		return isAuthor;
 	});
 	const pubsToRender = mode === 'authored' ? authoredPubs : communityPubs;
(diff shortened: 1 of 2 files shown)


client/containers/User/User.tsx
```diff
@@ -31,7 +31,7 @@ const User = (props: Props) => {
 	const authoredPubs = communityPubs.filter((pub) => {
 		const collaborators = pub.attributions || [];
 		const isAuthor = collaborators.reduce((prev, curr) => {
-			if (curr.user.id === loginData.id && curr.isAuthor) {
+			if (curr.isAuthor) {
```
isTravis, line 34: This will set `isAuthor` to true if _any_ of the collaborators are an author. We need to still only check the `.isAuthor` key for the collaborator that lines up with the current profile. I think that simply means:

```
if (curr.user.id === userData.id && curr.isAuthor) {
```

We could also do this a bit more cleanly now using `.some`:

```
const isAuthor = collaborators.some((collaborator) => {
	return collaborator.user.id === userData.id && collaborator.isAuthor;
});
```

Or - since this seems to have tripped me up 3 years ago, we can be more descriptive with an extra variable:
```
const isAuthor = collaborators.some((collaborator) => {
	const isCurrentProfileCollaborator = collaborator.user.id === userData.id;
	return isCurrentProfileCollaborator && collaborator.isAuthor;
});
```

isTravis (approved):

gabestein:
gotta re-do to fix the package lock fml

PR #2226, opened 2022-09-06
Facets (2/x): Server-side queries and API

idreyn:
This PR adds server-side methods to read facets from a given scope — or many scopes at once. The process looks a little like this:

1. Given a list of `{ community, collection, pub }` IDs, compute an extended list of these IDs that includes ancestor scopes. This is what `resolveScopeIds()` does. This produces what I call a "scope stack" — a list containing a scope's ID and all its parents' IDs.
2. Use the `FacetBinding` model to load facets associated with these scopes from the database. This is `loadFacetInstancesForBindingIds()`.
3. Put the two together: use the scope stacks to determine which of the loaded facet instances to cascade for each scope. This is `cascadeFacetsForScopeStack()`.

Since step (1) involves expanding a single scope ID into a list of many scope IDs, it made sense to treat loading the facets for a single scope as a special case of loading facets for many scopes. This would help us get a jump on using facets in places where we need to read from many scopes at once (e.g. Pages). The performance characteristics of this remain to be seen — in a growing list of Pubs, each with a growing list of facets, a caching mechanism may be needed.

It may also eventually make sense to move step (3) to the client, since it produces an object with a lot of repetition. For example, a Community-level facet will be duplicated `n` times in the cascades for each of `n` Pubs loaded within that Community.

<hr> 

Separately, I've also added an API route for updating facets (this wraps server-side logic from #2116


server/facets/resolveScopeIds.ts
```diff
@@ -0,0 +1,151 @@
+import * as types from 'types';
+import { Collection, CollectionPub, Pub } from 'server/models';
+import { bucketBy, indexById } from 'utils/arrays';
+import { getPrimaryCollection } from 'utils/collections/primary';
+import { assert } from 'utils/assert';
+
+import { ScopeKind, ByScopeKind, createByScopeKind } from '../../facets';
+
+export type ScopeStack = {
```
qweliant, line 9: super not important, I feel `ScopeStack` sounds like an actual stack is being made and that it should be treated as such. Is this true?

utils/objects.ts
```diff
@@ -22,3 +22,14 @@ export const pickKeys = <T extends Record<any, any>, K extends keyof T>(
 	});
 	return nextObject as Pick<T, K>;
 };
+
+export const mapObject = <Keys extends string, From, To>(
```
qweliant, line 26: neat!

qweliant (approved):
type system go burr. LGTM!!

Im assuming `fetchFacetsForScopeId/s` and `ingestFacets` will be used in the next pr

3mcd (approved):
LGTM!

PR #2237, opened 2022-09-07
Facets (4/x): Client-side state management

idreyn:
To recap our Facets journey so far:

- We have core code for manipulating facets and adding them to the database (#2222)
- We have server-side database-munging code and an API to retrieve and update facets (#2226)
- We pass down initial facets states to the client in routes that request it (#2238)

The next step is code that lets us manipulate facet state from the client! This PR adds a component, `FacetsStateProvider`, which lives at the root of our application and exposes a React context containing the facets for the current scope, and methods to change them. We're trying out [`zustand`](https://github.com/pmndrs/zustand) as a state management framework, which was a @3mcd suggestion from a while back which I quite like.

To access the facets state from anywhere in the application, please find two new hooks in `client/utils/useFacets`:

- **`useFacetsState()`** basically just pulls the Zustand store from context. This includes a lot of "behind-the-scenes" state related to validation, saving, etc. I expect that only UI for _editing_ facet state will want to touch this.
- **`useFacetsQuery()`** is a simpler way to find out what a current facet's value is. Usage looks like e.g. `const theme = useFacetsQuery(F => F.PubHeaderTheme)`.

This should be all the facets state management we need for the foreseeable future — there should be no need to interact with `/api/facets` from anywhere else in the application. In the far future, when there may be more than one scope active on a page at a time (e.g. in an SPA) then we may have to expand this to hold the facets for multiple scopes at once.


client/components/FacetsStateProvider/types.ts
```diff
@@ -0,0 +1,30 @@
...
+	Facets,
+} from 'facets';
+
+import { bindActionsToStore } from './actions';
+
+export type FacetState<Def extends FacetDefinition = FacetDefinition> = {
+	facetDefinition: Def;
+	persistedCascadeResult: FacetCascadeResult<Def>;
+	cascadeResult: FacetCascadeResult<Def>;
+	latestAndPossiblyInvalidCascadeResult: FacetCascadeResult<Def>;
```
qweliant, line 15: what is the difference between cascadeResult and persistedCascadeResult?
does this mean the facets for a scope can be in three different states?
why?

qweliant: oh maybe a facet can be updating(`persistedCascadeResult`) while what is still shown is `cascadeResult` and be failed(`latestAndPossiblyInvalidCascadeResult`)??

client/components/FacetsStateProvider/state.ts
```diff
@@ -0,0 +1,56 @@
...
+	initialCascadeResult: FacetCascadeResult<Def>,
+): FacetState<Def> {
+	return {
+		facetDefinition,
+		persistedCascadeResult: initialCascadeResult,
+		cascadeResult: initialCascadeResult,
+		latestAndPossiblyInvalidCascadeResult: initialCascadeResult,
+		persistableChanges: {},
+		invalidProps: {},
+		hasInvalidChanges: true,
```
qweliant, line 27: could this be null or true? I don't understand having an initial state of `true` for a property that should be changed by an update

client/components/FacetsStateProvider/actions/updateFacet.ts
```diff
@@ -0,0 +1,128 @@
...
+		const { value } = entry;
+		const patchedEntry = {
+			...entry,
+			value: { ...value, ...patch },
+		};
+		const nextStack = stack.concat();
+		nextStack.splice(patchableIndex, 1, patchedEntry);
+		return nextStack;
+	}
+	return [...stack, { scope, facetBindingId: null, value: createFacetInstance(def, patch) }];
```
3mcd, line 38: FFTI, thoughts on using immer here in some of these updater functions to simplify? I'm still on the fence about immer myself.

3mcd (approved):
LGTM, and thanks for the recap in the PR description.

client/components/FacetsStateProvider/state.ts
idreyn: That's because this is supposed to be `false` 😅

client/components/FacetsStateProvider/actions/updateFacet.ts
idreyn: I'd be in favor of a future refactor using Immer!

PR #2238, opened 2022-09-07
Facets (3/x): Load facets in route handlers

idreyn:
This is a relatively small change that touches a lot of files, so I split it into a separate PR. In short, it adds a `facets` property to `ScopeData` that holds the facets for the current scope. We don't need this on all routes, though, so I've added a bit of plumbing to `getInitialData()` that lets you explicitly request facets (I could see this becoming the default in the future). I changed the signature of that function to add a second `options` parameter, and then had to go change all the callsites — hence the breadth of this PR.


server/utils/queryHelpers/pubGet.ts
```diff
@@ -16,22 +17,37 @@ const resolveGetPubWhereQuery = (where: GetPubWhere) => {
...
-	const pubData = await Pub.findOne({
-		where: resolveGetPubWhereQuery(where),
-		...buildPubOptions({
-			getMembers: true,
-			getCollections: true,
-			getExports: true,
-			getEdges: 'approved-only',
-			...options,
+	where = resolveGetPubWhereQuery(where);
+	const [pubData, facets] = await Promise.all([
```
qweliant, line 33: interesting, yeah i see why it could be the default way to get facets off scopes(though i like property instead if facet. lol sorry for the late contrib). 

can you explain how TS might find a facet for, say `CitationStyleKind, and infer that to be the same `CitationStyleKind` type for `pub`

qweliant: or is this just what zod does

3mcd (approved):
LGTM!

server/utils/initData.ts
```diff
@@ -23,9 +23,17 @@ const getNotificationData = async (
...
+	isDashboard?: boolean;
+	includeFacets?: boolean;
+};
 
+export const getInitialData = async (
+	req,
+	options: GetInitialDataOptions = {},
+): Promise<types.InitialData> => {
+	const hostname = req.hostname;
+	const { isDashboard = false, includeFacets = isDashboard } = options;
```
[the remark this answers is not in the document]
idreyn: No, this is right — we want to default to loading facets for dashboard routes.

server/utils/queryHelpers/pubGet.ts
idreyn: Sorry to be 15 days late to this — the key is that we're getting rid of the `pub.citationStyle` property, so nothing has to do any matching to Pub properties. They're disappearing from the Pub per se and reappearing on facets attached to the Pub.

PR #2242, opened 2022-09-08
bug: Fix author pub count

qweliant:
See [here](https://github.com/pubpub/pubpub/pull/2223)


gabestein (approved):
Thanks! Was previously reviewed by @isTravis, so, LGTM.

PR #2244, opened 2022-09-12
enahncement: Redirect Logged-Out Users

qweliant:
this fixes an issue where logged-out users would not be redirected anywhere


howebenhowe (approved):
Tidy.
Shipit!

PR #2245, opened 2022-09-12
booooo package-lock

gabestein:
for our call tomorrow, here's what I get on node 18.7.0 npm 8.15.0


3mcd (approved):
So excited to never have package-lock problems ever again!

(narrator: they would, in fact, have package-lock problems)

PR #2260, opened 2022-09-21
fix: pass collection to collection page metadataComponent

tefkah:
See https://github.com/pubpub/pubpub/discussions/2124#discussioncomment-3704285


gabestein (approved):
Good catch! Thanks so much for the fix, and sorry it took this long for something this simple, I thought it was going to be a bit more complex.

PR #2265, opened 2022-09-26
fix: make sure metadata is set correctly for collections

tefkah:


gabestein:
Closing in favor of #2307, which expands on this to be a bit more comprehensive. Thanks very much for the initial fixes and PR!

PR #2268, opened 2022-09-29
bump forked prosemirror-math to get history plugin

howebenhowe:
like it says on the tin

brings `prosemirror-history` compatibility to the `prosemirror-math` plugin, enabling undo and redo while remaining within a math node selection.


idreyn (approved):
It's good!

PR #2289, opened 2022-10-11
Visiting Commenter

qweliant:
So a good chance to implement [MinUser](https://readwise.io/reader/shared/01ggqp3wbhsdthpfz8h2ybpjgx) has come up in the form of commenters.

Commenters are users without accounts. 

MinUser is similar to what Ian described as phantom users. Phantom users were used by khan academy to enable randos to use stuff and import whatever progress they had through the onboarding process. People from arcadia are wanting to present pubs at a conference. It would be great if anyone could comment on conference presentations through PubPub. We should consider whether or not users are truly anonymous or if they will be tracked via session cookies. Or could this behavior be facetized? 

For now, a commenter model will be added. 

Anyone can comment on a pub using the link. This link is found in the draft sharing button in the Pub header

Test plan:

Create a Pub and release it. Visit a Pub in a draft. I suggest opening two new tabs. From the share-button copy each link, one for draft comments, one for release comments, into a tab. Create a discussion at each link. Try creating discussions logged in and logged out. You should also try commenting on different versions of a draft as a visiting commenter to be sure comments only show up in the correct history.


qweliant:
Known issue: given a link users can create a discussion on a draft or the latest Pub release at any point now. discussions render appropriately for a draft page, a release page, and a release comment page, but not the draft comment page. That page has the release comments on them and I am not sure why

server/thread/queries.ts
```diff
@@ -63,3 +64,48 @@ export const canUserSeeThread = async (options: CanUserSeeThreadOptions): Promis
...
 };
+
+type CreateThreadWithCommentOptions = {
+	text: string;
+	content: types.DocJson;
+	userId: null | string;
+	commenterName: null | string;
+};
+
+export const creatThreadCommentWithUserOrCommenter = async (
```
idreyn, line 75: This should be in `server/threadComment/queries.ts` since it creates a `ThreadComment`, and doesn't deal with `Thread` objects at all.

server/routes/pubDocument.tsx
```diff
@@ -104,6 +107,17 @@ const getEnrichedPubData = async ({
 		pubData.reviewHash = reviewHash;
 	}
 
+	if (!pubData.commentHash) {
```
idreyn, line 110: It's generally better to do this sort of thing as a migration so the codebase doesn't accumulate these ad-hoc backfills. If you keep adding them eventually you get to a point where it's hard to reason about the state of any Pub. "Just one" is okay but we should really strive to hold the line on this.

server/routes/pubDocument.tsx
```diff
@@ -291,3 +305,64 @@ app.get(
...
+		const { historyKey: historyKeyString, pubSlug } = req.params;
+		const { canView } = initialData.scopeData.activePermissions;
+		const { hasHistoryKey, historyKey } = checkHistoryKey(historyKeyString);
+		const customScripts = await getCustomScriptsForCommunity(initialData.communityData.id);
+
+		if (!canView) {
+			throw new NotFoundError();
+		}
+
+		const pubData = await Promise.all([
```
idreyn, line 326: You can simplify this whole invocation to:

```ts
const pubData = await getEnrichedPubData(...);
```

client/containers/Pub/PubHeader/TitleGroup.tsx
```diff
@@ -16,12 +16,14 @@ type Props = {
...
 	const { pubData, updatePubData } = props;
-	const { title, htmlTitle, description, isRelease, isReviewingPub } = pubData;
+	const { title, htmlTitle, description, isRelease, isReviewingPub, isAVisitingCommenter } =
+		pubData;
 	const { communityData, scopeData, featureFlags } = usePageContext();
 	const { submissionState } = usePubContext();
 	const isUnsubmitted = submissionState?.submission.status === 'incomplete';
 	const { canManage } = scopeData.activePermissions;
-	const canModify = canManage && !isRelease && !isUnsubmitted && !isReviewingPub;
+	const canModify =
```
idreyn, line 25: I don't think we need to modify this check here. If a user `canManage` it doesn't really matter whether `isReviewingPub` or `isAVisitingCommenter` — we still want to give them the ability to do it.

server/routes/pubDocument.tsx
```diff
@@ -291,3 +305,64 @@ app.get(
 		return renderPubDocument(res, pubData, initialData, customScripts);
 	}),
 );
+
+app.get(
```
idreyn, line 309: ⚠️ I'm kind of iffy about adding two new routes for this purpose. The user already has to supply `commentHash` in the URL in order to be able to make comments as a logged-out user. Shouldn't we be able to infer that they're in "visiting commenter" mode by that alone?

server/discussion/model.ts
```diff
@@ -10,9 +10,10 @@ export default (sequelize, dataTypes) => {
 			/* Set by Associations */
 			threadId: { type: dataTypes.UUID, allowNull: false },
 			visibilityId: { type: dataTypes.UUID, allowNull: false },
-			userId: { type: dataTypes.UUID, allowNull: false },
+			userId: { type: dataTypes.UUID, allowNull: true },
 			anchorId: { type: dataTypes.UUID },
 			pubId: { type: dataTypes.UUID },
+			commenterId: { type: dataTypes.UUID, allowNull: true },
```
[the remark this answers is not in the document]
qweliant: There's a problem when trying to post comments. I get errors when this is not associated, but I can also try reproducing the errors and presenting them to see where i may have gotten it wrong

server/routes/pubDocument.tsx
qweliant: I'm going to solve this part when we get back, but I completely agree!

server/userSubscription/model.ts
```diff
@@ -5,7 +5,7 @@ export default (sequelize, dataTypes) => {
 			id: sequelize.idType,
 			setAutomatically: { type: dataTypes.BOOLEAN, allowNull: false },
 			status: { type: dataTypes.STRING, allowNull: false },
-			userId: { type: dataTypes.UUID, allowNull: false },
```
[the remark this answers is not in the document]
idreyn: Please revert this before landing!

idreyn (approved):

PR #2297, opened 2022-10-19
Add data-tab-related-count attribute to allow access in CSS to tabs in User Profile

joshdavenport:
This allows pubs to style based on the count via `[data-tab-related-count="0"]`, or use it via `attr(data-tab-related-count)` in CSS. 

We want it for the latter, so that we can change the wording here, but still have the count
(e.g. `:before { content: 'Custom Title (' attr(data-tab-related-count) ')'; }`

This only changes the way this content is output and doesn't change any behaviour, the result to the user will be indistinguishable but increases the surface area for extendibility.

### DIFF
--- a/client/containers/User/UserNav.tsx
+++ b/client/containers/User/UserNav.tsx
@@ -19,8 +19,8 @@ const defaultProps = {
 
 const UserNav = function (props) {
 	const tabs = [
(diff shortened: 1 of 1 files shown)


3mcd (approved):

3mcd (approved):
LGTM!

joshdavenport:
Thanks Eric! Green is a stunning album by the way, one of my favourites.

PR #2299, opened 2022-10-24
fix: Allow reviews to be submitted on Spubs

### LINKED ISSUE #2295: Reviews returning 500 error on spubs
1. Create a spub
2. Turn on reviews
3. Try to submit review
4. Get 500 error
5. Logs:

```
2022-10-18T13:29:21.299592+00:00 app[web.4]: Error in postReview:  Error: Not Authorized
2022-10-18T13:29:21.299609+00:00 app[web.4]:     at /app/dist/server/server/review/api.js:23:19
```

Does not seem to occur in normal Pubs.

Review submitted successfully.

e.g. https://commonplace.knowledgefutures.org/pub/ww9grt9c/draft

===

William / SAR
Allison V. in #2290

idreyn:
Resolves #2295 

The error described therein turned out to be down to an intentional line of code in `getScope()`:

```ts
canCreateReviews = !activePub.submission && ...
```

I added this check while developing Submissions because at the time the feature was intended to be mutually exclusive with Reviews — this is no longer the case. The practical effect of the change at the time was to hide the "Request Release" button in the Pub header. I think we want to continue to hide that, so I've moved this check to the frontend.

_Test plan:_
- Visit a Collection with open submissions to create a Spub
- Verify that the Request Release button is not shown
- (optional) Complete the submission process
- Visit `/pub/<slug>/review/1` and verify that you can add a review from the little dropdown bit

I also fixed two peripherally related issues here:

- A boolean logic error that was causing the `<ReviewHeaderSticky>` to appear when it shouldn't
- A bizarre (and bizarrely old) chunk of CSS that was mangling the `<FileUploadButton>`

### DIFF
--- a/client/components/FileUploadButton/fileUploadButton.scss
+++ b/client/components/FileUploadButton/fileUploadButton.scss
@@ -1,16 +1,5 @@
 .file-upload-button-component {
 	display: block;
-	.upload-button {
-		position: relative;
-		label {
-			position: absolute;
-			left: 0;
-			right: 0;
-			top: 0;
-			bottom: 0;
-			cursor: pointer;
-		}
-	}
 	input.file-input {
 		display: none;
 	}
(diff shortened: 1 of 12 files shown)


3mcd (approved):
LGTM!

PR #2305, opened 2022-10-27
feature: code blocks w/ plugin

howebenhowe:
We are going with this approach!

_Test Plan_
create code block in a pub and play with them generally. Steps below. NB: the code blocks behave as a text block, which is to say, they are not deletable if they are the only text block in the document just as the only `paragraph` block would not be deletable. We can change this behavior, but it would be a departure from block behavior to accommodate image-block-like behavior.
Make sure to hit each of the below:
+ create a code block by typing three backticks to enter it
+ create a code block by clicking the lego-block-shaped codeblock toggle button in the formatting bar
+ create a code block by selecting `code` from the dropdown at the left edge of the formatting bar
+ exit the codeblock by pressing `return` from an empty line in the codeblock
+ exit the codeblock by navigating to the right from the last character (per NB above, this is impossible if there isn't a following text block to navigate into)
+ while inside a codeblock, select different language syntax highlighting grammars, then reload the page
+ copy + paste into/out of codeblocks
+ delete codeblocks with backspace (provided there is a text node above the codeblock in the document)

TODO:
+ implement the language selector dropdown component using Blueprint components for visual consistency, using the `createSelect` api from the vendor module.
+ PDF export

### DIFF
--- a/client/components/Editor/plugins/code.ts
+++ b/client/components/Editor/plugins/code.ts
@@ -1,0 +1,21 @@
+import { undo, redo } from 'prosemirror-history';
+import {
+	codeMirrorBlockPlugin,
+	defaultSettings,
+	languageLoaders,
+	legacyLanguageLoaders,
+} from 'prosemirror-codemirror-block';
+
+export default (schema) => {
+	if (schema.nodes.code_block) {
+		return [
+			codeMirrorBlockPlugin({
+				...defaultSettings,
+				languageLoaders: { ...languageLoaders, ...legacyLanguageLoaders },
+				undo,
+				redo,
+			}),
+		];
+	}
+	return [];
+};
--- a/client/components/Editor/plugins/code.ts
+++ b/client/components/Editor/plugins/code.ts
@@ -1,16 +1,68 @@
 import { undo, redo } from 'prosemirror-history';
 import {
+	CodeBlockSettings,
 	codeMirrorBlockPlugin,
 	defaultSettings,
 	languageLoaders,
 	legacyLanguageLoaders,
 } from 'prosemirror-codemirror-block';
 
+import { EditorView } from 'prosemirror-view';
+import { Node } from 'prosemirror-model';
+
+export const myCreateSelect = (
+	settings: CodeBlockSettings,
+	dom: HTMLElement,
+	node: Node,
+	view: EditorView,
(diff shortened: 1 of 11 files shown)


3mcd:
Looks and works great! 🎉  Tested a few other scenarios like collaboration and undo/redo and they worked great. Waiting for styling tweaks to approve.

client/components/Editor/plugins/code.ts
```diff
@@ -0,0 +1,73 @@
...
+	node: Node,
+	view: EditorView,
+	getPos: (() => number) | boolean,
+) => {
+	if (!settings.languageLoaders) return () => {};
+	const wrapper = document.createElement('div');
+	wrapper.className = 'bp3-html-select codeblock-select-wrapper';
+	const select = document.createElement('select');
+	const carets = document.createElement('span');
+	carets.className = 'bp3-icon bp3-icon-double-caret-vertical';
```
idreyn, line 25: `bp3-icon-double-caret-vertical` is having no effect as far as I can see (Firefox). I also think that we should use `caret-down` for consistency with other dropdowns in Blueprint.

client/components/Editor/plugins/code.ts
```diff
@@ -0,0 +1,73 @@
...
+		if (typeof getPos === 'function') {
+			view.dispatch(
+				view.state.tr.setNodeMarkup(getPos(), undefined, {
+					...node.attrs,
+					lang,
+				}),
+			);
+		}
+	};
+	// Delete code.
```
idreyn, line 56: Is this a note to self about this code, or is it about the code block itself?

idreyn (approved):
LGTM, I have one usability concern I'd like to see addressed before landing — we need to make sure the `<PubInlineMenu>` isn't showing up while you're inside a code block. We'll have to add another clause to this chunk of code here:

https://github.com/pubpub/pubpub/blob/8064b262695e9bc959589c8e222a5936ca3664ca/client/containers/Pub/PubDocument/PubInlineMenu.tsx#L30

I'd recommend breaking your new logic along with the existing logic into a helper function and then computing it inside a `useMemo()` for best performance.

image

My other thought is, maybe the font size for code blocks should be a few px smaller?

3mcd (approved):

Issue #2218, opened 2022-09-01
Dashboard menu on mobile off left of screen


gabestein:
### What went wrong, step-by-step?
1. Visit on a mobile device
2. Click top dashboard menu
3. Flows off left of screen!

### What did you expect to happen?
- Stays on screen.

### What browser/operating system were you using?
iPhone / iOS

===

gabestein:
An example on iPhone Safari in portrait mode: https://hdsr.mitpress.mit.edu/pub/ftwzgrij/release/1

djagdish:
1. when the width is capped at 300px
2. when width is capped, and text is wrapped

<img width="443" alt="image" src="https://user-images.githubusercontent.com/316939/202485530-3c63128d-4bf3-4c14-9245-1823b8809c0a.png">

PR #2324, opened 2022-11-03
fix: Dashboard Menu position in header nav (#2218)

djagdish:
This change affects non-mobile browsers also by moving the popover to the right by 30px, but that looks fine. So we'll stick to this solution for now.


qweliant (approved):

gabestein:
Implemented in deepak's landing page refresh branch.

PR #2327, opened 2022-11-07
dev: extract bp3 to namespace variable

howebenhowe:
Addresses prerequisits for #2328 
Adds blueprintjs eslint rule (with accompanying dependency)
Extracts `*.scss` located instances of `bp3` to variable
Move from hardcoded classes it `.tsx` to using `@blueprintjs/core.CLASSES`

Various tiny housekeeping including:
+ enforcing `STYLEGUIDE.md` with respect to import order
+ unique keys on components from `map`'d function

## Test Plan
+ for `.css` consistency, compare a freshly generated `dist/client/main.css` from `master` to the the same generated file from this branch; they should be identical
+ for `.tsx`/components...?


3mcd (approved):
I did a quick smoke test, including the Pub release and draft pages (including connections), every dashboard page, page layouts, submissions, and probably some others I'm forgetting. There were no inconsistencies with existing styles.

I'm giving this the go-ahead, but I wonder if there's a way to remove the `vendor.` prefix from each place we interpolate the `$bp-namespace` variable in our scss files.

Great work @howebenhowe!

PR #2336, opened 2022-11-09
fix: Include `structuredValue` as HTML in Pandoc footnotes

### LINKED ISSUE #2296: Word and Markdown exports are duplicating footnote text

* That exporting a Pub with footnotes to Word and/or Markdown would produce only a single copy of each footnote
* Instead, the footnote text of each footnote is printed twice, with a line break in between

* Any Pub with footnotes, as far as my testing

Gabe's test: https://demo.pubpub.org/pub/o8c78ua9/draft

* Safari for macOS Ventura beta

===

duplicate-footnotes

idreyn:
Resolves #2296

There is a goof (in recent code that I wrote) that double-adds the `unstructuredValue` of a footnote (the rich text content) instead of including `structuredValue` (from Cite.js) as well. This fixes that. We also strip most of the HTML content from the structured value so it appears cleanly in Markdown.

_Test plan:_
Create a Pub containing a footnote with a structured and unstructured value. Export it to Markdown and Word and verify that both are shown in the footnote.

### DIFF
--- a/workers/tasks/export/notes.ts
+++ b/workers/tasks/export/notes.ts
@@ -12,7 +12,8 @@ export type PandocNote = {
 	id: string;
 	cslJson: Record<string, any>;
 	hasStructuredContent: boolean;
-	unstructuredHtml: string;
+	unstructuredHtml: null | undefined | string;
+	structuredHtml: null | undefined | string;
 };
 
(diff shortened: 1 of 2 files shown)


3mcd (approved):
Noice!

PR #2337, opened 2022-11-09
dev: Update `chokidar`

idreyn:
[`chokidar`](https://www.npmjs.com/package/chokidar) an npm package that abstracts over `fs.watchFile()` and some other platform-specific APIs for detecting when files change. We use it in the hot-reloading code, which is recently giving us trouble. I updated it to the latest version, and things seem a _little_ better, though I don't know if this is really the fix. Still, it's harmless to try!

_Test plan:_
Start your dev server and make sure hot reloading of server-side code is happening!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -179,7 +179,7 @@
 				"@typescript-eslint/parser": "^5.3.0",
 				"autoprefixer": "^9.5.0",
 				"chalk": "^2.4.2",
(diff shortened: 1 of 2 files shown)


3mcd (approved):
I (fortunately) didn't have any issues with live reloading before, but I can confirm this didn't break my setup.

PR #2341, opened 2022-11-10
feature: Restore Pub to previous point in history

idreyn:
As a little thursday afternoon treat I've added a button to the Pub history viewer that lets you restore Pub contents back to a specific point in history. Users with edit access to a Pub draft will see it here:

image

When you click that, you get this dialog:

image

When you click _that_, we:

- Grab the document at the current `historyKey`
- Create a new Prosemirror `ReplaceStep` that replaces the old document in its entirety with the new one
- Write that step into a Firebase change

The history remains append-only — if you want to undo your revert, you can just go back one step and revert from there.

I decided to do this on the server side. Although we have all the information we need to execute this on the client, we'd actually need to:

- Place the target document in some intermediate state
- Manipulate the Pub page state so the history viewer disappears
- Wait until the Pub collab plugin loads again
- And finally, fire the `ReplaceStep` into the editor with a new transaction.

This starts to feel brittle, so I just created an API route that does this instead.

_Test plan:_
Visit a Pub and add a bunch of contents...keep typing...good. Now click the draft history button in the Pub header. Use the new restore controls to revert back to a previous state. Then verify that the old state is still in the history and you can revert back to _that_. Refresh the page and verify that nothing has gone terribly wrong.

### DIFF
--- a/client/components/ConfirmDialog/ConfirmDialog.tsx
+++ b/client/components/ConfirmDialog/ConfirmDialog.tsx
@@ -3,14 +3,15 @@
  * (bring your own button) -- it renders a `renderButton` prop with an `open` callback that can be
  * attached to a Button onClick handler, or anything else.
  */
-import React, { useState } from 'react';
-import { Button, Classes, Dialog, Intent } from '@blueprintjs/core';
+import React, { useCallback, useState } from 'react';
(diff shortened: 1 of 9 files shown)


3mcd:
Looks good to me! Tested and works as expected.

3mcd (approved):

PR #2349, opened 2022-11-14
fix: Pass a DepositTarget into DOI settings

### LINKED ISSUE #2343: Manual DOI postfix not possible in facet-land
1. Visited a Pub in a community with a custom DOI prefix target set (e.g. https://pubpub.maastrichtuniversitypress.nl/dash/pub/abuse-proportionality-and-the-burden-of-proof-in-cjeus-case-law-on-direct-taxation/settings/doi)
2. Attempted to deposit a Pub in the new DOI settings tab
3. Could not add a custom DOI postfix. Only an auto-generated one.

Access to custom postfix depositing interface.

https://pubpub.maastrichtuniversitypress.nl/dash/pub/abuse-proportionality-and-the-burden-of-proof-in-cjeus-case-law-on-direct-taxation/settings/doi

===

Maastricht

idreyn:
Resolves #2343

Another facets merge issue here :/ Looks like we lost the `depositTarget` that was being passed down into `PubSettings`. In the future we could consider loading it as an association to `initialData.community` instead, but I like that we're only loading it precisely where it's needed here.

_Test plan:_
- On your devserver, switch to a Community with an associated `DepositTarget` (like `mitp-arch`)
- Create and release a new Pub
- From Pub Settings, start the process of depositing the Pub and verify that you can customize the DOI suffix with this control:

image

### DIFF
--- a/client/containers/DashboardSettings/PubSettings/Doi.tsx
+++ b/client/containers/DashboardSettings/PubSettings/Doi.tsx
@@ -243,8 +243,7 @@ class Doi extends Component<Props, State> {
 			canIssueDoi &&
 			// a deposit has not been submitted yet for this work
 			!(justSetDoi || pubData.crossrefDepositRecordId) &&
-			// the Pub is not a supplement to another work
-			// and the community has a custom, hardcoded DOI prefix
+			// and the Community has a custom DOI prefix
 			this.props.depositTarget &&
 			doiPrefix !== PUBPUB_DOI_PREFIX
 		);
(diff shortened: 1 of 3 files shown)


client/containers/DashboardSettings/PubSettings/Doi.tsx
```diff
@@ -243,8 +243,7 @@ class Doi extends Component<Props, State> {
 			canIssueDoi &&
 			// a deposit has not been submitted yet for this work
 			!(justSetDoi || pubData.crossrefDepositRecordId) &&
-			// the Pub is not a supplement to another work
-			// and the community has a custom, hardcoded DOI prefix
```
idreyn, line 247: This duplicates a comment further down in the file where this check actually occurs.

client/containers/DashboardSettings/PubSettings/Doi.tsx
```diff
@@ -243,8 +243,7 @@ class Doi extends Component<Props, State> {
 			canIssueDoi &&
 			// a deposit has not been submitted yet for this work
 			!(justSetDoi || pubData.crossrefDepositRecordId) &&
-			// the Pub is not a supplement to another work
```
idreyn, line 246: This duplicates a comment further down in the file where this check actually occurs.

3mcd (approved):

Issue #2261, opened 2022-09-22
Change names of author profile lists


gabestein:
### What's your enhancement request?
Rather than "All Pubs" and "Authored Pubs," we should come up with better wording that, at minimum, deemphasizes authorship. E.g. "Byline Pubs".

### What would this enhancement help you accomplish?
De-emphasize authorship as a core value.
---

### Who requested this? (If not you)
Megan / Arcadia

PR #2358, opened 2022-11-27
Kill author and implement contributor taxonomy

zverhope:
Addresses #2261 - eliminates "author" in all instances where a wider range of contributor roles should be represented, and implements clean attribution/contribution taxonomy.

### DIFF
--- a/client/containers/DashboardEdges/DashboardEdges.tsx
+++ b/client/containers/DashboardEdges/DashboardEdges.tsx
@@ -133,7 +133,7 @@ const DashboardEdges = (props: Props) => {
 					<NonIdealState
 						icon="layout-auto"
 						title="No connections from other Pubs"
-						description="When other authors add connections to this Pub, they will be listed here."
+						description="When other contributors add connections from their Pubs to this Pub, they will be listed here."
 					/>
 				)}
 			/>
--- a/client/containers/DashboardEdges/DashboardEdges.tsx
+++ b/client/containers/DashboardEdges/DashboardEdges.tsx
@@ -133,7 +133,7 @@ const DashboardEdges = (props: Props) => {
 					<NonIdealState
 						icon="layout-auto"
 						title="No connections from other Pubs"
-						description="When other contributors add connections from their Pubs to this Pub, they will be listed here."
+						description="Corresponding connections will be created here when users connect to this Pub from another."
 					/>
 				)}
 			/>
--- a/client/containers/Pub/PubDocument/PubBottom/Discussions/LabelFilter.tsx
+++ b/client/containers/Pub/PubDocument/PubBottom/Discussions/LabelFilter.tsx
@@ -205,8 +205,7 @@ class LabelFilter extends Component<Props, State> {
 												content={
 													label.publicApply ? (
 														<span>
-															All discussion contributors can apply
-															this label.
+															All discussants can apply this label.
 														</span>
 													) : (
 														<span>
@@ -305,9 +304,7 @@ class LabelFilter extends Component<Props, State> {
 									<Tooltip
 										content={
 											label.publicApply ? (
-												<span>
-													All discussion authors can apply this label.
-												</span>
+												<span>All discussants can apply this label.</span>
 											) : (
 												<span>Only managers can apply this label.</span>
 											)
--- a/client/containers/Pub/PubDocument/PubBottom/Discussions/LabelFilter.tsx
+++ b/client/containers/Pub/PubDocument/PubBottom/Discussions/LabelFilter.tsx
@@ -205,8 +205,8 @@ class LabelFilter extends Component<Props, State> {
 												content={
 													label.publicApply ? (
 														<span>
-															All discussion contributors can apply this
-															label.
+															All discussion contributors can apply
+															this label.
 														</span>
 													) : (
 														<span>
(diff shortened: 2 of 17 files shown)


3mcd (approved):
Looks good to me! We'll definitely want @gabestein's eyes on it, too.

client/containers/DashboardEdges/DashboardEdges.tsx
```diff
@@ -133,7 +133,7 @@ const DashboardEdges = (props: Props) => {
 					<NonIdealState
 						icon="layout-auto"
 						title="No connections from other Pubs"
-						description="When other authors add connections to this Pub, they will be listed here."
+						description="When other contributors add connections from their Pubs to this Pub, they will be listed here."
```
gabestein, line 136: Not a blocker, but this one strikes me as slightly confusing. Contributors can't technically add connections from their Pubs. Only Pub Admins can. So perhaps we should say "When other users add connections from their Pubs..."

client/containers/Pub/PubDocument/PubBottom/Discussions/LabelFilter.tsx
```diff
@@ -205,8 +205,8 @@ class LabelFilter extends Component<Props, State> {
 												content={
 													label.publicApply ? (
 														<span>
-															All discussion authors can apply this
-															label.
+															All discussion contributors can apply
```
gabestein, line 208: Again, strikes me as a bit awkward here, since we we're now making Contributors an important word, and this isn't really an instance of it (at least not now). So, perhaps: `Anyone can apply this label to their discussions.`

client/containers/Pub/PubDocument/PubDiscussions/Discussion/LabelSelect.tsx
```diff
@@ -116,7 +116,7 @@ class LabelSelect extends Component<Props, State> {
 												<Tooltip
 													content={
 														label.publicApply
-															? 'All discussion authors can apply this label.'
+															? 'All discussion contributors can apply this label.'
```
gabestein, line 119: Same as above, might do something like: `Anyone can apply this label to their discussions.`

client/containers/Pub/PubDocument/PubFileImport/MetadataEditor.tsx
```diff
@@ -247,7 +247,7 @@ const MetadataEditor = (props: MetadataEditorProps) => {
 		if (attributions && attributions.length) {
 			return (
 				<>
-					<h6>Authors</h6>
+					<h6>Contributors</h6>
```
gabestein, line 250: Import automatically adds folks as byline attributions, so we should change this to `Attributions`

server/discussion/__tests__/api.test.ts
```diff
@@ -17,7 +17,7 @@ const couldApplyManagedLabel = {
 };
 
 const publicLabel = {
-	title: 'Authors can apply this label to their own discussions',
+	title: 'Contributors can apply this label to their own discussions',
```
gabestein, line 20: Should match the language you use above.

gabestein (changes requested):
Code wise, looks great, and I think you caught all the instances and made it a very clear taxonomy. A few small thoughts on discussions and import that I think we should discuss.

client/containers/Pub/PubDocument/PubBottom/Discussions/LabelFilter.tsx
zverhope: Yes, this is a great point. Some discussants may qualify as recognized contributors, but most won't. "Anyone" works better than "contributor," and I think that more than suffices, but I also think it's useful to specify for managers that only logged-in/authenticated users--`members` seems the best word here--will be able to discuss and use these labels. So, `All members can apply this label.`

client/containers/Pub/PubDocument/PubDiscussions/Discussion/LabelSelect.tsx
zverhope: Changed to `All discussants`, as above.

client/containers/Pub/PubDocument/PubFileImport/MetadataEditor.tsx
zverhope: ✔️ - makes sense that byline attribution would be the default on import. Thanks for catching that.

server/discussion/__tests__/api.test.ts
zverhope: I changed the sentence to the same as the public labels above in order to avoid the ouroboros of discussants applying labels to their discussions. (Sans period to be consistent with adjacent strings.

client/containers/DashboardEdges/DashboardEdges.tsx
zverhope: I was tempted to go with `members` here, as you can only use predictive search to connect to Pubs within your own Community. But connections are still made between Pubs in separate Communities when you enter the Pub URL or DOI, so the Community-restricted term doesn't quite encompass all use cases. (I'm just writing this out in case there's an error in my logic/understanding.) I did invert the syntax of the sentence to remove the conditional structure and make it a touch more descriptive. I also made it "users" instead of "other users," because a user could create connections between two or more Pubs that they administer. And because having permissions to make connections doesn't necessarily assume ownership, I removed the possessive, "their." So, we end up with: `Corresponding connections will be created here when users connect to this Pub from another.`

gabestein (approved):

Issue #2359, opened 2022-11-28
Anonymous comments fail to create or render anchors


isTravis:
### What went wrong, step-by-step?
Creating in-line comments using the anonymous Can-Comment-on-draft links fails to create anchors. Further, anonymous commenters (i.e. those using the can-comment link) cannot see anchored comments made by others on the draft.

Perhaps due to the firebase auth requirements that exist on draft but do not exist on releases. 

### Have any screenshots? (Attach them by dragging the image here)

Video attached. The top screen has a logged in user. The bottom screen is using a can-comment link.
https://recordit.co/OtBOQs7mhN

![OtBOQs7mhN](https://user-images.githubusercontent.com/1000455/204330258-4677d219-884a-4acb-84a0-3348bada58ab.gif)

PR #2369, opened 2022-11-30
Render Anchors 

qweliant:
adresses #2359 

set plugin to true, allow anchors to render on draft


client/containers/Pub/usePubBodyState.ts
```diff
@@ -40,15 +40,15 @@ export const usePubBodyState = (options: Options): PubBodyState => {
 	} = usePageContext();
 
 	const discussionAnchors = useMemo(() => {
-		if (isRelease) {
+		if (isRelease || isAVisitingCommenter) {
```
qweliant, line 43: i am not sure why i need to do this for new route because draft pubs just render discussion groups

idreyn (approved):
LGTM :+1:

PR #2377, opened 2022-12-01
feature: SpamTag model, queries, etc.

### LINKED ISSUE #2371: `SpamTag` model, queries, API
- Model associated with a Community to hold the spam properties we compute for it
- Code to check strings in a Community against words in our spamlist
- Sequelize hooks to automatically compute a `SpamTag` for a Community when it is updated/changed
- API to mark communities as maybe/definitely/not spam

idreyn:
This resolves #2371. We add a `SpamTag` model associated with `Community`. It holds two things:

- A `spamScore` which is automatically computed by seeing how often spammy words appear in different Community metadata fields. We're basically making up the algorithm for this as we go along, and we can change it to our hearts' content later. This is done automatically by a Sequelize hook whenever a Community is created or updated.
- A `status` which marks a Community as unreviewed, definitely spam, or definitely not spam. This is done manually by us (via the `/api/spamTags` route).

These two fields don't interact or depend on each other in any way — in other words, **there is no automatic spam blocking in this feature, though we have all the pieces we need to trivially add it later!** Instead, the `spamScore` is used to help surface Communities that are _probably_ spam to make it easy for us to remove them.

_Test plan:_
You can run some very basic API tests:

```
npm run test-dev server/spamTag
```


server/spamTag/phrases.ts
```diff
@@ -0,0 +1,44 @@
...
+	'we are a',
+	'we are an',
+	'are you',
+	'would you',
+	'do you',
+	'we supply',
+	'our website',
+	'our site',
+	'if you are',
+	'ivermectin',
```
qweliant, line 43: 😭

qweliant (approved):

PR #2252, opened 2022-09-15
Superadmin dash and basic landing page management

idreyn:
This PR is in service of the new landing page, which is shaping up nicely. It makes these additions:

1. A new `isSuperAdmin` field on `User`. We'll want to set this to `true` for all of our personal accounts as well as `pubpub-team`. We should never create a way to manipulate this field without direct access to our database. This replaces our hard-coded check for the `pubpub-team` UUID.
2. A new dashboard at `/superadmin` where we can put stuff for superadmins, i.e. us!
3. A tab on that dashboard that lets you manipulate the lists of Pubs and Communities that will appear on the landing page. These are collectively called `LandingPageFeatures` and there's a new model for them.

It looks like this — I'm nothing without `communityAccentDark`!

image

_Test plan:_

- Run the unit tests:

```
npm run test-dev server/landingPageFeature
```

- Visit `/superadmin` and make sure you can add, reorganize, and remove Pubs and Communities
- Visit the homepage (you'll need to select "Base PubPub" from the Community switcher) and make sure your changes are reflected there.

PR #2378, opened 2022-12-01
feature: Community search (for spam dashboard)

### LINKED ISSUE #2372: API to search/query Communities
For the spam superadmin dashboard we need:

- A way to query and sort Communities by properties on their associated `SpamTag`
- A way to search Communities by title, subdomain, etc.

This will probably look a lot like `/api/pubs/many` and I expect to reuse some of the same patterns. I will scope this code to spam for now, but we could use it in other Community search contexts, too.

idreyn:
Resolves #2372. This adds an API route (clunkily, a POST handler at `/api/spamTags/queryCommunitiesForSpam`) to search Communities by various fields inside the superadmin spam dashboard. This is fairly generalizable code, and it might seem like slight overkill here, it would be easy to repurpose for other Community searches in the future!

_Test plan:_
Uhh, none for now. We can test this out once the spam dashboard is implemented (see #2252)


server/spamTag/communities.ts
```diff
@@ -0,0 +1,83 @@
...
+const getSpamTagWhereQuery = (status: undefined | types.SpamStatus[]) => {
+	if (status) {
+		return { where: { status: { [Op.in]: status } } };
+	}
+	return {};
+};
+
+const getQueryOrdering = (ordering: types.SpamCommunityQueryOrdering) => {
+	const { field, direction } = ordering;
+	return [[...orderableFields[field], direction]];
```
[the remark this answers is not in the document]
qweliant: wait, this is the wildest way to say we can order a query

server/spamTag/communities.ts
```diff
@@ -0,0 +1,83 @@
...
+					getWhereQueryPartForFreeformEntry(searchTerm),
+					getWhereQueryPartForUrl(searchTerm),
+				],
+			},
+		};
+	}
+	return {};
+};
+
+const getSpamTagWhereQuery = (status: undefined | types.SpamStatus[]) => {
```
qweliant, line 51: `getSpamTagStatusWherQuery`?

server/spamTag/communities.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+const getCommunityWhereQuery = (searchTerm: undefined | string) => {
+	if (searchTerm) {
+		return {
+			where: {
+				[Op.or]: [
+					getWhereQueryPartForFreeformEntry(searchTerm),
+					getWhereQueryPartForUrl(searchTerm),
+				],
```
qweliant, line 44: why use two queries when one always a subdomain?

qweliant (approved):

3mcd (approved):

server/spamTag/communities.ts
idreyn: Good question — the first one is comparing the `searchTerm` against directly against e.g. the `subdomain` field, so if  you type in `kfg-notes` then it will find that Community. The second is checking whether the `searchTerm` is a valid URL, and if so, it's extracting parts of that URL and checking _those_ against e.g. `subdomain`. So that'll match if you paste in `https://kfg-notes.pubpub.org` (or since we're also checking the `domain` field) `https://notes.knowledgefutures.org`.

PR #2396, opened 2022-12-06
🩹 fix: resize codeblocks + drop default language

howebenhowe:
Also split start script to allow quick-restart of dev environment when it mysteriously hangs and bump `caniuse` browserslist to hush the warnings.


idreyn (approved):
LGTM!

3mcd (approved):

Issue #2418, opened 2022-12-12
Tooltips for the quick access icons in Dashboard Dropdown Menu


djagdish:
### What's your enhancement request?
- Show tooltips for the li'l right-aligned icons in the Dashboard Dropdown Menu.
- Could be done with the Tooltip component, or as suggested by @idreyn, could use the `title` attribute.

### What would this enhancement help you accomplish?
- Better understanding of where clicking on the icons would lead to.

---

### Any attachments or reference images? (Drag them here)
<img width="491" alt="image" src="https://user-images.githubusercontent.com/316939/207126985-5c69a397-5c17-42c2-9e5d-3cdf9e9aab3b.png">

PR #2419, opened 2022-12-12
Adds Tooltips to Components

qweliant:
solves #2418

tooltips are now in the navigation dropdown 

image

### DIFF
--- a/client/components/ScopeDropdown/ScopeDropdown.tsx
+++ b/client/components/ScopeDropdown/ScopeDropdown.tsx
@@ -8,6 +8,7 @@ import { getPrimaryCollectionPub } from 'utils/collections/primary';
 import { sortByRank } from 'utils/rank';
 import { Collection, Pub } from 'types';
 import { pubPubIcons } from 'client/utils/icons';
+import { Tooltip } from '@blueprintjs/core';
 
 require('./scopeDropdown.scss');
 
@@ -178,7 +179,9 @@ const ScopeDropdown = (props: Props) => {
 												mode: 'settings',
 											})}
 										>
-											<Icon icon={pubPubIcons.settings} iconSize={12} />
+											<Tooltip content="View settings">
+												<Icon icon={pubPubIcons.settings} iconSize={12} />
+											</Tooltip>
 										</a>
 										<a
 											href={getDashUrl({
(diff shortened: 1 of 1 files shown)


client/components/ScopeDropdown/ScopeDropdown.tsx
```diff
@@ -178,7 +179,9 @@ const ScopeDropdown = (props: Props) => {
 												mode: 'settings',
 											})}
 										>
-											<Icon icon={pubPubIcons.settings} iconSize={12} />
+											<Tooltip content="View settings">
```
idreyn, line 182: I think it would be equally clear to drop "View" here (and elsewhere) and just say "Settings".

client/components/ScopeDropdown/ScopeDropdown.tsx
```diff
@@ -209,7 +216,9 @@ const ScopeDropdown = (props: Props) => {
 													mode: 'layout',
 												})}
 											>
-												<Icon icon={pubPubIcons.layout} iconSize={12} />
+												<Tooltip content="Visit collection">
```
idreyn, line 219: "Collection" should always be capitalized in user-facing text. Also, this button doesn't (or at least, shouldn't) take you to the Collection itself, but to the Collection's Layout tab. So it should just say "Layout".

client/components/ScopeDropdown/ScopeDropdown.tsx
```diff
@@ -231,7 +242,9 @@ const ScopeDropdown = (props: Props) => {
 												(scope.slugs && scope.slugs.pageSlug) || '/'
 											}`}
 										>
-											<Icon icon="globe" iconSize={12} />
+											<Tooltip content={`${scope.title} landing page`}>
```
idreyn, line 245: Pub titles can get _really_ long and probably aren't suitable for a tooltip. I'm not sure what this one should say. Maybe just "Visit"? I would defer to @djagdish.

idreyn (changes requested):
Thanks for the quick turnaround on this! There are some things we'll need to smooth out before shipping, though.

client/components/ScopeDropdown/ScopeDropdown.tsx
qweliant: "Home" maybe ???

idreyn (approved):
LGTM 👍

PR #2423, opened 2022-12-14
fix/enhancement: Mobile table styling

gabestein:
Adds some pretty meh table styling for mobile, mostly by setting `overflow-x: scroll` on the `tableWrapper` and a sane min-width for th and td (140px). I feel like there has to be a better way to do this, but the only other thing I can figure out is to set `white-space: nowrap` on table cells, which has the effect of...you guessed it...never wrapping, which is not good.

So, this effectively sets the width of every column to the min-width value (currently 140px), regardless of whether a custom width is set, which seems not great but potentially unavoidable (unless we set the table cells not to wrap at all).

Screenshot 2022-12-13 at 19 16 52

_Test plan_
1. Create a table and don't set any column or table widths
2. View on mobile
3. Make sure columns are at least 140px wide
4. Set the widths of some of the columns, and of the table itself (by adjusting the width of the last border in the table)
5. Make sure columns display at 140px wide

### DIFF
--- a/client/components/Editor/styles/table.scss
+++ b/client/components/Editor/styles/table.scss
@@ -1,10 +1,20 @@
+@import 'styles/mixins.scss';
+
 /* This clunky table, tableWrapper thing is because */
 /* isReadOnly removes the wrapping div, so margins don't */
 /* collapse as expected in all situations. */
 .tableWrapper {
 	margin: 1em 0;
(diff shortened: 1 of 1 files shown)


3mcd (approved):
Tested on my phone and emulator, works as expected.

Issue #2250, opened 2022-09-14
"Login to Create Pub" button does nothing


kyledenlinger:
_Before filing your bug report, please search existing issues to make sure a similar one doesn't already exist. If one does, please leave a comment or reaction in support of the issue rather than creating a new one._

_To fill out the report, please leave the below headers (**) in tact, and fill out each section. If a section isn't applicable, leave it blank or write n/a under it. Please do not delete any template sections._

### What went wrong, step-by-step?
1. Collection layout includes a submission button banner; button is set to "Create Pub." 
2. For users who are not logged in, the button text changes to "Login to Create Pub." 
3. Clicking that button does nothing.  
...

### What did you expect to happen?
Take me to the login or signup page

### What URL can we use to see the issue?
https://newsfeminism.pubpub.org/blog

### What browser/operating system were you using?
Chrome & Firefox / Mac

===

PR #2425, opened 2022-12-14
bug: Login to create pub 

qweliant:
addresses #2250 

now a redirect link is made

### DIFF
--- a/client/components/Layout/LayoutBanner.tsx
+++ b/client/components/Layout/LayoutBanner.tsx
@@ -74,14 +74,26 @@ class LayoutBanner extends Component<Props, State> {
 			this.props.content.buttonType || (this.props.content.showButton && 'create-pub');
 
 		const buttonText = getButtonText(buttonType, this.props.content.buttonText, isLoggedIn);
-		const buttonUrl =
-			buttonType === 'link'
-				? this.props.content.buttonUrl
-				: isLoggedIn && buttonType === 'create-pub'
-				? `/login?redirect=${this.props.locationData.path}`
-				: buttonType === 'signup'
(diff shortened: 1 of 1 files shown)


client/components/Layout/LayoutBanner.tsx
```diff
@@ -74,10 +74,11 @@ class LayoutBanner extends Component<Props, State> {
 			this.props.content.buttonType || (this.props.content.showButton && 'create-pub');
 
 		const buttonText = getButtonText(buttonType, this.props.content.buttonText, isLoggedIn);
+
 		const buttonUrl =
 			buttonType === 'link'
 				? this.props.content.buttonUrl
-				: isLoggedIn && buttonType === 'create-pub'
+				: !isLoggedIn && buttonType === 'create-pub'
```
kalilsn, line 81: These are just my style opinions, so feel free to ignore, but I liked your earlier commit that was unwinding these conditional ternaries. In practice, I think they're hard to read and it's really easy to lose track, especially with this 

IMO it's actually worth modifying this to be a little more explicit (either with parentheses, a more expressive indentation, or splitting it into a few separate conditional steps like you had done) to help prevent future mixups here since we don't have tests covering this.

kalilsn: Also if you do go back to the if/else version, I think you can drop that final `else` clause since after `let buttonUrl`, `buttonUrl` will already have the value `undefined`

kalilsn (commented):
Seems right to me, but I'd still like to let @howebenhowe approve as well since I'm new!

kalilsn (approved):

PR #2426, opened 2022-12-14
Render correct meta tags for authors with roles

### LINKED ISSUE #2352: Always display byline authors in citation_author metadata
1. Visited a Pub with byline authors who have roles
2. These authors did not show up in citation_author metadata.

Have them show up. This is required for Google Scholar inclusion.

https://research.arcadiascience.com/pub/result-comparative-crawling/draft

===

kalilsn:
Fixes #2352

- Reworks the <meta> elements for pub citations so that anyone with a byline will always either be cited as an author or an editor


server/utils/ssr.tsx
```diff
@@ -220,14 +220,21 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 	}
 
 	if (attributions) {
-		const authors = attributions.sort(sortAttributions).filter((item) => {
-			return item.isAuthor && !item.roles;
-		});
+		const authors: Attribution[] = [];
+		const contributors: Attribution[] = [];
 
 		const getPrimaryRole = (contributor: Attribution) => contributor.roles?.[0];
```
kalilsn, line 226: @gabestein I wanted to confirm two more things about how this _should_ work:

1. We're only considering the first role in the list
2. The `<meta name="dc.author">` tags should follow the same rules as the `<meta name="citation_author">` ones (create one for each author whose primary role is not a contributor role).

gabestein: 👍 to both!

One note for posterity: it does look like [Dublin Core](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/) has shifted its taxonomy to use `dc.creator` rather than `dc.author`, but I'm not sure how well supported this is by the tools consuming this metadata, so we should probably stick with author for now. At some point we may want to make the switch.

server/utils/__tests__/ssr.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { generateMetaComponents } from '../ssr';
+
+Enzyme.configure({ adapter: new Adapter() });
+const defaultProps = {
+	initialData: dataInitial,
+	attributions: pubData.attributions,
+	title: 'Test Pub',
+};
+
+const MetaComponents = (props) => <>{generateMetaComponents(props)}</>;
```
qweliant, line 15: maybe use a react fragment, but im being picky

kalilsn: `<></>` gets converted to a React.Fragment tag actually (which is why this line is needed). are you saying I should use the longhand `<React.Fragment>` version for readability?

qweliant: yeah but seriously dont if you dont want to

kalilsn: nope just making sure i understood! i think the shorthand is nice where there's lots of html but that's definitely not the case here

qweliant:
How do I run your new test? `npm run test server/utils/__tests__/ssr.test.tsx`? does not work, though I do not think I have ever run a snapshot test so it may be me

kalilsn:
> How do I run your new test? `npm run test server/utils/__tests__/ssr.test.tsx`? does not work, though I do not think I have ever run a snapshot test so it may be me

I'm running `npm run test-dev server/utils/__tests__/ssr.test.tsx` which works for me. What error are you seeing?

qweliant (approved):

3mcd (approved):
LGTM!

PR #2434, opened 2022-12-15
update: cleanup some tools

gabestein:
- Makes my exportCollection tool actually useful by recursing until all the exports are completed.
- Updates Eric's encrypt tool to be usable via npm run tools, and makes the input a more explicit `--text=text`.


gabestein:
whoops, just pushed a rebased version

Issue #2351, opened 2022-11-15
Column-width wrapping for Inline equations


Dawit-Tegbaru:
In-line equation looks fine on laptop but bleeds off screen when viewing on mobile.  

### What URL can we use to see the issue?
https://jbt.pubpub.org/pub/b0nisrdj/release/1?from=18905&to=18931

### What browser/operating system were you using?
Chrome / Android

===

### Have any screenshots? (Attach them by dragging the image here)

Here's the eq. on laptop: 

![image](https://user-images.githubusercontent.com/72410195/201976780-992b003a-b5b4-4317-a307-198210cb9796.png)

On mobile:

![Screenshot_20221115-105147_Chrome](https://user-images.githubusercontent.com/72410195/201979544-e109d566-1348-4226-87b7-674499eaae21.jpg)

gabestein:
Did a little research — KaTeX support just isn't there (yet?) for block equations (https://github.com/KaTeX/KaTeX/issues/1023). 

There are a few workarounds discussed in [this thread](https://github.com/KaTeX/KaTeX/issues/327) that I think might be worthwhile. The simplest/easiest/most aesthetically viable may be allowing overflowed equations to be scrolled until KaTeX figures out how to break block math effectively.

PR #2435, opened 2022-12-15
Improve mobile experience for content wider than the screen

kalilsn:
#2351 

This PR fixes the CSS intended to prevent pubpub from being wider than the screen, which didn't work on my browsers. Now if an element (really just math as far as I can tell) is wider than the viewport, it will have internal scrollbars, but the page itself will not expand.

This pub has some examples of content that overflows on small screens:
[duqduq version](https://demo.duqduq.org/pub/hilwxpnd/release/4)
[review app version](https://pubpub-pipel-kalilsn-ov-bfcnoe.herokuapp.com//pub/hilwxpnd/release/4)

Feedback I'm looking for:
- Are there other places where we might need to add an explicit `overflow-x: auto`? It seems like most elements from the editor get wrapped in a `<p>` tag, but I'm not so sure about other parts of the site.
- At intermediate sizes where the discussions are on the right side, should the wide content overflow into the sidebar the way images can?
- Does this need some kind of indicator to let users know that these elements can be scrolled (since it seems like scrollbars are usually hidden by default and we can only style those in webkit browsers)?

### DIFF
--- a/client/containers/App/app.scss
+++ b/client/containers/App/app.scss
@@ -5,7 +5,7 @@
 	flex-direction: column;
 	min-height: 100vh;
 	width: 100%;
-	overflow-x: hidden;
+	overflow-x: clip;
 	#main-content {
 		outline: none;
 		flex-grow: 1;
--- a/client/containers/App/app.scss
+++ b/client/containers/App/app.scss
@@ -5,7 +5,6 @@
 	flex-direction: column;
 	min-height: 100vh;
 	width: 100%;
-	overflow-x: clip;
 	#main-content {
 		outline: none;
 		flex-grow: 1;
--- a/client/containers/App/app.scss
+++ b/client/containers/App/app.scss
@@ -4,6 +4,8 @@
 	display: flex;
 	flex-direction: column;
 	min-height: 100vh;
+	width: 100%;
+	overflow-x: hidden;
 	#main-content {
 		outline: none;
 		flex-grow: 1;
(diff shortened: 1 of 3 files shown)


gabestein (changes requested):
Mostly looks good from my perspective! Thanks for taking a holistic look at this. 

One small regression to address: the addition of overflow-x on #app appears to have broken sticky editor functionality, unfortunately. I'm not sure exactly how to solve, but as you noted, it may not strictly be needed because you've also set it on editor `p` tags.

On your questions
- I can't think of explicit places that need overflow except the p tags...I added some embeds and iframes, and all works as expected, as do tables with the caveat that I just pushed a mobile change, but I don't think there are any conflicts (see below).
- I'll leave the intermediate widths determination to @idreyn.
- I don't think we need visual indicators of overflow, at least for now; for one thing, we've struggled in the past to build them in a way that feels good in places like the header. For another, I think it's starting to become more widely understood that that's the experience on mobile webkit.

One non-blocking request prior to pushing, once others have had a look: please take a quick peek at master, where I just merged in a mobile tables fix (https://github.com/pubpub/pubpub/pull/2423) to make sure there are no conflicts. Taking a look at it myself locally, though, I couldn't find any. It seems to work fine.

kalilsn:
> I'll leave the intermediate widths determination to @idreyn.

I've updated this to only apply at small widths after seeing how the sidebar content interacts with overflowing elements (basically fine on a computer IMO).

> One small regression to address: the addition of overflow-x: hidden on #app appears to have broken sticky Pub editor functionality, unfortunately.

It's true, I could just leave this off if we don't think it's useful. For now, I've used overflow: clip, which also fixes the sticky issue, but won't work in older versions of safari, in which case the header will just be static .

kalilsn:
I'm merging this as-is, although I've noticed on Firefox that these wide images do cause a small horizontal scroll effect. If we find that comes up more in the future we should bring back `overflow-x: clip`!

Issue #2383, opened 2022-12-05
Not all Footnotes Shown below Pub


MaPeras:
_Before filing your bug report, please search existing issues to make sure a similar one doesn't already exist. If one does, please leave a comment or reaction in support of the issue rather than creating a new one._

_To fill out the report, please leave the below headers (**) in tact, and fill out each section. If a section isn't applicable, leave it blank or write n/a under it. Please do not delete any template sections._

### What went wrong, step-by-step?
1. Imported docx-file with 245 footnotes including references.
2. Import worked.
3. Below the document (footer) only 235 footnotes appear.
...

### What did you expect to happen?

All 245 footnotes should appear in the footer as there is no difference between them.

The only difference might be, that some footnotes within the docx were added via the Zotero-Plugin and some were added manually.

### What URL can we use to see the issue?

https://informationsfreiheit.pubpub.org/pub/7wt3s8fg/release/1#iii.-bekanntgabe 

### What browser/operating system were you using?

Firefox / Linux
Firefox / Windows
Chromium / Linux

===

### Have any screenshots? (Attach them by dragging the image here)

![Screenshot from 2022-12-05 14-25-14](https://user-images.githubusercontent.com/49619613/205648269-3ee1c468-8618-4bb3-a7d1-d9607bf5ea69.png)

PR #2439, opened 2022-12-19
bug: Footnotes Won't be Deduped

qweliant:
addresses #2383 

Now, all of the footnotes will show

test plan: using this [doc](https://docs.google.com/document/d/1qjnN17jUp8_-iVXVi7ojH0XYzce0MoxwvEOS_oORPs0/edit?usp=sharing) you can test that there are 396 footnotes when importing to pubpub

before there would have only been 3

### DIFF
--- a/utils/notes.ts
+++ b/utils/notes.ts
@@ -40,12 +40,17 @@ const getAlphabetizableValue = (note: RenderedNote) => {
 		.trim();
 };
 
-const getRenderedNotes = (
-	notes: Note[],
-	renderedStructuredValues: RenderedStructuredValues,
-	includeNumbers: boolean,
-): RenderedNote[] => {
-	return unique(notes, (note) => note.fingerprint).map((note, index) => {
+type RenderNoteOptions = {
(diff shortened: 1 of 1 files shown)


utils/notes.ts
```diff
@@ -40,12 +40,17 @@ const getAlphabetizableValue = (note: RenderedNote) => {
...
+type RenderNoteOptions = {
+	notes: Note[];
+	renderedStructuredValues: RenderedStructuredValues;
+	includeNumbers: boolean;
+	isFootnote: boolean;
+};
+
+const getRenderedNotes = (options: RenderNoteOptions): RenderedNote[] => {
+	const { notes, renderedStructuredValues, includeNumbers, isFootnote } = options;
+	const obj = isFootnote ? notes : unique(notes, (note) => note.fingerprint);
```
idreyn, line 52: I think it would be nice to use a descriptive variable name here.

idreyn (approved):
LGTM!

PR #2440, opened 2022-12-19
Return 404 for base pubpub /rss.xml

### LINKED ISSUE #2432: Respond to requests to pupub.org/rss.xml with 404
1. visit https://pubpub.org/rss.xml
2. get 500

404

Bots crawl this path a lot, and it results in 500s that sometimes cause our heroku alerts to trigger.

===

kalilsn:
Fixes #2432 

- Handles the missing community id in the initial data when on base pubpub

### DIFF
--- a/server/rss/__tests__/api.test.ts
+++ b/server/rss/__tests__/api.test.ts
@@ -29,3 +29,9 @@ it('respects query parameters', async () => {
 		.expect(200);
 	expect(countItems(body.text)).toEqual(2);
 });
+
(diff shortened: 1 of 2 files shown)


idreyn (approved):
Nice, LGTM!

fwiw it seems like the idiom for this is the `hostIsValid(req, 'community')` that is sprinkled all over our routes. We really need an abstraction over Express routes to manage the things that almost all of our routes do (and we thus mindlessly copy-paste when a new one is created).

Issue #2442, opened 2022-12-20
If a pub is in a collection, only the collection title ends up in the `citation_title` and `dc.title` meta tags


tefkah:
Heyyy, I'm back again to complain about meta tags, sorry 🙆

### What went wrong, step-by-step?
1. Create pub
2. Add pub to collection (issue here)
3. Title of the pub does not show up in the `citation_title` etc  meta tag.
<img width="986" alt="image" src="https://user-images.githubusercontent.com/21983833/208717476-3d36015a-a7fd-4040-9e0e-39652e2255b9.png">

This is probably due to the earlier changes (on my request, sorry!) here https://github.com/pubpub/pubpub/pull/2307 

I will make a PR to (hopefully) rectify this. Unfortunately it's hard for me to check whether this works!

### What did you expect to happen?
The title shows up in the meta tag

### What URL can we use to see the issue?
https://journal.trialanderror.org/pub/systemic-academia/release/2

===

### Have any screenshots? (Attach them by dragging the image here)
<img width="700" alt="image" src="https://user-images.githubusercontent.com/21983833/208701342-36a75ce5-d15f-466a-8e33-2925359f90f1.png">

gabestein:
Thanks for catching this! I've proposed what I think is a bit of a cleaner version in #2450. Please take a look and feel free to comment. Basically, what I think is sanest (until we do a refactor of the meta entirely) is:

- If the thing is a pub, we use solely the pub title in og:title, dc:title, and citation_title.
- Otherwise, we use solely the collection title for those three.

PR #2307, opened 2022-10-27
fix/update: add metadata for collections and fix filter

gabestein:
Improves Collection-level metadata for Zotero/Mendeley/etc. and fixes a bug where DOIs were appearing twice in collection headers.

_Test plan_
- Visit any collection with a DOI and make sure it only shows once in the collection header block
- Visit Conference and Issue collections with all metadata filled in. Make sure: citation_doi and citation_date HTML metadata are set according to the collection metadata.
- Visit a Book collection. Make sure citation_title, og:title, citation_book_title, citation_DOI are set according to metadata. Make sure og:type is set to 'book'.
- Visit a tag collection and make sure no metadata is set except basic og metadata.
- Visit pages and pubs in and out of collections and make sure they load as expected.

Thanks to @tefkah for initial pull request and fixes.

PR #2443, opened 2022-12-20
fix(#2442): set citation_title to collection title only if it is not a pub

tefkah:
Because of changes introduced in #2307, Pubs in a Collection will only show the collection title in the `citation_title` etc meta tags, see #2442.

This PR introduces two variables which check whether a page is a Pub, and whether it is a book and thus needs to have it's citation_title and dc.title set to the name of the book.

If the latter is the case, that is done. 

This is not the most elegant way to do this, probably the `og:title` should not be said like this as well.


gabestein:
Thanks for flagging! Closing in favor of #2450.

PR #2450, opened 2023-01-03
Fix: title handling in meta tags

gabestein:
Due to a change in #2307, some Pubs are being given their collections' `dc.title`, `og:title`, and `citation_title`. This lead to some citation managers and scrapers grabbing the collection title instead of the pub title. The behavior we want is:

- If the resource is a Pub, we use its title (without context) for all three.
- If the resource is a collection that has a title, we use the collection title (without context).

This PR accomplishes that.

_Test plan_
1. Visit a pub in a collection. Verify that:
- The page title and twitter title are titleWithContext (`pubTitle • communityTitle`).
- The og:title and dc.title are the pub title (without context)
1. Visit a pub not in a collection. Verify that it loads and that the titles are the same as above.
1. Visit a collection page. Verify that it loads and:
- The page title and twitter title are titleWithContext (usually `collectionTitle • communityTitle`).
- The og:title and dc.title are the collection title (without context)
1. Visit a few other non-pub and non-collection pages (search, homepage, etc.) and verify that the titles are as expected (`PageTitle • CommunityTitle` for all)

### DIFF
--- a/server/utils/ssr.tsx
+++ b/server/utils/ssr.tsx
@@ -68,7 +68,7 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 	} = initialData.communityData;
 
 	const url = `https://${initialData.locationData.hostname}${initialData.locationData.path}`;
-	const isPub = !!initialData.scopeData.elements.activePub;
+	const isPub = !!initialData.scopeData?.elements.activePub;
 	const favicon = initialData.communityData.favicon;
 	const avatar = image || initialData.communityData.avatar;
 	const titleWithContext = contextTitle ? `${title} · ${contextTitle}` : title;
--- a/server/utils/ssr.tsx
+++ b/server/utils/ssr.tsx
@@ -69,6 +69,7 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 
 	const url = `https://${initialData.locationData.hostname}${initialData.locationData.path}`;
 	const isPub = !!initialData.scopeData?.elements.activePub;
+	const useCollectionTitle = !isPub && collection?.title;
 	const favicon = initialData.communityData.favicon;
 	const avatar = image || initialData.communityData.avatar;
 	const titleWithContext = contextTitle ? `${title} · ${contextTitle}` : title;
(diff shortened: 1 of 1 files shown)


server/utils/ssr.tsx
```diff
@@ -68,6 +68,7 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 	} = initialData.communityData;
 
 	const url = `https://${initialData.locationData.hostname}${initialData.locationData.path}`;
+	const isPub = !!initialData.scopeData.elements.activePub;
```
gabestein, line 71: the check on scopeData is needed because, I believe, the test data in ssr.test.tsx is out of date and does not include scopeData. We could fall back to checking the url for the existence of `/pub/` for determining if it's a pub if that feels better.

server/utils/ssr.tsx
```diff
@@ -68,6 +68,7 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 	} = initialData.communityData;
 
 	const url = `https://${initialData.locationData.hostname}${initialData.locationData.path}`;
+	const isPub = !!initialData.scopeData?.elements.activePub;
```
gabestein, line 71: the check on scopeData is needed because, I believe, the test data in ssr.test.tsx is out of date and does not include scopeData. We could fall back to checking the url for the existence of /pub/ for determining if it's a pub if that feels better.

tefkah:
Looks good to me! I was indeed unsure about the `activePub` prop, but that seems to be a cleaner solution to me yeah.

server/utils/ssr.tsx
kalilsn: The test data is definitely out of date (as are the types for initialData). IMO it feels a little wrong to make changes to the app code to account for bad test data, but this seems fine.

server/utils/ssr.tsx
```diff
@@ -92,12 +93,16 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 			<meta
 				key="t2"
 				property="og:title"
-				content={collection?.kind === 'book' ? collection.title : title}
+				content={!isPub ? collection?.title || title : title}
```
kalilsn, line 96: Feel free to ignore these opinions, but I find the combination of the ternary and the short circuiting `||` in this expression to be a little taxing to read. `!isPub && collection?.title ? collection?.title : title` is a bit easier on my eyes personally. I also think this section would scan a bit better if it were extracted to a variable either way!

kalilsn (approved):
Looks good to me!

server/utils/ssr.tsx
gabestein: thanks! Definitely see what you mean. Lmk what you think of the update I just pushed. If nothing else, I think the extra variable helps to understand the intent here much more.

server/utils/ssr.tsx
gabestein: Yeah, agreed. I tried to add scopeData to the test data from the storybook but that seemed to cause more problems than it solved, so I punted. I've added a ticket to remind us to update this.

server/utils/ssr.tsx
kalilsn: update looks great, much clearer!

kalilsn (approved):

Issue #2445, opened 2023-01-03
/legal/random should return 404


gabestein:
### What went wrong, step-by-step?
1. visit https://pubpub.org/legal/blah
2. get 200

### What did you expect to happen?
Should return 404

===

PR #2453, opened 2023-01-04
404 for nonexistent tabs under /legal

kalilsn:
Addresses #2445 

Explicitly specifies the routes under `/legal` that pubpub responds to ~since we weren't using the `:tab` parameter anyways.~ <-- untrue as Gabe pointed out below.

To test:

Working legal pages: https://pubpub-pipel-kalilsn-le-xv3ufv.herokuapp.com/legal/terms

404 from a bad route under legal: https://pubpub-pipel-kalilsn-le-xv3ufv.herokuapp.com/legal/term


gabestein (commented):
Thanks! It might be environment, but when I visit those pages locally or in the review app environment, the pages are blank. I wonder if the routes require the `:tab` variable to be passed in via express to function?

PR #2457, opened 2023-01-10
feature: Table breakout (and smaller font)

### LINKED ISSUE #754: Width: Allow tables to pop out of the central column
As an author or editor, I often have text, tables, or images wider than the default column.  

For _text_, smaller fonts are needed; I occasionally use superscript as a hack, but expect that to break in the future.

For _images_ and other embeds, there's the handy option to pop out media or iframe and set the width percentage.  Ideal -- though I sometimes want a borderless result.

For _tables_, I want a combination of the above:  control over width and control over text size.  Often the underlying table has a minimum reasonable width in ems (proportional to the font size), so at least one of size/width needs to be flexible, preferably both.   

As with images and embeds, a table caption should align / center with the table.

idreyn:
Closes #754

This PR adds two new features to tables:

- A "smaller font" option that makes table text 20% smaller
- A "breakout" menu that lets tables break out of the Pub body the same way figures can, at some set fractions of viewport sizes.

These appear as two new menu items in the hovering table menu:

image

Table breakouts at 50% and 75% look kind of funny on smaller viewports (and they revert back to the Pub column on mobile layouts). But we talked it over and decided we're going to introduce this to the editor as-is and see what people say.

This PR is quite dense with changes — I had to do some refactoring to the formatting bar state management to permit the new nested breakout menu, but I think it makes the code easier to understand.

_Test plan:_
- Create a table in a Pub
- Make sure you can toggle the "smaller font" option
- Make sure you can set the table breakout value
- Also just generally test the formatting bar, since I did a bit of refactoring there.


client/components/Editor/views/TableView.ts
```diff
@@ -0,0 +1,48 @@
+// @ts-expect-error (TableView is exported by this package but not present in its .d.ts definitions)
```
idreyn, line 1: This was moved from `plugins/table.ts` to match our newer convention of putting `NodeView` classes in this directory.

client/components/Editor/plugins/table.ts
```diff
@@ -22,49 +21,6 @@ function wrapDomSerializer(domSerializer: DOMSerializer) {
 	});
 }
 
-/**
```
idreyn, line 25: Moved to `views/TableView.ts`.

kalilsn (approved):
The refactoring makes sense to me and it looks like this works as intended, except that it has some issues when combined with explicitly set column widths, which we'll address in the future.

metasj:
This changes everything.  Looks great :)

Issue #2455, opened 2023-01-09
Heroku router timeouts


gabestein:
We've been seeing regular (a few times per day) spikes of H12 timeouts from the heroku router - meaning our application didn't send any data before the 30s timeout and the router returns a 503.

To start getting a better sense of what's causing these, we should set up Sentry's distributed tracing and performance metrics, as well as set an explicit timeout and warning for slow requests so we can find out what's going on.

PR #2467, opened 2023-01-19
Sentry update and timeout warning

kalilsn:
#2455 

This PR has a few small updates to our sentry config intended to get some insight into these timeout errors (which are coming from the heroku router, and therefore not captured by sentry).

- Updates to latest sentry versions
- Switch to sentry react package over sentry browser
- Adds a message to sentry when no response is sent after 25s
- Move frontend to its own sentry project


client/components/Editor/plugins/malformedDoc.ts
```diff
@@ -60,15 +60,8 @@ export default () => {
...
-					try {
-						throw new Error('Uh-oh!');
-					} catch (err: any) {
-						theError = err;
-					}
-					Sentry.captureEvent({
-						message: 'Evil Prosemirror Transaction',
-						contexts: { transactionMetadata, error: theError },
+					Sentry.captureException(new Error('Evil Prosemirror Transaction'), {
+						contexts: { transactionMetadata },
```
kalilsn, line 64: This refactor was to fix a type error. I think it's mostly equivalent but let me know if I misunderstood the goal of this block!

idreyn: No, thank you for removing this ridiculous little bit of code!

idreyn (approved):
LGTM and I'm excited to get closer to the heart of the timeout mystery!

client/components/Editor/plugins/malformedDoc.ts
howebenhowe: I have re-read the previous implementation here a few times and can't reason why there was a try-catch block there. It reads like a code-koan.
"Try to fail, and if you do, you have your answer, so continue."

howebenhowe (approved):
LGTM!

qweliant (approved):
LGTM

PR #2470, opened 2023-01-23
fix: Make sure ActivityItem is created with new Discussion

### LINKED ISSUE #2469: `pub-discussion-comment-added` ActivityItem not created for new Discussions
The refactoring from #2289 reorders some object creation so that a `Thread` and a `ThreadComment` before associating it with a `Discussion`. When the `ThreadComment` is created, the Sequelize hook that fires can't find a parent discussion and doesn't create the activity item.

Fix and add a test to catch this regression in the future.

idreyn:
Resolves #2469. Back in #2289 we added a new abstraction that creates a new `Thread` and a `ThreadComment` in the same query function. This is a really good idea, and for months I didn't notice a small problem. When a `ThreadComment` is created, a [Sequelize hook](https://sequelize.org/docs/v6/other-topics/hooks/) fires that wants to create a `pub-discussion-comment-added` item. But unless the parent thread is already associated with a `Discussion`, we can't do that, and no item is created. So we really need to create a Discussion and a Thread and associate the two before we create a comment.

There are other potential fixes here (deferring the hook somehow, or maybe adding another hook to `Discussion.afterCreate`) but for now I've elected to re-order the creation of these items, and add a test that checks that the activity item is created so we'll know if this breaks again.

_Test plan:_

Run the unit tests at:
```
npm run test-dev server/discussion/
```

You can test manually by commenting on a Pub, then visiting the Activity tab of that Pub's dashboard and verifying that a record of your comment exists there.

### DIFF
--- a/server/activityItem/fetch.ts
+++ b/server/activityItem/fetch.ts
@@ -198,7 +198,9 @@ const getActivityItemAssociationIds = (
 			thread.add(item.payload.threadId);
 			if (item.payload.threadComment) {
 				threadComment.add(item.payload.threadComment.id);
-				user.add(item.payload.threadComment.userId);
+				if (item.payload.threadComment.userId) {
+					user.add(item.payload.threadComment.userId);
+				}
 			}
 		} else if (item.kind === 'pub-review-comment-added') {
 			review.add(item.payload.review.id);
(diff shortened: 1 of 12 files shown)


server/activityItem/fetch.ts
```diff
@@ -198,7 +198,9 @@ const getActivityItemAssociationIds = (
 			thread.add(item.payload.threadId);
```
idreyn, line 198: These few changes relating to `threadcomment.commenterId` are not strictly related to the fix — they're downstream from a change to our TS `ThreadComment` type where we now assert that both `userId` (for a real PubPub user) or `commenterId` (for a guest commenter) can be nullable.

qweliant (approved):

Issue #2262, opened 2022-09-22
Allow custom CSS to target expanded discussion highlight color


gabestein:
### What's your enhancement request?
When you click to open a discussion, the color of the highlight defaults to the light accent color of the community. That's fine in most cases. But the color is added by injecting the CSS inline into the style when the discussion is opened. Thus, there's no current way to override only that color — you can override the color for all highlights in all states, but not in open/closed. It would be nice to have a way to target discussion color based on state.

<img width="1404" alt="Screen Shot 2022-09-22 at 13 33 44" src="https://user-images.githubusercontent.com/639110/191835677-bf42d673-fd90-4f71-aaf7-104ac55a34ee.png">

<img width="1324" alt="Screen Shot 2022-09-22 at 13 36 10" src="https://user-images.githubusercontent.com/639110/191835715-2cd3f76b-6df9-4f13-a74c-063832379791.png">

### What would this enhancement help you accomplish?
Set custom colors for discussion highlights in various states.
---

PR #2472, opened 2023-01-23
Custom CSS Targets `ActiveThread` background

qweliant:
Addresses #2262 

adds a css variable `--pubpub-active-discussion-highlight-color` to the style script in `DiscussionNav` that can be used to override the discussion highlight color.

test plan:
Create a custom css selector
`:root {
  --pubpub-active-discussion-highlight-color: your-color(but we all know chartreuse is the best);
}`

navigate to a pub, create an inline discussion, then expand the discussion bubble


idreyn (approved):
LGTM!

Issue #2477, opened 2023-01-25
sequelize models refactor


howebenhowe:
With an eye toward sequelize 6/7, and generally towards a consistent `model.ts` convention, clean up our model definitions.

## server-side work
+ give privilege to associations over explicitly declaring foreign key fields (e.g. don't include `userId` as an attribute, but use `[model].belongsTo(User);` instead, which creates the fk for you)
+ move `[foreignKey]: { allowNull: false }` constraints to the associations
+ use `[attribute]: dataTypes.[typeName]` shorthand over `[attribute]: { type: dataTypes.[typeName] }`
+ lowercase first letter of model names so that generated `foreingId`'s generate with lc first letter
+ explicitly declare the `tableName: [CapitalizedPluralizedModelName]` to preserve DB unchanged
+ create counterpart associations (e.g. for every `hasMany`, include the `belongsTo` on the target model)

## stretch
+ ensure all associations have both sides of the association explicitly declared

## testing
+ api test suite runs clean
+ no changes to DB schema

thorough testing will require local DB spin-up for side-by-side comparison

PR #2478, opened 2023-01-25
refactor sequelize model definitions

howebenhowe:
addresses #2477

Sweeping cleanup + trimming of model definitions.
does not yet guarantee implementation of both sides of associations

__test plan__
run api test script with `npm run test-dev`

__completed test__
together with @kalilsn I spun up a local database with current tables as created from `master`'s sequelize definitions, dumped that database, then created db again from this branch's definitions, dumped, and diffed the dump files. The table specifications were identical.


3mcd:
Again, closing for now. I'll leave the branch up and we can decide what to do with it later.

PR #2486, opened 2023-02-01
dev: Giant README for facets

idreyn:
This PR has the big `README.md` for the `facets` directory. As I was writing, I realized that we don't ever use the `defaultValue` associated with a facet prop. I think it's clearest for now to omit this option, and make `createEmptyFacetInstance()` return all-null props. So I removed the `defaultValue` option.

I also found some parameters and types and stuff that could have better names, and so I went in and changed them. The most consequential of these is changing `FacetCascadedType<T>` to `FacetValue<T>`. I think that's an improvement, since we now have three distinct but relatively easy to remember names for related things:

- facet _definitions_
- facet _instances_
- facet _values_

Which, I hope the readme will explain 😆

_Test plan:_
If this compiles and the tests pass, we should be good to go. I'm confident that removing `defaultValue` will have no effect on production code.


pilotSaad:
"Facets" is a term used in different contexts and can refer to various things. Can you provide more information or clarify what you mean by "facets"? This will help me better understand what you're looking for in a ["Giant README."](<a href="https://northlandz.com/outdoor-trains">outdoor train<a)

3mcd (approved):

Issue #2475, opened 2023-01-25
Create URL Paste Input Rule for PM


qweliant:
# Overview

ProseMirror requires regex matching and input rules for capturing pasted links in the editor. This is not clear, but there are examples and a few discussions about it. 

https://github.com/ProseMirror/prosemirror/issues/90#issuecomment-202933072

https://discuss.prosemirror.net/t/paste-behavior-research/469/37

## TODO

Add new input rule to handle auto linking a pasted URL in the editor.

PR #2487, opened 2023-02-01
Autolink on Paste

qweliant:
Addresses #2475 

When pasting into the editor with links they will be auto created


3mcd:
We are considering holding off on this PR until users are comfortable with the auto-link on type feature. Pasting a link and pressing "space" will instantly linkify the text, so this plugin may not be necessary if users are comfortable with the current behavior.

PR #2490, opened 2023-02-02
dev: Remove remnants of old Pub fields

idreyn:
One more for old times' sake!

This just removes a bunch of the fields on the `Pub` model that were moved to facets (and a few more that appear to have fallen out of use). The Pub model is looking pretty clean and generic and Scope-like these days. Food for thought! ;)

This code doesn't need to be deployed as such. Instead, you can run the command below at your leisure once this hits `master`, and then a second time with `PUBPUB_PRODUCTION=true` appended to hit prod. Then other branches will need to merge this code in to keep their `server/pub/model.ts` in sync.

_Test plan:_

I ran:
```
npm run tools migrate -- --name 2023_02_02_removeOldPubFields
```

and observed that it removed all the fields. I ran it again with `--down` and verified that the reverse migration (which we hopefully will not need) also works.


workers/tasks/import/bulk/README.md
```diff
@@ -456,20 +456,8 @@ _The remainder of the options map directly to attributes on the Pub model._
 
 **`avatar: string`**
 
-**`headerStyle: string`**
```
idreyn, line 459: We don't have a replacement for these yet, but I don't think we've ever actually used them in the bulk importer.

3mcd (approved):

Issue #2413, opened 2022-12-12
zotero: OAuth UI in User settings


howebenhowe:
In user privacy settings, display Zotero OAuth status with button to revoke if active, or activate if authentication token is absent.

Immediate follow for #2414, necessary for their mutual iterative testing.

gabestein:
may only need revoke settings

Issue #2408, opened 2022-12-12
picker: live search results from zotero


howebenhowe:
Per [the spec](https://notes.knowledgefutures.org/pub/ck6f4zsx#alternative-live-search-results), pass typed queries to the linked zotero account in real time, surfacing in a dropdown the best matches.

If authentication has  been revoked, doesn't exist, or results in error for any other reason, prompt user to re-authenticate.

First incarnation, drop it immediately into the current citations builder:

![Image](https://user-images.githubusercontent.com/14115194/211365183-fc505521-be4d-49ae-9209-cc4983e37cc0.png)

PR #2520, opened 2023-02-09
Bh/citations oauth

### LINKED ISSUE #2414: zotero: OAuth key exchange
Notes:
If authentication token is absent or invalid, re-prompt user to authenticate with Zotero.

### implementation steps:
1. register application with zotero
2. add pubpub's client key and secret key to `config.js`
3. set up callback url for redirection from zotero
4. add ExternalLoginProvider model for storing authentication token, linked to User model (following [this model](https://vertabelo.com/blog/user-authentication-module/))
5. implement standard oauth handshake pattern [per zotero's criteria](https://www.zotero.org/support/dev/web_api/v2/oauth)

### new dependencies
one of the [TS](https://github.com/retorquere/zotero-sync) or [JS](https://github.com/tnajdek/zotero-api-client) clients (probably the latter)
investigate tradeoff with passport or oauth.io for credential storage

### Server side implications
+ change user model to hold:
  + OAuth long-lived token
  + zotero user id
+ new routes for OAuth callback urls

### Testing
1. Successfully grant OAuth access to PubPub account
2. Abandon then resume OAuth handshake
3. After success, revoke authentication

howebenhowe:
## Issue(s) Resolved
#2414
#2413 

+ Introduces `Integration` and `IntegrationData[auth schema name]` models which store auth data serverside.
+ Adds new route for authentication at `/auth/zotero`
+ Adds 'activate'/'delete' integration buttons to the privacy settings at `/legal/privacy`
+ introduces a js zotero client, which will be developed further in #2408

## Test Plan
in order to test this in dev, you will need to add `process.env.ZOTERO_CLIENT_KEY` and `process.env.ZOTER_CLIENT_SECRET` variables to your `config.js` (I saved these to the onepass secure note for config at the top of the 'global' section), as well as get the login credentials for the knowledge futures zotero test account (saved to the pubpub vault on onepass)

test creation:

1. log in
2. navigate to user privacy settings
3.  click 'activate' for zotero integration
4. proceed through oauth flow, including login to zotero and return to privacy settings
5. back on privacy settings, confirm the zotero integration offers option to delete integration

test deletion:

1. following the above, get to privacy settings
2. click 'delete' for the zotero integration
3. it should immediately switch apparent status in UI
4. refresh page to confirm the integration is deleted

more granular testing available by observing rows in the tables `Integrations` and `IntegrationDataOAuth1` in dev DB.

Screenshot from 2023-02-08 22-35-34

### Notes/Context/Gotchas
repeat visits to the auth url (i.e. `/auth/zotero` should refresh the authentication token in the `IntegrationDataOAuth1` row


server/integration/queries.ts
```diff
@@ -0,0 +1,58 @@
+import * as types from 'types';
+
+import { Integration } from 'server/models';
+import { upperCaseFirstLetter } from 'server/integration/queryHelpers';
+
+type CreateOptions = {
+	name: string;
+	userId: string;
+	integrationId: string;
+	authSchemeName: string;
```
3mcd, line 10: Should this be of type `AuthSchemeName`?

server/integration/utils/auth.ts
```diff
@@ -0,0 +1,74 @@
...
+					model: Integration,
+					where: { name: 'zotero' },
+					required: false,
+					include: { model: IntegrationDataOAuth1, required: false },
+				},
+			})
+				.then((user) => {
+					if (!user.integrations.length) {
+						return user
+							.createIntegration({
```
3mcd, line 47: Is this a sequelize magic method? I'm not sure if I've seen them used in the PubPub codebase. I recommend doing the old fashioned `Integration.create({ userId: user.id })`.

server/integration/utils/auth.ts
```diff
@@ -0,0 +1,74 @@
...
+							})
+							.then((integration) =>
+								integration.createIntegrationDataOAuth1(integrationDataOAuth1),
+							);
+					}
+					const integration = user.integrations[0];
+					return integration
+						.getIntegrationDataOAuth1()
+						.then((oldData) => {
+							if (!oldData) {
```
3mcd, line 60: Would this ever happen? i.e. would a "zotero" integration ever be created without an `IntegrationDataOAuth1`? If not, we may want use the `assert` utility here to assert that `oldData` is always truthy. That way we're throwing an error intentionally if the integration ends up in an invalid state.

server/integration/utils/auth.ts
```diff
@@ -0,0 +1,74 @@
...
+								authSchemeName: 'OAuth1',
+								externalUserData,
+							})
+							.then((integration) =>
+								integration.createIntegrationDataOAuth1(integrationDataOAuth1),
+							);
+					}
+					const integration = user.integrations[0];
+					return integration
+						.getIntegrationDataOAuth1()
```
3mcd, line 58: Same here. IMO we should avoid using magic methods, since they are harder to statically type, can be misspelled easily, etc.

server/integration/model.ts
```diff
@@ -0,0 +1,19 @@
+export default (sequelize, dataTypes) =>
+	sequelize.define(
+		'integration',
+		{
+			name: dataTypes.TEXT,
+			authSchemeName: dataTypes.ENUM('OAuth1'), // this frames polymorphism, add schema names here and in corresponding type
```
3mcd, line 6: What are your thoughts on simplifying this a bit? Maybe a user could have an association with a `ZoteroIntegration` that has an association to an `OAuth1Integration` table that stores Zotero's OAuth1-related state. This way we're still storing all OAuth1 integrations in the same table, but each specific integration get its own base table?

The reason I bring this up is because polymorphic relations like this in SQL seem kind of like an antipattern as we aren't able to enforce that an `Integration` can have at most a single `IntegrationData*` association.

client/components/FormattingBar/controlComponents/ControlsFootnoteCitation/ControlsFootnoteCitation.tsx
```diff
@@ -29,6 +31,12 @@ type Props = {
 	};
 };
 
+const fetchCitations = async (query) => {
+	const { items } = await apiFetch(`/api/citations/zotero?q=${query}`);
+	console.log('bump', { items });
```
3mcd, line 36: Leftover log statement here.

3mcd (changes requested):

server/integration/model.ts
howebenhowe: I'm tracking the thinking, but I believe we have the tooling to enforce the single `IntegrationData*` association.
the model for `Integration`:
```
...
authSchemeName: dataTypes.ENUM('OAuth1'),
...
integration.hasOne(integrationDataOAuth1);
...
```
the model for `IntegrationDataOAuth1`:
```
integrationDataOAuth1.belongsTo(integration, { scope: { authSchemeName: 'OAuth1'} });
```

If I'm getting this right, the Integration can have at most one IntegrationDataOAuth1, and the IntegrationDataOAuth1 can only be related to Integration models when they have the appropriate authSchemeName property. I read the intersection of those conditions as enforcement of 1:1 between `Integration` and `IntegrationData*`. Does that check out?

howebenhowe: quoting off-band communication with Eric to explain subsequent commits abandoning polymorphic approach:

>3mcd: AFAIK this doesn’t enforce data integrity at the database level but at the ORM level. I think the association scope `authSchemeName: ‘OAuth1’` boils down to a `WHERE` clause when Sequelize builds underlying queries to fetch associated models. It doesn’t enforce any kind of DB-level constraint (like a foreign key). I’m not sure if Sequelize would even consider the association scope if we accidentally added more than one `IntegrationData*` to an `Integration` instance/row. Your solution is definitely elegant and easy to understand, just thinking about the data integrity. If we’re cool with that, I’m okay with approving this as-is.

3mcd (approved):

PR #2559, opened 2023-03-15
citations real time

### LINKED ISSUE #2408: picker: live search results from zotero

howebenhowe:
## Issue(s) Resolved
#2408 

### To Authenticate

0. log in
1. create you citations-testing pub
2. create a citation
3. in the citation editor, if you haven't yet authenticated, you'll see this button: 
Screenshot from 2023-03-23 16-11-45
4. click 'Connect Zotero', which will take you to your privacy settings page
5. click 'Activate' in the zotero integration section
6. complete the zotero-side oauth flow using login credentials available from 1Password until you are redirected to a 404 pubpub.org url (something along the lines of `https://www.pubpub.org/auth/zotero/redirect?oauth_token=c789475127776014731f&oauth_verifier=fae9f767121206365c3b`)
7. edit your url, replacing the part up 'til `pubpub.org` with the review app url (`https://pubpub-pipel-bh-citatio-4arxsu.herokuapp.com/`) and then hit enter
8. you should be returned to the privacy settings tab and the zotero integration section should offer you the choice to remove the integration:
Screenshot from 2023-03-23 16-17-32

### Once Authenticated

0. make sure zotero integration is active
10. open your citation pub
11. create/click on your citation to open the editor
13. type a search query into zotero search prompt box
14. after 300ms debounce, request is fired + loading state
15. one of the suggested results, should produce formatted structured citation below
16. click 'update citation' -> voila

Without integration/log-in (above)
With integration
Screenshot from 2023-03-23 16-20-13

After search
Screenshot from 2023-03-23 16-20-28

### Supporting Docs
[spec](https://notes.knowledgefutures.org/pub/ck6f4zsx#alternative-live-search-results)


server/zoteroIntegration/api.ts
```diff
@@ -0,0 +1,12 @@
+import app, { wrap } from 'server/server';
+
+import { getZoteroIntegration } from './queries';
+
+app.get(
+	'/api/zoteroIntegration',
+	wrap((req, res) =>
+		getZoteroIntegration(req.user.id).then((integration) =>
+			res.status(201).json({ id: integration.id }),
```
3mcd, line 9: We may want to return a 404 if `integration` is `null`.

client/components/FormattingBar/controlComponents/ControlsFootnoteCitation/ControlsFootnoteCitation.tsx
```diff
@@ -63,20 +64,26 @@ const ControlsFootnoteCitation = (props: Props) => {
...
 	const [renderedStructuredValue, setRenderedStructuredValue] = useState(
 		noteManager.getRenderedValueSync(structuredValue),
 	);
-	const [debouncedValue] = useDebounce(structuredValue, 250);
 	const html = renderedStructuredValue?.html;
 	const showPreview = html || unstructuredValue;
+	const [debouncedValue] = useDebounce(structuredValue, 250);
+
+	const onSelectSuggestedCitation = (selected) => {
+		if (selected.bibtex) {
```
3mcd, line 83: We should create a type alias for this object that we can also use on the API side after fetching the results from Zotero. Is it something like `ZoteroCSLJSON & { bibtex: string }`?

server/citation/api.ts
```diff
@@ -24,9 +27,13 @@ app.get('/api/citations/zotero', (req, res) => {
 			}
 			const zoteroId = parseInt(zoteroIntegration.zoteroUserId, 10);
 			const myApi = zoteroClient(integrationDataOAuth1.accessToken).library('user', zoteroId);
-			return myApi.items().get({ q });
+			return myApi.items().get({ q, include });
```
3mcd, line 30: Something like `zoteroApi` would be more descriptive here.

3mcd (changes requested):
Looks great. Left a few suggestions. Also getting a white screen of death issue when pressing 'Enter' in the citation picker input:
Screenshot 2023-03-21 at 1 09 27 PM

client/components/FormattingBar/controlComponents/ControlsFootnoteCitation/CitationBuilder.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+	useEffect(() => {
+		apiFetch('/api/zoteroIntegration').then((res) => {
+			setHasZoteroIntegration(res.id);
+		});
+	}, []);
+
+	useEffect(
+		() => {
+			if (hasZoteroIntegration) {
```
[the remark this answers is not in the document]
howebenhowe: Definitely can choose either-button-or-suggest; I thought to include a `<Suggest disabled={true} .../>` component in both cases so the non-integrated user would have a more complete picture of what they are missing out on.

howebenhowe:
some thoughts from my own testing:
+ after switching from using citation-js on our end to requesting bibtex-formatted results directly from zotero, the turn-around time seems to have gone up significantly
+ though debounced, you can still get a second query to fire before the first comes back, triggering a race condition for the results displayed, sometimes resulting in incorrect

3mcd (approved):
LGTM 👍

zverhope:
To repeat myself, just awesome work, Ben - authorization and resulting connection are pretty seamless. No issues with rate limiting, timeouts, or any other critical errors, at least that I could generate. I have some additional feedback below. I think you're already addressing the point about returning the n-1 search results for n characters (returning `mark` when searching for `marks`) by what you mentioned re: queuing search queries. Aside from that, the only points that I think it might be important to address before a release are the first one (adding rather than overwriting) and the last two - changing parsedDate to the year alone and adding the title, preferably the short title when specified by the user's Zotero record. Of course, I'd love to talk through any and all of these. Hopefully one or two of them are helpful.

- choosing a new item from the Zotero library overwrites the existing content of the structured data input field, making it difficult to cite more than one item. This would be particularly detrimental to lit reviews and introductory discussions, which often construe a particular line of thought regarding a topic, requiring the writer to cite multiple instances of the tendencies they're identifying
	- expected behavior: selecting a second item should add the item (prepend?) to the existing content of the structured data input field
	- ideally (meaning in a future iteration), the search bar would recognize and highlight the references already present in the structured data input field. Clicking them again would deselect the reference and remove it from the field.

- because we want to allow for multiple items, also enabling the user to refine the particular citation by modifying the data--e.g., adding page numbers--it would be best to declutter the BibTeX that we're pulling into the structured data input field
	- as you'll see if you look at most of the citations in our test library, the bulk of the space is taken up by three fields: abstract, keywords, and note. None of this is necessary for any of the citation styles we support. Removing those from the record would make the structured data more readable and easier to modify.

- would it be better to wait to initiate a call to Zotero until a certain number of characters--probably 3--have been typed? I can't imagine a useful search of only 1 or 2 characters, apart perhaps (although unlikely) from tags. And we want to introduce a specific filter for tags in a future iteration anyway.  (Typing # before a string, hitting spacebar, and then having the subsequent string search only the set of references contained within that tag.)

- related to the last point, would it reduce the wait time for search results if we specified a limit for the number of results (10–20) in our call to Zotero? There's a good chance you've already imposed a limit, but I did notice a longer hang on early (and generally less useful) searches that were returning more (but less uniformly relevant) results. I don't find the wait very long at all, so this isn't something I think needs fixing - I'm more just curious about what the benefits would be if we limited results.

- I've noticed that if I type n characters into the search field, I'll occasionally return search results for the n-1 search. (I mean that if I type `marks` hoping to find `markson`, I sometimes still get items involving people named "Mark" in my search results.)

- `creatorSummary` is perfect for a short but recognizable attribution. I don't like `parsedDate` as much, however, as it returns date strings of variable length - most are `YYYY`, many are `YYYY-MM`, and a few are YYYY-MM-DD. All anyone will care about is the year, so providing more than just the year for some items creates visual asymmetry and, more importantly, a bit of extra clutter. I'd strongly recommend just pulling the year, then, which corresponds perfectly to the author-date shorthand that researchers typically use between themselves to refer to specific papers.

- related to this: we need to include titles in search results. I often don't know the date that a certain article was published, so there's no way for me to distinguish between works by the same author(s) in those instances. And it's more difficult to do without titles on occasions when an author publishes more than one thing in a year.
	- in order to economize space within the search result string, I'd first check to see if an item includes a short title - I believe it's `Shorttitle` in bibtex and typically `shortTitle` in CSL-JSON . (Short titles are most commonly used specified for science papers, which can have ridiculously long titles.) This field will likely be empty for most items, so then you'd pull the full title, perhaps truncating it to n characters, as a user should be able to recognize their citation with only the first five or six words as a cue
	- the typical format for this (recognizable to anysone who uses the built-in picker in their reference manager) would look exactly like the author-date you're using now, followed by a spaced dash and the title -> `creatorSummary (year) - title`

PR #2608, opened 2023-04-13
store redirect host on user model in oauth flow

### LINKED ISSUE #2598: zotero oauth flow redirect doesn't return to community url
this is a fast-follow for #2581

1. when proceeding through oauth flow to link/integrate a user account with zotero, the flow takes you away from user privacy settings page to zotero login
2. after completing zotero login, user is redirected to pubpub.org/legal/settings rather than [community-subdomain].pubpub.org/legal/settings
3. that's a 404
...

get back to user privacy settings _within_ their community

howebenhowe:
## Issue(s) Resolved

#2598

## Test Plan
Perform oauth flow from 2 consecutive communities
1. log in as community-picker-enabled user (e.g. team@pubpub.org)
2. navigate to user privacy settings
3. if zotero integration is currently active, remove it
4. click 'Activate' for zotero integration
5. proceed through zotero oauth login + screens (credentials available in onepass)
6. after login, you should be returned to user privacy settings page within same community
7. pick a new community
8. return to user privacy settings page
9. zotero integration _should_ be active
10. remove zotero integration
11. perform oauth flow again
12. you should be returned from zotero back to most recently visited community

### DIFF
--- a/server/routes/authenticate.tsx
+++ b/server/routes/authenticate.tsx
@@ -1,6 +1,7 @@
 import passport from 'passport';
 import app from 'server/server';
 import { User } from 'server/models';
+import { isDevelopment } from 'utils/environment';
 
 app.get(
 	'/auth/zotero',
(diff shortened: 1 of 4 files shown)


server/routes/authenticate.tsx
```diff
@@ -1,14 +1,28 @@
...
+import { User } from 'server/models';
+import { isDevelopment } from 'utils/environment';
 
-app.get('/auth/zotero', passport.authenticate('zotero'));
+app.get(
+	'/auth/zotero',
+	(req, res, next) => {
+		if (req.user) {
+			const authRedirectHost = req.get('host');
+			User.update({ authRedirectHost }, { where: { id: req.user.id } });
```
3mcd, line 11: Since `User.update` returns a promise, we should handle it like

```ts
if (req.user) {
  User.update({ authRedirectHost }, { where: { id: req.user.id } }).then(() => {
    next();
  }).catch(error => {
    next(error);
  });
}
```

Or, with async/await:

```ts
async (req, res, next) => {
  if (req.user) {
    try {
      await User.update(...);
    } catch (error) {
      next(error);
    }
  } else {
    next();
  }
}
```

Which is what our `wrap` utility does!

3mcd (approved):
I'm going to go ahead and approve this, but we should add a migration that adds the new `authRedirectHost` column to the Users table.

PR #2613, opened 2023-04-18
More DataCite fixes

3mcd:
## Issue(s) Resolved

N/A

## Test Plan

Using the `eric` community...

1. Verify that you can set a DOI for a Pub prior to initial release.
2. Verify that custom DOI suffixes actually show up in Fabrica.
3. Verify that the _Publisher_ community setting is used as the value for `publisher` in a deposit.


gabestein (approved):
LGTM! Thanks for the quick turnaround.

PR #2615, opened 2023-04-19
Controls Link Enhancement

### LINKED ISSUE #2498: Create New Link Modal Functionality
# Overview

Using the new design/dev process we implement the updated link modal. 

Image

## TODO

This will need to:

- Toggle `target=_blank` on the enclosing link tag property.
     - this means creating a way to edit the range for said link in the prosemirror block. 
- Add the ability to create connections to the current Pub
     - potentially update pubs to show connections immediately when added
- Update the modal component with the current design and styling

qweliant:
## Issue(s) Resolved

#2498

## Test Plan

Run dev server 

For the next part:
Use a link to a pub ( from demo.duqduq )
Use a link to a DOI or an actual DOI  ( https://arxiv.org/pdf/2304.00353.pdf, https://doi.org/10.48550/arXiv.2304.00353 )
Use a link to some random site (https://blueprintjs.com/)

*use each 

create a link in a pub document
set a connection ( use different relationships and/ or directions for each link )
save a connection

be sure the attributions and metadata for link is correct
be sure a new collection edge appears in the pub 
be sure the collection has the correct relationship and direction

reload the page, be sure the connection is present with "create connection still selected"
update the direction and/or relationship
be sure the changes persist

remove a connection from the link modal
be sure it is removed from the pub document and dashboard

remove a connection from the dashboard
be sure it is removed in the pub document 
be sure 'Create Connection' is not checked

open your pub document set open in new tab for link ratios  2:1 open | 1:2 open or whatever just make sure there's a disjoint between things opening in a new tab and not

publish the pub
navigate to release
be sure links that are checked open in new tab open in tab and ones that dont, do not

image
image
image
[Screencast from 04-25-2023 10:43:11 AM.webm](https://user-images.githubusercontent.com/34730449/234313875-541f7bc8-5d93-4827-a279-f07cd265307d.webm)

### Notes/Context/Gotchas

edges with a direction away from the pub do not update in real time when adding

### Supporting Docs
https://notes.knowledgefutures.org/pub/kz8y3bt6/release/2


client/components/FormattingBar/controlComponents/ControlsLink.tsx
```diff
@@ -61,6 +107,198 @@ const ControlsLink = (props: Props) => {
 		}
 	};
 
+	const checkedOpenInNewTab = activeLink.attrs.target === '_blank';
+
+	const handleLinkAttr = () => {
+		setIsLoadingAttr(true);
```
3mcd, line 113: Was this necessary to get the attribute to update in the DOM? If not, we might be able to remove this state.

3mcd (changes requested):
Code looks great, but I think we need to spend a little more time with it because I found a few issues when testing.

When creating a connection to an external publication like so:
image

I get the following error:

image

Connection creation did work for a Pub URL, but when I clicked "Save connection", the new connection didn't show up in the Pub. The connection _had_ been created in the dashboard, but when I reloaded the page, the link did not have a connection tied to it.

After I deleted the link (and connections) to get a clean slate, I tried to add a connection to the same Pub through a link. The previous issue happened again, so I clicked "Save connection" another time. The connection still did not show up at the bottom of the Pub, and the connections looked like this in the dashboard:
image

client/components/FormattingBar/controlComponents/ControlsLink.tsx
```diff
@@ -71,9 +346,35 @@ const ControlsLink = (props: Props) => {
...
+					title="Options"
+					icon="chevron-down"
+					onClick={() => setIsOpen(!isOpen)}
+				/>
+				<Button
+					small
+					minimal
+					title="Remove"
+					icon="disable"
+					onClick={activeLink.removeLink}
```
qweliant, line 362: The onclick for this should remove the connection as well. May not be important just yet but a quick add on to this should be alright

3mcd (approved):
LGTM!

PR #2629, opened 2023-04-24
feat: use explicit status field

3mcd:


client/components/FormattingBar/controlComponents/ControlsLink.tsx
```diff
@@ -39,6 +39,19 @@ type Props = {
...
+enum Status {
+	LoadingEdge,
+	EditingLink,
+	EditingLinkExtra,
+	FetchingEdgeProposal,
+	EditingEdge,
+	UpdatingEdge,
+}
+
+function isStatus(status: Status, ...statuses: Status[]) {
```
qweliant, line 51: Why are are `statuses` spread into `isStatus`

ex: 
```
{isStatus(
    status,
    Status.FetchingEdgeProposal,
    Status.EditingEdge,
    Status.UpdatingEdge,
    )
}
```

Is this function using the second argument to say the input will be a list of destructured values

And whether or not some status has been set

qweliant (approved):
passing `isStatus` directly makes things harder to read, but it is pretty awesome loading state

client/components/FormattingBar/controlComponents/ControlsLink.tsx
3mcd: It's just a little helper that turns code like
```ts
if (status === Status.X || status === Status.Y || status === Status.Z) {}
```
into
```ts
if (isStatus(status, Status.X, Status.Y, Status.Z)) {}
```

PR #2636, opened 2023-04-25
feat: drop cookie on login/logout for caching

### LINKED ISSUE #2617: Add logged in cookie for caching
### What?
Add a cookie that indicates when a user is logged in, and clears it when it doesn't.

### Why?
Better support for widespread caching strategies.

Current cookie is session ID that's inscrutable and does not clear when logged out.

gabestein:
Adds a cookie for caching when logged in, and changes it when logged out. Went this route instead of a login cookie because it allows us to override it in the future if needed for caching purposes, rather than tying exclusively to login. Decided not to clear it on logout because it allows us to target it specifically in caching controls if needed (though the default will be to cache on pages, collections, and user profiles unless it sees pp-no-cache set).

## Issue(s) Resolved
Closes #2617 

## Test Plan
1. Visit test deployment
2. Logout and clear cookies
3. Login
4. Verify that `pp-cache` is set to `pp-no-cache`
5. Logout
6. Verify that `pp-cache` is set to `pp-cache`
7. Clear cookies
8. Create a new test account
9. Login and verify cookie was set on first login.

### Notes/Context/Gotchas
I was unable to modify the existing cookie because it's simply a session token set by passport.

### DIFF
--- a/server/login/api.ts
+++ b/server/login/api.ts
@@ -113,6 +113,7 @@ app.post('/api/login', (req, res, next) => {
 				if (err) {
 					throw new Error(err);
 				}
+				res.cookie('pp-cache', 'pp-no-cache');
 				return res.status(201).json('success');
 			});
 		})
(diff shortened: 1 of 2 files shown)


3mcd (changes requested):
Tested in the review app. Logging in does not seem to set the `pp-cache` to any value. Logging out correctly sets `pp-cache` to `pp-no-cache`.

gabestein:
Sigh, sorry, that's because somehow I managed moved it to the wrong block in the login flow. Should work once it's redeployed.

3mcd (approved):
LGTM!

PR #2637, opened 2023-04-26
Safari CSS Webkit Issue

### LINKED ISSUE #2634: Fix removal suggestion styles not appearing on safari
What?
On safari, our `text-decoration: line-through` styling for removal suggestions doesn't work. We should use the right property for safari and make sure the styles show up on all the browsers we target. 

Initial estimate
1

qweliant:
## Issue(s) Resolved
#2634 

## Test Plan
Visit Safari browser and another browser and view http://localhost:9876/pub/pegl1ofe/draft
be sure the text looks like below in both browsers

image
image

### Notes/Context/Gotchas
the neater one line webkit solution did not work so i split them up

### Supporting Docs
https://stackoverflow.com/questions/51181010/text-decoration-line-through-css-does-not-work-on-safari-with-my-completed-cla

### DIFF
--- a/client/components/Editor/styles/suggestedEdits.scss
+++ b/client/components/Editor/styles/suggestedEdits.scss
@@ -44,7 +44,6 @@ span[data-suggestion-mark='true'][data-suggestion-kind='addition'] {
 
 span[data-suggestion-mark='true'][data-suggestion-kind='removal'] {
 	text-decoration-thickness: 2px;
-	-webkit-text-decoration-thickness: 2px;
 	text-decoration-color: $removal;
 	-webkit-text-decoration-color: $removal;
 	text-decoration-line: line-through;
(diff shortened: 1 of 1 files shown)


qweliant:
Should i go through and test any other potential css bugs from the text-decorator?

client/components/Editor/styles/suggestedEdits.scss
```diff
@@ -43,7 +43,12 @@ span[data-suggestion-mark='true'][data-suggestion-kind='addition'] {
 }
 
 span[data-suggestion-mark='true'][data-suggestion-kind='removal'] {
-	text-decoration: 2px $removal line-through;
+	text-decoration: 2px;
+	-webkit-text-decoration: 2px;
```
kalilsn, line 47: safari doesn't seem to recognize this property, so the line is 1px. however, it looks like there's wide support for `text-decoration-thickness`, including unprefixed on safari

kalilsn (changes requested):
i wouldn't spend time looking for css bugs right now, but if you noticed anything else broken then let me know and open an issue!

client/components/Editor/styles/suggestedEdits.scss
qweliant: oof, thought i copied from the stash on master

kalilsn (approved):
Looks great!

PR #2638, opened 2023-04-26
Add loop video option

### LINKED ISSUE #2618: Add looping option for videos in editor
### What?
Add option to loop natively uploaded videos in editor.

### Why?
No good way to add looped videos right now. Some folks add GIFs, but that means they are grouped with "Images" if they use automated captions, which is not ideal.

Can simply toggle HTML 5 [video loop attribute](https://www.w3schools.com/tags/att_video_loop.asp).

kalilsn:
## Issue(s) Resolved

Resolves #2618

## Test Plan

- Open a pub draft
- Add a video (or choose an existing native video node)
- Turn loop on and off, test that the video does or does not repeat

Loop option checked:
image

Unchecked:
image


qweliant (approved):

gabestein (approved):
LGTM!

PR #2640, opened 2023-04-26
move SE buttons to right

### LINKED ISSUE #2572: Move suggestion buttons to right side of toolbar
### What?
Suggested edit controls should be moved to the less crowded right hand side of the editor

### Why?
- We're copying google docs
- These controls are quite different from other editor functions
- We'll need some space for their tooltips etc.

### Initial estimate
1

qweliant:
## Issue(s) Resolved
#2572 

## Test Plan
View a pub community with suggested edits on. Make sure you can use the buttons and 

@djagdish feel free to comment if we should arrange it differently
image

### DIFF
--- a/client/containers/Pub/PubDocument/PubHeaderFormatting.tsx
+++ b/client/containers/Pub/PubDocument/PubHeaderFormatting.tsx
@@ -19,7 +19,7 @@ type Props = {
 
 const PubHeaderFormatting = (props: Props) => {
 	const { disabled, editorWrapperRef } = props;
-	const { scopeData } = usePageContext();
+	const { scopeData, featureFlags } = usePageContext();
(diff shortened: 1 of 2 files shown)


client/containers/Pub/PubDocument/PubHeaderFormatting.tsx
```diff
@@ -51,6 +51,15 @@ const PubHeaderFormatting = (props: Props) => {
 				}}
 			/>
 			<div className="right-content">
+				<FormattingBar
```
kalilsn, line 54: This does render a few divs with no content, but some aria tags in the editor. Can you wrap this in the suggested edits flag so it doesn't render anything when the flag is off?

kalilsn (approved):
Looks good! Tested both the suggested edits community and demo on the review instance

PR #2642, opened 2023-04-27
Run link input rule on enter

### LINKED ISSUE #2518: Link on typing should work after newline
Auto-create links after newlines

The new functionality in #2482 works when you type a link and then hit space, but it should also work on newlines, and doesn't. This doesn't seem possible purely with regex, perhaps because newlines create new PM blocks rather than actual newline characters? In either case, we should figure out an approach to support this.

kalilsn:
## Issue(s) Resolved
Resolves #2518 

## Test Plan

On the review instance or a local branch, navigate to a pub editor and test:
	- typing a url or email address and pressing space
	- typing a url or email address and pressing enter

Also be sure to test these with and without content preceding and following (like `example.comandarunonsentence`, then press enter after `.com`, or `example.com<inline math node>`)

### DIFF
--- a/client/components/Editor/plugins/inputRules.ts
+++ b/client/components/Editor/plugins/inputRules.ts
@@ -1,5 +1,5 @@
 import { Fragment, MarkType, NodeType } from 'prosemirror-model';
-import { EditorState } from 'prosemirror-state';
+import { EditorState, Transaction } from 'prosemirror-state';
 import {
 	inputRules,
 	wrappingInputRule,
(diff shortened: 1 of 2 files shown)


client/components/Editor/plugins/inputRules.ts
```diff
@@ -104,31 +104,54 @@ const inlineMathRule = (
 // textblock starting with two dollar signs into a math block.
 const blockMathRule = (nodeType) => makeBlockMathInputRule(REGEX_BLOCK_MATH_DOLLARS, nodeType);
 
-const HTTP_MAILTO_REGEX = new RegExp(
-	// eslint-disable-next-line no-useless-escape
-	/(?:(?:(https|http|ftp)+):\/\/)?(?:\S+(?::\S*)?(@))?(?:(?:([a-z0-9][a-z0-9\-]*)?[a-z0-9]+)(?:\.(?:[a-z0-9\-])*[a-z0-9]+)*(?:\.(?:[a-z]{2,})(:\d{1,5})?))(?:\/[^\s]*)?\s$/,
-);
+const emailOrUriRegexBase =
+	'(?<emailOrUri>(?:(?:(https|http|ftp)+)://)?(?:\\S+(?::\\S*)?(?<atSign>@))?(?:(?:([a-z0-9][a-z0-9-]*)?[a-z0-9]+)(?:\\.(?:[a-z0-9-])*[a-z0-9]+)*(?:\\.(?:[a-z]{2,})(:\\d{1,5})?))(?:/[^\\s]*)?)';
```
kalilsn, line 108: I dropped the `\` in `[a-z0-9\-]` which is what eslint was complaining about. I checked myself that `[a-z0-9-]` unambiguously matches the `-` character.

kalilsn: The other changes to the regex here (since it's hard to look at) are that I pulled the `\s` at the end out since I only want it some of the time, and I added a few named capture groups to make the code below more readable.

client/components/Editor/plugins/keymap.ts
```diff
@@ -150,10 +152,46 @@ export default (schema) => {
...
+
+		if (dispatch) {
+			// Call our custom split block command, then add the link marks as part of the same
+			// transaction. This should be safe to do without mapping the start and end positions
+			// because the changes caused by splitBlock will all be after the cursor
+			splitBlockPreservingAttrs(['textAlign', 'rtl'])(state, (tr) => {
+				const addLinkMark = linkRuleHandler(schema.marks.link, false, tr);
+				const start = $cursor.pos - match[0].length;
+				const end = $cursor.pos;
+				dispatch(addLinkMark(state, match, start, end));
```
kalilsn, line 184: This is all necessary because we're running this code as one of the possible enter key handlers, and we still want the usual behavior inside a text block of creating a new paragraph when you press enter.

When I first started, I misunderstood how `chainCommands` below worked, and assumed we could add our command at the beginning, and then let the rest of them run as usual. However, the next command isn't invoked if any of the previous ones returns true, and users are explicitly cautioned not to create side effects inside of a command that returns false (https://discuss.prosemirror.net/t/chaincommands-seems-to-be-working-wrong/1321/2). So the only way to retain the normal enter handler behavior with this new command is to combine the link changes into the same transaction as whatever other command we run. Luckily in this case (inside a text block, $cursor selection), I'm pretty sure only `splitBlockPreservingAttrs(['textAlign', 'rtl'])` could possibly run.

client/components/Editor/plugins/inputRules.ts
```diff
@@ -104,31 +104,54 @@ const inlineMathRule = (
...
+	return (state: EditorState, match: RegExpMatchArray, start: number, end: number) => {
+		const resolvedStart = state.doc.resolve(start);
+		if (!resolvedStart.parent.type.allowsMarkType(markType)) {
+			return state.tr;
+		}
+		const emailOrUri = match.groups!.emailOrUri;
+
+		const linkAttrs = match.groups!.atSign
+			? { href: 'mailto:' + emailOrUri }
+			: { href: emailOrUri, target: '_blank' };
```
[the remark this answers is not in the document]
gabestein: Agreed, should not be target blank by default.

client/components/Editor/plugins/keymap.ts
```diff
@@ -150,14 +152,55 @@ export default (schema) => {
...
+
+		if (dispatch) {
+			// Call our custom split block command, then add the link marks as part of the same
+			// transaction. This should be safe to do without mapping the start and end positions
+			// because the changes caused by splitBlock will all be after the cursor
+			customBlockSplitter(state, (tr) => {
+				const addLinkMark = linkRuleHandler(schema.marks.link, false, tr);
+				const start = $cursor.pos - match[0].length;
+				const end = $cursor.pos;
+				dispatch(addLinkMark(state, match, start, end));
```
kalilsn, line 189: This is all necessary because we're running this code as one of the possible enter key handlers, and we still want the usual behavior inside a text block of creating a new paragraph when you press enter.

When I first started, I misunderstood how `chainCommands` below worked, and assumed we could add our command at the beginning, and then let the rest of them run as usual. However, the next command isn't invoked if any of the previous ones returns true, and users are explicitly cautioned not to create side effects inside of a command that returns false (https://discuss.prosemirror.net/t/chaincommands-seems-to-be-working-wrong/1321/2). So the only way to retain the normal enter handler behavior with this new command is to combine the link changes into the same transaction as whatever other command we run. Luckily in this case (inside a text block, $cursor selection), I'm pretty sure only `splitBlockPreservingAttrs(['textAlign', 'rtl'])` could possibly run.

client/components/Editor/plugins/inputRules.ts
```diff
@@ -104,31 +104,50 @@ const inlineMathRule = (
...
+) => {
+	return (state: EditorState, match: RegExpMatchArray, start: number, end: number) => {
+		const resolvedStart = state.doc.resolve(start);
+		const tr = transaction ?? state.tr;
+		if (!resolvedStart.parent.type.allowsMarkType(markType)) {
+			return tr;
+		}
+		const emailOrUri = match.groups!.emailOrUri;
+
+		const href = `${match.groups!.atSign ? 'mailto:' : ''}${emailOrUri}`;
```
qweliant, line 130: Adding a mailto link inserts mailto in the link twice
[Screencast from 05-01-2023 05:27:38 PM.webm](https://user-images.githubusercontent.com/34730449/235534695-ae898d45-a0f8-4dc6-8c71-fd84a4e7cd60.webm)

client/components/Editor/plugins/inputRules.ts
```diff
@@ -104,31 +104,50 @@ const inlineMathRule = (
...
+
+		const link = state.schema.text(emailOrUri, [state.schema.mark(markType, { href })]);
```
qweliant, line 132: Not a regression but a quick enhancement, if this link is not prepended with https:// beforehand it to be afterwards. otherwise the link will route to a page not found.

client/components/Editor/plugins/inputRules.ts
kalilsn: Since this is also not a regression and I'm trying to close stuff out this sprint, I'd like to just write up issues for both of these changes instead of putting them in this PR.

In this case, I'm also not really sure what the right behavior is. Should we be linkifying the entire string of `mailto:user@example.com`?

client/components/Editor/plugins/inputRules.ts
kalilsn: Same as above, I'd like to make an issue. Also, I've noticed this too. It's way too easy to create a broken link of the form `pubpub.org/pub/<pubid>/<absolute url>`. I wonder if our links should default to interpreting the URL as absolute rather than relative when there's no scheme or leading `[.]/`, but that feels like a big default change to make! Also I definitely want to make sure it's consistent across all the ways of creating a link, which feels outside the scope of this (already too-complicated) fix.

3mcd (approved):

PR #2644, opened 2023-04-27
Dev env improvements

kalilsn:
A few potential dev environment improvements I wanted to suggest:

- Setting up the prosemirror dev tools, which I have found very useful (and removing them from the prod bundle)
- Getting rid of the postinstall command, because it's slow and I almost never want to run it. Heroku admitted that forcing everyone to do this was a mistake and now they provide alternatives: heroku runs your `"build"` or `"heroku-postbuild"` scripts, so now you can change a package without having to do a full webpack build from scratch.
- Switching to concurrently to run the backend and webpack dev server at the same time, which seems a little more stable for me


package.json
```diff
@@ -24,7 +24,6 @@
 		"check": "tsc --noEmit",
 		"pretty": "prettier --write '**/*.{js,scss}'",
 		"install-git-hooks": "rm -f ./.git/hooks/* && cp ./.githooks/* ./.git/hooks && chmod +x ./.git/hooks/*",
-		"postinstall": "npm run build-prod",
```
kalilsn, line 27: This might be necessary for heroku somewhere, but if so I'd like to do it without the postinstall hook (or at least make it conditional on some heroku only env var)

kalilsn: Not necessary! https://help.heroku.com/P5IMU3MP/heroku-node-js-build-script-change-faq

PR #2666, opened 2023-05-09
Enhance/prevent release

### LINKED ISSUE #2589: Prevent creating releases from drafts with unresolved suggestions
### What?
If a user attempts to create a release from a pub with unresolved suggestions, show them a dialog saying they must resolve them first. Include buttons to accept all and reject all suggestions. Depends on #2571

Also ensure no suggestion attributes are present in releases, since they currently appear to be rendered even when there's no value.

### Why?
We don't want to render suggestions for pub readers and users may unintentionally attempt a release before resolving their suggestions.

### Initial estimate
3 days

qweliant:
## Issue(s) Resolved
#2589 

## Test Plan
1. Head to a pub 
2. Create some text and suggest edits
3. Try to release the pub
4. Be sure release button is hidden and the ignore / accept buttons are not
5. perform an action on suggested edits(accept or ignore)
6. Be sure the release button now appears
7. Try 2-6 with an opposite action

image

### Notes/Context/Gotchas
this should only effect the pub release dialog component

### DIFF
--- a/client/components/PubReleaseDialog/PubReleaseDialog.tsx
+++ b/client/components/PubReleaseDialog/PubReleaseDialog.tsx
@@ -47,7 +47,7 @@ const createRelease = ({ historyKey, pubId, communityId, noteContent, noteText }
 
 const PubReleaseDialog = (props: Props) => {
 	const { isOpen, onClose, historyKey, pub, onCreateRelease } = props;
-	const { communityData } = usePageContext();
+	const { communityData, featureFlags } = usePageContext();
 	const { collabData } = usePubContext();
 	const [noteData, setNoteData] = useState<{ content?: {}; text?: string }>({});
 	const [isCreatingRelease, setIsCreatingRelease] = useState(false);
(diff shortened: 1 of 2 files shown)


client/components/PubReleaseDialog/PubReleaseDialog.tsx
```diff
@@ -236,10 +300,14 @@ const PubReleaseDialog = (props: Props) => {
 				{renderReleaseResult()}
 			</div>
 			<div className={Classes.DIALOG_FOOTER}>
-				<div className={Classes.DIALOG_FOOTER_ACTIONS}>
-					{createdRelease && renderPostReleaseButtons()}
-					{!createdRelease && renderPreReleaseButtons()}
-				</div>
+				{hasSuggestions() ? (
```
kalilsn, line 303: this line should also check that the suggested edits feature flag is enabled for this community

kalilsn (changes requested):
overall works great, just needs to consider the feature flag!

kalilsn (approved):

PR #2668, opened 2023-05-10
Fix list behavior (mostly)

### LINKED ISSUE #2592: Fix suggested edits issues with lists
### What?
Highlighting the table which contains an empty list in demo pub `j0w2ho8g` causes the app to crash, because the document contains a suggestion with a bad position. This happens when the first element of a list item is deleted in suggestion mode, which seems to add a suggestion outside the boundaries of the deleted content.

Additionally, if you create a list with two items and delete the second one, the cursor doesn't move and nothing appears to happen. In this case the suggestion appears to be added correctly but we have no styling targeting the number headings in an ordered list. There's also a difference in how the editor behaves here: in suggested edits mode deleting the first element of the list creates a new paragraph node and moves the cursor to it, whereas deleting the last item of the list does not.

### Initial estimate
2

kalilsn:
Fixes a number of bad interactions between suggested edits mode and lists by
- Changing the algorithm we use to restore deleted content:
	- Previously it worked from the end of the document to the beginning, restoring any content removed by the user, then marking it as a suggested removal, while assuming that earlier changed positions in the document would be unaffected by these restorations
	- Now it works in the opposite order, explicitly mapping the changed positions each time the document is modified
	- This properly handles complex steps that result in document lifts/merges/joins etc., like deleting the last item in a list, or pressing enter inside of an empty list item (which removes the item and creates a new paragraph outside the list).
- Updating prosemirror-schema-list (take a look at all the user-facing bugs between 1.1.2 and 1.2.2: https://github.com/ProseMirror/prosemirror-schema-list/blob/master/CHANGELOG.md#112-2019-11-20)
- Replacing our list markers with nearly equivalent pseudo `li:before` elements
	- Without this, we can't render the leaf node suggestion elements (the green/red line), which made suggestions pretty confusing when working with list items
- Also fixes the source of some crashes in the suggestion resolution algorithm by returning `false` in our `nodesBetween` callback after resolving a suggestion in a way that deletes a node from the document. This prevents `nodesBetween` from recursing into that now deleted node.

## Issue(s) Resolved

Resolves #2592 

## Test Plan
- Look at some existing lists, ordered and unordered (and rtl too ideally), and make sure the seem okay across browsers and operating systems (so far I've only tested on a recent macos + safari/chrome/firefox). There's a real risk of this behaving differently on different OSes because these characters are most likely coming from a system font.
- Try creating and modifying these lists in and out of suggested edits mode, and resolving suggestions too. Report any crashes or just "weird" behavior

### DIFF
--- a/client/components/Editor/plugins/suggestedEdits/appendTransaction/textAndStructure.ts
+++ b/client/components/Editor/plugins/suggestedEdits/appendTransaction/textAndStructure.ts
@@ -107,23 +107,26 @@ export const indicateTextAndStructureChanges = (context: SuggestedEditsTransacti
 		// was created by Suggested Edits and is therefore safe to outright delete.
 		const netRemovalSlice = getNetSlice(removedSlice, 'removal');
 		if (netRemovalSlice.size > 0) {
+			const numSteps = newTransaction.steps.length;
 			// Add the net removal slice back in...
 			newTransaction.replace(newStart, newStart, netRemovalSlice);
 
-			// Map the positions, again being careful to choose position before any inserted content
-			newStart = newTransaction.mapping.map(change.fromB, -1);
-			const removalEnd = newStart + netRemovalSlice.size;
+			if (numSteps !== newTransaction.steps.length) {
+				// Map the positions, again being careful to choose position before any inserted content
+				newStart = newTransaction.mapping.map(change.fromB, -1);
+				const removalEnd = newStart + netRemovalSlice.size;
 
-			// Now mark it as a suggested removal...
-			if (newStart < removalEnd) {
-				addSuggestionToRange('removal', context, newStart, removalEnd);
-			}
-			// And do the same things to its constituent nodes, as we did above for additions.
-			newTransaction.doc.nodesBetween(newStart, removalEnd, (node: Node, pos: number) => {
-				if (pos >= newStart && pos <= removalEnd && node.isBlock) {
-					addSuggestionToNode('removal', context, pos, node);
+				// Now mark it as a suggested removal...
+				if (newStart < removalEnd) {
+					addSuggestionToRange('removal', context, newStart, removalEnd);
 				}
-			});
+				// And do the same things to its constituent nodes, as we did above for additions.
+				newTransaction.doc.nodesBetween(newStart, removalEnd, (node: Node, pos: number) => {
+					if (pos >= newStart && pos <= removalEnd && node.isBlock) {
+						addSuggestionToNode('removal', context, pos, node);
+					}
+				});
+			}
 		}
 	});
 };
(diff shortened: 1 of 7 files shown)


client/components/Editor/styles/main.scss
```diff
@@ -67,3 +67,31 @@ sub {
...
+	position: relative;
+}
+
+li::before {
+	margin-left: -1em;
+	position: absolute;
+}
+
+ul > li::before {
+	content: '•';
```
kalilsn, line 88: This isn't really a solution, since it breaks our nested unordered lists
pub pub now:
image

pubpub after this change:
image

3mcd: Does something like this work (not sure if the characters are right)?
```scss
ul > li::before {
  content: "•";
}

ul > li > ul > li::before {
  content: "◦";
}

ul > li > ul > li > ul li::before {
  content: "■";
}
```

3mcd: I wrote a terrible sass function that would let us target list items of arbitrary depth.
```scss
@use 'sass:selector';

@function at-depth($child, $depth: 0, $parent: '*', $extend: false) {
	$result: '#{$parent} > #{$child}';
	@for $i from 0 through $depth {
		@if $i > 0 {
			$result: $result +
				selector.nest(
					'> #{$parent}',
					"#{if($i == $depth and $extend == true, '', '>')}#{$child}"
				);
		}
	}
	@return $result;
}

#{at-depth('li', 0, 'ul')}::marker { content: '•'; }
#{at-depth('li', 1, 'ul')}::marker { content: '◦'; }
#{at-depth('li', 2, 'ul', true)}::marker { content: '■'; }
```

client/components/Editor/plugins/suggestedEdits/appendTransaction/textAndStructure.ts
```diff
@@ -95,19 +107,26 @@ export const indicateTextAndStructureChanges = (context: SuggestedEditsTransacti
...
-			// Now mark it as a suggested removal...
-			if (newStart < removalEnd) {
-				addSuggestionToRange('removal', context, newStart, removalEnd);
-			}
-			// And do the same things to its constituent nodes, as we did above for additions.
-			newTransaction.doc.nodesBetween(newStart, removalEnd, (node: Node, pos: number) => {
-				if (pos >= newStart && pos <= removalEnd && node.isBlock) {
-					addSuggestionToNode('removal', context, pos, node);
+
+			if (numSteps !== newTransaction.steps.length) {
```
3mcd, line 114: Could we get a comment reminder why this is necessary?

3mcd (approved):

PR #2672, opened 2023-05-17
Inline suggestion buttons

### LINKED ISSUE #2590: Move accept/reject buttons inline
### What?
Move the accept/reject buttons into the doc, inline with the selected suggestion(s). 

_Should these show on hover, selection, or all the time?_

### Why?
- Users currently can't see which suggestion will be resolved when they click accept or reject in the toolbar. Moving the buttons inline with the suggestion range clarifies this interaction
- It's closer to what other suggestion mode interfaces look like
- The formatting bar is already crowded

### Initial estimate
4 days

qweliant:
## Issue(s) Resolved
#2590  

## Test Plan
Once again, i'm asking you to dona.. I mean make a pub

Copy some text in. 
Make edits to the text
Try accepting or rejecting the suggestions from the new inline menu
Try selecting a range in the doc with no suggestion edits

image

### DIFF
--- a/client/containers/Pub/PubDocument/PubInlineMenu.tsx
+++ b/client/containers/Pub/PubDocument/PubInlineMenu.tsx
@@ -7,7 +7,7 @@ import { usePageContext } from 'utils/hooks';
 import { Icon, ClickToCopyButton } from 'components';
 import { FormattingBar, buttons } from 'components/FormattingBar';
 import { setLocalHighlight, moveToEndOfSelection, isDescendantOf } from 'components/Editor';
-import { getSuggestionAttrsForNode } from 'components/Editor/plugins/suggestedEdits/operations';
+import { acceptSuggestedEdits } from 'components/Editor/plugins/suggestedEdits/resolve';
 
 import { usePubContext } from '../pubHooks';
 
@@ -25,19 +25,10 @@ const PubInlineMenu = () => {
 	const { canView, canCreateDiscussions } = scopeData.activePermissions;
 	const selection = collabData.editorChangeObject!.selection;
 	const shouldHide = useMemo(() => {
-		const selectionInSuggestionRange = (): boolean => {
-			if (!collabData.editorChangeObject || !collabData.editorChangeObject.view) return false;
-			const doc = collabData.editorChangeObject.view.state.doc;
-			let inSuggestionRange = false;
-			doc.nodesBetween(selection.$anchor.pos - 1, selection.$head.pos + 1, (node) => {
-				if (inSuggestionRange) return;
-				const present = getSuggestionAttrsForNode(node);
-				if (present) inSuggestionRange = true;
-			});
-			return inSuggestionRange;
-		};
-
-		const inRange = selectionInSuggestionRange();
+		if (!collabData.editorChangeObject || !collabData.editorChangeObject.view || !selection)
+			return true;
+		const state = collabData.editorChangeObject.view.state;
+		const inRange = acceptSuggestedEdits(state);
 
 		return (
 			inRange ||
--- a/client/containers/Pub/PubDocument/PubInlineMenu.tsx
+++ b/client/containers/Pub/PubDocument/PubInlineMenu.tsx
@@ -7,7 +7,7 @@ import { usePageContext } from 'utils/hooks';
 import { Icon, ClickToCopyButton } from 'components';
 import { FormattingBar, buttons } from 'components/FormattingBar';
 import { setLocalHighlight, moveToEndOfSelection, isDescendantOf } from 'components/Editor';
-import { acceptSuggestedEdits } from 'components/Editor/plugins/suggestedEdits/resolve';
+import { getSuggestionAttrsForNode } from 'components/Editor/plugins/suggestedEdits/operations';
 
 import { usePubContext } from '../pubHooks';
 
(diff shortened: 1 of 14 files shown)


client/containers/Pub/PubDocument/PubInlineSuggestedEdits.tsx
```diff
@@ -0,0 +1,88 @@
...
+			if (!collabData.editorChangeObject || !collabData.editorChangeObject.view) return false;
+			const doc = collabData.editorChangeObject.view.state.doc;
+			let inSuggestionRange = false;
+			// inclusive range is sel-1, sel+1
+			doc.nodesBetween(selection.$anchor.pos - 1, selection.$head.pos + 1, (node) => {
+				if (inSuggestionRange) return;
+				const present = getSuggestionAttrsForNode(node);
+				if (present) inSuggestionRange = true;
+			});
+			return inSuggestionRange;
```
kalilsn, line 34: I know I suggested this `nodesBetween` solution for looking through ranges initially (my bad!), but in this case I do think it's worth using the precomputed ranges in the suggested edits plugin state. And I think you should probably do that the same way the suggestion buttons determine if they should be enabled or not: https://github.com/pubpub/pubpub/blob/master/client/components/Editor/plugins/suggestedEdits/commands.ts#L35

If you call these functions without passing `dispatch`, the transaction never gets run, and they return false when the selection isn't inside a suggestion and true when it is.

client/containers/Pub/PubDocument/PubInlineMenu.tsx
```diff
@@ -24,7 +25,22 @@ const PubInlineMenu = () => {
...
+			let inSuggestionRange = false;
+			doc.nodesBetween(selection.$anchor.pos - 1, selection.$head.pos + 1, (node) => {
+				if (inSuggestionRange) return;
+				const present = getSuggestionAttrsForNode(node);
+				if (present) inSuggestionRange = true;
+			});
+			return inSuggestionRange;
+		};
+
+		const inRange = selectionInSuggestionRange();
```
kalilsn, line 40: I don't think you need the nodesBetween approach here either, and this could look identical to your function in the SE menu

client/containers/Pub/PubDocument/PubInlineSuggestedEdits.tsx
```diff
@@ -0,0 +1,68 @@
...
+	const shouldHide = useMemo(() => {
+		const selectionInSuggestionRange = (): boolean => {
+			if (!collabData.editorChangeObject || !collabData.editorChangeObject.view) return false;
+			const state = collabData.editorChangeObject.view.state;
+			return acceptSuggestedEdits(state);
+		};
+
+		const inRange = selectionInSuggestionRange();
+
+		return !inRange || !selection;
```
kalilsn, line 31:
```suggestion
		if (!collabData.editorChangeObject || !collabData.editorChangeObject.view) {
			return false;
		}
		const state = collabData.editorChangeObject.view.state;
		return !acceptSuggestedEdits(state); || !selection;
```

I don't think you need this function or the variables, since its only used here

qweliant: image


i get this error when applying that change, i tried doing this in the am and kept getting this err so maybe im doing something wrong?

kalilsn: oh yeah me too. it's because that `return false;` should be a `return true;`: we do want to hide this menu when those objects aren't present. relatedly, i think it's a mistake to assert `collabData.editorChangeObject!` before this block, so here's a rewritten version that works for me:

```typescript
	const selection = collabData.editorChangeObject?.selection;

	const shouldHide = useMemo(() => {
		if (!collabData.editorChangeObject || !collabData.editorChangeObject.view || !selection) {
			return true;
		}
		const state = collabData.editorChangeObject.view.state;

		return !acceptSuggestedEdits(state);
	}, [collabData.editorChangeObject, selection]);
```

client/containers/Pub/PubDocument/PubInlineSuggestedEdits.tsx
```diff
@@ -0,0 +1,68 @@
...
+	}, [collabData.editorChangeObject, selection]);
+
+	// range of editable editor space
+	const selectionBoundingBox: Record<string, any> =
+		collabData.editorChangeObject!.selectionBoundingBox || {};
```
kalilsn, line 36:
```suggestion
		collabData.editorChangeObject?.selectionBoundingBox || {};
```

client/containers/Pub/PubDocument/PubInlineSuggestedEdits.tsx
```diff
@@ -0,0 +1,68 @@
...
+			const state = collabData.editorChangeObject.view.state;
+			return acceptSuggestedEdits(state);
+		};
+
+		const inRange = selectionInSuggestionRange();
+
+		return !inRange || !selection;
+	}, [collabData.editorChangeObject, selection]);
+
+	// range of editable editor space
```
kalilsn, line 34: This isn't really true, right? This is the bounding box for the selection itself, not the whole editor.

client/containers/Pub/PubDocument/PubInlineSuggestedEdits.tsx
qweliant: ngl i was going crazy trying to figure out what was going on and stumbled into solving something you already did with acceptedEdits, i see that it really does always return true.  had to rebase again lol

kalilsn (approved):
🚀

Issue #2591, opened 2023-04-11
Show author of suggestions inline


kalilsn:
### What?
Display the avatar of the author of each suggestion inline (like the accept/reject buttons).

### Why?
- A common use case for suggestions is having editor(s) make suggestions which the author then resolves, in which case it's important to know who made the suggestion

### Initial estimate
3 days

kalilsn:
![Image](https://user-images.githubusercontent.com/473542/233148029-1f5844bb-2622-429c-90e3-cd9e6c21459a.png)



![Image](https://user-images.githubusercontent.com/473542/233148106-d11000b0-d12c-473a-928b-b9ccec69d2f4.png)

djagdish:
Proposed design option that shows:

-  author avatar on clicking the Suggested Edit area.
- on hover over the avatar, tooltip can show details about author name (and possibly edit time as well).
- slightly faded quick decision buttons on hover default. when hover over the avatar, or the buttons, their opacity comes 100%.
- Accept & Reject submission buttons have associated tooltips. 



![Image](https://user-images.githubusercontent.com/316939/234303816-25ad6e5e-87c5-456c-9a38-6d6513b72653.png)

djagdish:
Alternate option: 

- on clicking on the Suggested Edit area, show tooltip and buttons.

![Image](https://user-images.githubusercontent.com/316939/234307594-cd55b0b8-0a27-4e8e-af65-7d45e269fb52.png)


- on hover of button:

![Image](https://user-images.githubusercontent.com/316939/234307997-4bc3cb06-a3d3-4912-9278-6d8603630d50.png)

qweliant:
Leaving the tooltip open makes it stick to the range without updating its pos
![Kooha-2023-05-25-10-10-11](https://github.com/pubpub/pubpub/assets/34730449/7f7a18f3-f225-484f-8301-d297fee86439)




I think this is a happy medium
![Kooha-2023-05-25-10-05-40](https://github.com/pubpub/pubpub/assets/34730449/b2d7f322-3478-4c0f-bc70-c16694942583)

PR #2674, opened 2023-05-17
API method for getting user

qweliant:
## Issue(s) Resolved
Part #2591 

## Test Plan
run the test in user/__test__

be sure all pass

### Notes/Context/Gotchas
this will be an api call made on the client when getting user info for a suggested edit

### Supporting Docs
https://notes.knowledgefutures.org/pub/cthxjiio#who-made-the-suggestion-and-when


kalilsn (approved):
looks great!

PR #2675, opened 2023-05-17
Check for suggestions inside non-empty selections

kalilsn:
- This fixes the suggestion buttons always showing as enabled when selecting a range of the document, even if it had no suggestions. 
- It also reduces the likelihood of any bugs causing editor crashes by modifying how the formatting bar checks if the accept/reject suggestions buttons should be enabled. Now those calls will no longer actually try resolving suggestions (and discarding the transaction) and will just check that the selection intersects with a suggestion
- This will be useful for other work we're doing, both on our UI and to make math blocks function properly.

## Test Plan
Try a few different selections (cursor, range, node) both with and without suggestions inside. Ensure the buttons are enabled or disabled as appropriate.

### DIFF
--- a/client/components/Editor/plugins/suggestedEdits/resolve.ts
+++ b/client/components/Editor/plugins/suggestedEdits/resolve.ts
@@ -131,8 +131,8 @@ export const rejectSuggestions = (state: EditorState, from: number, to: number) 
 export const acceptSuggestedEdits = (state: EditorState, dispatch?: Dispatch): boolean => {
 	const range = getResolvableRangeForSelection(state);
 	if (range) {
-		const tr = acceptSuggestions(state, range.from, range.to);
 		if (dispatch) {
+			const tr = acceptSuggestions(state, range.from, range.to);
 			dispatch(tr);
 		}
 		return true;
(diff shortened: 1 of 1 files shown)


client/components/Editor/plugins/suggestedEdits/resolve.ts
```diff
@@ -16,16 +16,16 @@ const getResolvableRangeForSelection = (
...
-			return null;
+		const suggestionRange = suggestionRanges.find((range) => {
+			const selectionStartIntersectsSuggestion =
+				range.from <= selection.from && selection.from <= range.to;
+			const selectionEndIntersectsSuggestion =
+				range.from <= selection.to && selection.to <= range.to;
+			return selectionStartIntersectsSuggestion || selectionEndIntersectsSuggestion;
+		});
+		if (suggestionRange) {
+			return selection.empty ? suggestionRange : selection;
```
kalilsn, line 27: When an empty selection is passed (or a cursor) we can assume that the first suggestion found will be the only one, so we return that. But when a range of text is selected, we return the actual selection if any suggestions were found inside, so that we can still resolve _all_ the suggestions in that range.

qweliant: lol thats pretty cool

what doe these lines mean?
```
const selectionStartIntersectsSuggestion =
		range.from <= selection.from && selection.from <= range.to;
const selectionEndIntersectsSuggestion =
		range.from <= selection.to && selection.to <= range.to;
return selectionStartIntersectsSuggestion || selectionEndIntersectsSuggestion;
```

is this saying a range can be more than the selection? or i suppose how are `suggestionRanges` made

qweliant (approved):

client/components/Editor/plugins/suggestedEdits/resolve.ts
kalilsn: hm yeah i think these are named confusingly and i should revisit this logic

kalilsn: actually these are named right but just miss a pretty important case: where the range is entirely included inside the selection but doesn't intersect with the start or end

PR #2676, opened 2023-05-19
Show which suggestions are currently selected

### LINKED ISSUE #2595: Make sure suggestion styles are clear to colorblind users
### What?
Review styles for suggested edits to ensure they're accessible

### Why?
Right now we mark suggestions with a mixture of red, yellow, and green. Ian mentioned having difficulty with this as a colorblind person.

### LINKED ISSUE #2683: Add visual indicator of suggestions to be resolved
### What?
Add visual indicator to the suggestion ranges under the cursor

### Why?
We don't have any way for users to see which suggestion would be accepted or rejected when they click those buttons.

### Initial estimate
3

kalilsn:
This PR adds a new UI element to suggested edits: floating +/- icons. 

## Issue(s) Resolved
Fixes #2595 
Fixes #2683 

## Test Plan

Go to the review instance, play with some suggested edits content!

image
image

### Notes/Context/Gotchas

These icons have a few different purposes: 
- Indicating suggestions that add or remove nested <li> elements. Previously I made this work using a hack to replace the bullets in bulleted lists with ::before pseudo-elements, but that had to be reverted because it caused other issues.
- Making sure colorblind users can tell the difference between the different kind of suggestions
- Further differentiate suggestions from regular text modifications (like a strikethrough)
- Showing which suggestions will be resolved for the current cursor position (since the plugin will resolve all contiguous suggestions by the same author)

Some other notable changes here:
	- I've reverted back to a slightly earlier version of the inline suggestion info, which just puts the username inside of the button tooltips, rather than in a separate floating box below the cursor. This was a choice I made to reduce clutter and make sure this works on phones, where we have to be careful that our menus don't overlap with the OS inline menu (for copy/paste etc.)
	- the SE plugin now adds a `.selected` class around the cursor's position
	- contiguous suggestions from different users are no longer joined, so they'll only both be resolved if your cursor selection includes both of them


qweliant (approved):
I am not sure when modified edits should show. It only came up for me when I brokje out the table but not when i added any columns

gabestein (approved):
Look and feel looks ~good~ great!

Two non-blocking comments that I don't think need to be addressed now:

- Agree with @qweliant  that table behavior is a tad wonky. Each individual cell needs to be approved, which creates scenarios like a few faux options to delete a cell that can't actually be deleted (like the bottom-right corner of a 2x2 table). Ideally, you would be able to approve changes to whole rows/columns when initially created, and then later individual changes. I know this is also weird on gdocs, fwiw, so I don't think it's big at all. Just noting it.
- People are going to want to add discussions to ranges containing suggestions. If this is crazy/impossible, that's fine. The better plan is to some day create discussions for each suggestion. But if it's as simple as adding discussion back to the context menu, we might consider as a follow up.

kalilsn:
I agree, the table indicators are pretty confusing. Part of the issue there is that the ranges aren't joining the way they do in text. That's probably worth fixing, but I think we've reached the limits of what we can represent visually and will need to generate descriptions of the changes like google:

image

And that nicely segues into the second issue:

>People are going to want to add discussions to ranges containing suggestions. If this is crazy/impossible, that's fine. The better plan is to some day create discussions for each suggestion, as I know we've discussed. But if it's as simple as adding discussion back to the context menu, we might consider as a follow up.

The inline suggestion menu being exclusive with the other inline menu is unfortunate. It's also impossible to create a permalink to a suggestion. Perhaps I should change the approach so instead of two different menus, we just show different buttons when it's opened in a suggestion: replace the formatting buttons with the accept/reject buttons, but keep the existing menu style + the discussion/permalink buttons.

That said, I haven't actually tested if this works. The plan was always to eventually enable suggestions to be tied to discussions (like google), so suggestions actually have a discussion id as part of their data structure. [However, Ian seemed to believe that suggestions would make placing discussions in the right part of the sidebar hard](https://notes.knowledgefutures.org/pub/cthxjiio#refactoring-the-pub-sidebar).  I don't really want to build the full thing he's describing in this section yet, but I do think we might be able to just allow discussions without tying them to the suggestion in any way.

PR #2680, opened 2023-05-22
Inline suggestion info

### LINKED ISSUE #2591: Show author of suggestions inline

qweliant:
## Issue(s) Resolved
#2591 

## Test Plan
go to a pub
add suggested edits
be sure your avatar comes up
find another pub with suggested edits from not you [here](http://localhost:9876/pub/35ifjc2m/draft)

Kooha-2023-05-24-12-03-11


kalilsn (approved):
Looks great! I slightly prefer the version in your screenshot but let's see how everything looks together first

PR #2690, opened 2023-05-24
feat: put custom scripts behind feature flag

3mcd:
## Issue(s) Resolved

N/A

## Test Plan

1. Visit a community with the `"customScripts"` feature flag enabled (e.g. https://demo.duqduq.org). You should be able to add custom JS via the dashboard.
2. Visit a community without the `"customScripts"` feature flag enabled. You should not be able to add custom JS.


qweliant:
i think you may need to enbale the FF for pubpub before test pass, but LGTM when you do

PR #2706, opened 2023-06-01
feat: add popover to suggested edits toggle button

### LINKED ISSUE #2593: Add tooltips for suggestion mode and resolution buttons
### What?
Add user friendly descriptions of how the suggested edit controls work

### Why?
We'll launch this feature with some big caveats and limitations, so we need to communicate those to our users

### Initial estimate
2 days

- Note which block level elements don't work with suggested edits
- Mention that this feature is in development
- Mention that releases can't be created without resolving suggestions

3mcd:
## Issue(s) Resolved

#2593

## Test Plan

1. Hover the suggestion mode toggle button. You should see a popover with some text that describes its function.

### DIFF
--- a/client/components/FormattingBar/buttons.ts
+++ b/client/components/FormattingBar/buttons.ts
@@ -41,6 +41,7 @@ import {
 	FormattingBarButtonDataControls,
 	FormattingBarButtonPopoverCondition,
 } from './types';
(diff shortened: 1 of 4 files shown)


qweliant (approved):
LGTM!

PR #2707, opened 2023-06-01
Bh/sequelize model refactor

howebenhowe:
rebased refresh of #2478, addresses #2477

Sweeping cleanup + trimming of model definitions.
does not yet guarantee implementation of both sides of associations

__test plan__
run api test script with `npm run test-dev`

__completed test__
together with @kalilsn I spun up a local database with current tables as created from `master`'s sequelize definitions, dumped that database, then created db again from this branch's definitions, dumped, and diffed the dump files. The table specifications were identical.

### DIFF
--- a/server/activityItem/__tests__/fetch.test.ts
+++ b/server/activityItem/__tests__/fetch.test.ts
@@ -51,6 +51,7 @@ const models = modelize`
 				Pub pub {
 					title: "Generic Pub"
 					Release release {
+						User {}
 						createdAt: "2021-01-01"
 						historyKey: 1
 					}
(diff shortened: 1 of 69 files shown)


howebenhowe:
@tefkah Here is the aborted work of yore.
I've rebased to bring it up to speed -- I _believe_ this branch would be a suitable place to fold into your own sequelize work.
Notably, though, this branch doesn't quite guarantee that every association is explicitly [defined for both related models](https://sequelize.org/docs/v6/core-concepts/assocs/#why-associations-are-defined-in-pairs).

PR #2713, opened 2023-06-07
feat: add file name in assets.pubpub.org uploads

### LINKED ISSUE #2708: Preserve filename of uploaded custom downloads
Preserve filename of uploaded custom downloads

Feel better about using custom PDFs

## Optional
We'll need some way to insure there are no collisions — easiest is probably appending a timestamp to the uploaded filename.

### Who requested this? (If not you)
Arcadia

3mcd:
## Issue(s) Resolved

Resolves #2708

## Test Plan

1. Upload a file (Pub media, download, etc)
2. Ensure the filename in the assets.pubpub.org URL looks like `https://assets.pubpub.org/{filename}-{timestamp}`

### DIFF
--- a/client/utils/upload.ts
+++ b/client/utils/upload.ts
@@ -26,10 +26,11 @@ const checkForAsset = (url): Promise<void> => {
 
 const getFileNameForUpload = (file: File) => {
 	const folderName = isProd() ? generateHash(8) : '_testing';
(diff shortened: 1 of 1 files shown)


gabestein (approved):
LGTM.

Note: I'm getting the same WSOD and odd error loading pre-existing pubs on this dev app as in https://github.com/pubpub/pubpub/pull/2714.

Probably not related, but worth checking before merge.

PR #2723, opened 2023-06-13
Upgrade sequelize and improve tests

### LINKED ISSUE #2575: Sequelize upgrade
### What?
Upgrade sequelize 

### Why?
There are a bunch of sequelize vulnerabilities which we're not currently affected by but could be in the future: https://github.com/pubpub/pubpub/pull/2539#issuecomment-1478498620

tefkah:
## Issue(s) Resolved

#2575 

## Test Plan

Make all tests pass, check if `pgdump` produces same results.

## Notes

The main change of this PR is to upgrade `sequelize` from v5 to v6. The motivation for this being both a routine upgrade and the (possible) typesafety improvements that come with it (more on that later).

This PR seems really big (it's probably too big, see below for suggestions on making it smaller) but most of the changed lines are from `package-lock.json` and indentation in the `model.ts` files!.

As a secondary change, this PR also adds (some) typesafety to the testing tools in `stubstub`, addresses some memory leaks in the tests which caused the tests to fail on my machine, and adds some minor typesafety improvements in other places.

### Future work

Since this is the most important I'll mention this first.

#### Typesafety

This PR does not add any type inference to the sequelize model definitions, as that is not something sequelize can do (at the moment). In a separate PR i will propose a way to do this.

#### Incorporating #2707

I first want to make sure nothing breaks when upgrading to sequelize v6, and then I will incorporate #2707. I will do this in a separate PR, as this PR is already quite large, and doing this at the same time will obfuscate the changes.

### Sequelize upgrade

The main change the upgrade to `sequelize` v6 induced is the deprecation of `sequelize.import`.
This means that rather than defining the sequelize instance in `server/models.ts` and then calling `sequelize.import` a bunch of times, we now define the `sequelize` instance in `server/sequelize.ts`, import it in every model file, and then call `sequelize.define` using that instance.
We then export the models from `server/models.ts` and the `sequelize` instance from `server/models.ts`, as before.

Result: everything still works!

#### Why move the sequelize instance to a different file?

I chose to instantiate the model in a different file instead of keeping it in `server/models.ts`, because otherwise we would be dealing with a circular dependency between `server/models.ts` and `server/<model>/model.ts`. I am not sure if this would actually lead to problems, but it seemed like a good idea to avoid it.

#### Why the `as any` after every model definiton?

`sequelize.define` does not do any type inference on its own, so without the `as any` every model would be defined as `ModelCtor<Model<any,any>>`, which would cause a bunch of type-errors. I chose to include the `as any` for now as that is what the models were defined as before, but I want to get rid of this as soon as possible.

#### Why not include #2707 here?

This PR is already wayy to big, I wanted to keep it as minimal as possible. I will include #2707 in a separate PR.

### Testing tools

While doing this upgrade I ran into a lot of nasty errors when testing, and I found it hard to tell whether they were due to the upgrade or due to something else I did, or just general errors.

To gain a better understanding of how the tests were run, I added types to most of the files in `stubstub`. I decided not to remove them for now, because I thought it might be useful to have them in the future. If you do think this is too much to include (in this PR or in general) let me know and I can remove it!

##### Database not being initialized properly

At first, because I separated the `sequelize` instance from `server/models.ts`, the database was not initialized properly when running the tests. This was because the models were not imported in the global setup, and thus when `sequelize.sync()` was called there it did not create any models! But, because in the `setup` function in `stubstub/prepare.ts` also does `sequelize.sync()`, the database was eventually populated, but only after the first test that did that setup was run, which wasn't always the first test.
As a result, depending on the order jest decides to run the tests in, the first few tests would sometimes fail because the database was not initialized properly.

To fix this, I re-export the `sequelize` instance from `server/models.ts` (where all the models are also imported/exported from), and then import the `sequelize` instance from there in `stubstub/global/setup.ts`. This way all the models are properly loaded.

I added a note that the sequelize instance should only be imported from `server/models.ts` to avoid this problem in the future.

##### Missing Feature Flags

The tests for the routes would sometimes randomly fail because the `customScript` feature flag wasn't set properly.

In short: normally the `server/customScript/__test__/api.ts` always runs pretty early in the testing suite. This test creates the `customScript` feature flag in the db, and so the routes (which check for this feature flag) will operate fine.

However, because of the large amount of changes in this PR, jest decided that there is a better way to run the tests sometimes (or something, no idea how jest does this), having the routes tests run before the `customScript` test, causing them to fail.

~To fix this, I created a custom test sequencer that always runs the `customScript` test first. This is not ideal, but it works for now. Please let me know if you have a better idea!~

Now this feature flag will be created during global setup, that way we don't need a sequencer.

##### Memory leaks

TL;DR: Some things were leaking memory, causing tests to fail on my machine. The main culprits are packages that monkey-patch core Node functionality such as `Request` and `fs`, and not closing connections to the database. The other seems to be an issue with Jest, and is solved by using a different runtime.

This was by far the most time consuming part of this PR: I could not run the tests without Node running out of memory. What's even weirder is that this also happened/happens on the master branch! But the tests run fine elsewhere, so I thought it was a problem with my machine. For ref, I was running the tests on an Macbook Pro M1 Max with 64GB of RAM, so I don't think it was per se a physical memory issue as CI runs on way less memory (presumably, I don't know how much memory CI has).

SCR-20230613-osnl

I've tried a lot of things to try and fix this, but here is what works:

- Switching to node 16.10 (https://github.com/jestjs/jest/issues/11956). After 16.11, Node apparently changed how they do VMs or something, which makes Jest very bad at cleaning up after itself. This is not ideal however, as we use v18.7. While the tests do not use any Node functionality that's in 18+ but not in 16.10 (like native fetch), this would cause a problem in the future.
- Adding `@side/jest-runtime` as the runtime. This basically halved the leaked memory from ~100mb per test to ~40mb per test. Still not ideal, but otherwise still breaks on my machine.
- Upgrading to `passport@0.5.0` from `0.4.0`. Known issue that it leaks in Jest (https://github.com/jaredhanson/passport/issues/711)
- Stubbing `express-slow-down` in tests. It leaks, but there's no fix. Since we don't test it, and it only affects `/pubDocument` routes, I thought it was fine to just stub it out.
- Closing the sequelize connection after all tests. In `setup(beforeAll)`, `sequelize.sync()` is called, which (not sure) opens up a new connection to the db, one which isn't closed. This is also an issue on master. To combat this, I added a simple `sequelize.close()` to `teardown`. This does yield some errors, but the tests still pass so 🤷
- Instead of providing the full Express app to `supertest`, create a connection to the `app` and pass that. This is a known `supertest` issue (see https://github.com/ladjs/supertest/issues/634), although this is also fine on Node 16.10.
  By passing just a connection, we can then close that connection after the tests are done.
  I also slightly changed how the `supertest` agent is created, upgraded to latest version of `supertest`, removed `supertest-as-promised`, as `supertest` now provides its own promises.

All this combined allows for somewhat stable memory allocation.
Some packages, such as `firebase`, still leak, but upgrading that is quite the undertaking, and only a few tests use it anyway.

SCR-20230613-prfu

#### Other changes to testing

- Upgraded Jest and ts-jest to latest version
- Disabled typechecking by default for jest, as that is already done by `eslint` and when compiling the app
- Added `npm run test-no-lint` to test without linting, as that is quite slow.

##### Updated

- `sequelize`: v5.22 to v6.31. The main point of this PR.
- `express-slow-down`: v1.4 to v1.6. Originally to debug the memory leak.
- `passport`: v0.4.0 to v0.5.0. To solve a memory leak issue

##### Updated

- `jest`: v26.6 to v29.5
- `supertest`: v4.0.2 to v.6.3.3
- `ts-jest`: v26.4.4 to v29.1.0
- `typescript`: v4.4.4 to v4.5.5. Fixes an issue with `jest-environment-jsdom`

##### Added

- `jest-environment-jsdom`: No longer included in `jest` by default, but needed for `supertest` to work properly.
- `@side/jest-runtime`: to fix the memory leaking of jest in Node 18+

##### Types packages added

I also took the (completely self-indulgent) liberty to add some types packages for some untyped libraries that were used.
I took care to match the version of the types package to the version of the library, and to only add types packages for libraries that were used in the codebase.

I discovered these missing types by temporarily setting `noImplicitAny` to `true` in `tsconfig.json`, which caused the compiler to complain about missing types.

If undesired these can be removed!

- `@types/request-promise`: ^4.1.48,
- `@types/rss`: ^0.0.30,
- `@types/sinon`: ^7.5.2,
- `@types/supertest`: ^2.0.12,
- `@types/unidecode`: ^0.1.1,
- `@types/uuid`: ^3.4.10,
- `@types/lodash.mergewith`: ^4.6.7,
- `@types/lodash.throttle`: ^4.1.7,
- `@types/lodash.unescape`: ^4.0.7,
- `@types/mudder`: ^1.0.0,
- `@types/passport`: ^0.4.7,
- `@types/pegjs`: ^0.10.3,
- `@types/express-session`: ^1.17.7,
- `@types/express-slow-down`: ^1.3.2,
- `@types/express-sslify`: ^1.2.2,
- `@types/file-saver`: ^2.0.5,
- `@types/filesize`: ^4.2.0,
- `@types/fs-extra`: ^8.1.2,
- `@types/fuzzysearch`: ^1.0.0,
- `@types/graphlib`: ^2.1.8,
- `@types/hasbin`: 1.2,
- `@types/jest`: ^29.5.2,
- `@types/jsonpath`: ^0.2.0,
- `@types/jsonwebtoken`: ^9.0.2,
- `@types/amqplib`: ^0.5.3,
- `@types/color`: ^3.0.3,
- `@types/compression`: ^1.7.2,
- `@types/cookie-parser`: ^1.4.3,
- `@types/cors`: ^2.8.13,
- `@types/crypto-js`: ^4.1.1,
- `@types/dateformat`: ^3.0.1,


.jest/setup-env.js
```diff
@@ -61,7 +61,15 @@ if (typeof document !== 'undefined') {
...
 }
 
 global.fetch = () => new Promise((resolve) => setTimeout(resolve), 1e4);
+
+if (typeof window === 'undefined') {
+	/**
+	 * This is here because Jest overrides setImmediate in a strange way.
+	 * This fixes @see{../utils/async/__tests__/async.test.ts}
+	 */
+	global.setImmediateNode = setImmediate
```
tefkah, line 74: This is a weird issue, maybe introduced in Jest 27+

client/components/Editor/utils/__test__/changes.test.ts
tefkah, line 1: Jest 28+ has a different snapshot format, more native js like

client/components/UserNotifications/ThreadNotifications.tsx
```diff
@@ -21,7 +21,7 @@ type Props = {
 
 const getAccentColors = (communityAccentColor: string, isRead: boolean) => {
 	const opaque = isRead ? '#aaa' : communityAccentColor;
-	const transparent = Color(opaque).alpha(0.04);
+	const transparent = Color(opaque).alpha(0.04).toString();
```
tefkah, line 24: this is to fix a type-error, which became known after adding types for the color package

custom-sequencer.js
tefkah, line 1: This is to fix the issue where the feature flag `customScripts` is only created in the db during the `customScript` test.  the `routes` tests rely on this flag existing.

This is just the default jest sequencer which forces the `customScripts` test to run first. not a very good solution, but it works

jest.config.js
```diff
@@ -10,9 +13,19 @@ module.exports = {
 		'@pubpub/deposit-utils/datacite':
 			'<rootDir>/node_modules/@pubpub/deposit-utils/dist/cjs/datacite/index.js',
 	},
-	globalSetup: '<rootDir>/stubstub/global/setup.js',
-	globalTeardown: '<rootDir>/stubstub/global/teardown.js',
+	transform: {
+		'^.+\\.[tj]sx?$': [
+			'ts-jest',
+			{
+				diagnostics: false,
```
tefkah, line 20: I disabled typechecking for tests, because `npm run test` already runs eslint which also does typechecking.
Remove that line to re-enable typechecking

jest.config.js
```diff
@@ -10,9 +13,19 @@ module.exports = {
...
+			},
+		],
+	},
+	moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+	globalSetup: '<rootDir>/stubstub/global/setup.ts',
+	globalTeardown: '<rootDir>/stubstub/global/teardown.ts',
+	runtime: '@side/jest-runtime',
 	setupFiles: ['<rootDir>/.jest/setup-env.js'],
 	testPathIgnorePatterns: ['__tests__/data'],
-	resolve: {},
```
tefkah, line 17: this didn't do anything

jest.config.js
```diff
@@ -10,9 +13,19 @@ module.exports = {
...
+			'ts-jest',
+			{
+				diagnostics: false,
+			},
+		],
+	},
+	moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+	globalSetup: '<rootDir>/stubstub/global/setup.ts',
+	globalTeardown: '<rootDir>/stubstub/global/teardown.ts',
+	runtime: '@side/jest-runtime',
```
tefkah, line 27: This is to lower memory consumption in node 18+

package.json
```diff
@@ -99,7 +100,7 @@
 		"classnames": "^2.2.6",
 		"color": "^3.1.1",
 		"compression": "^1.7.4",
-		"connect-session-sequelize": "^6.0.0",
+		"connect-session-sequelize": "^7.1.7",
```
tefkah, line 103: for compat with sequelize v6

package.json
```diff
@@ -207,7 +208,6 @@
 		"styled-components": "^5.3.0",
 		"tmp-promise": "^2.0.1",
 		"ts-node": "^9.0.0",
-		"typescript": "^4.4.4",
```
tefkah, line 210: should go in dev deps

server/activityItem/model.ts
tefkah, line 1: in general, all the sequelize redefs look like this.
Most of the changes are indentation, which is annoying

server/collectionPub/model.ts
```diff
@@ -1,37 +1,39 @@
...
+					onDelete: 'CASCADE',
+					as: 'pub',
+					foreignKey: 'pubId',
+				});
 			},
 		},
-	);
-};
+	},
+) as any;
```
tefkah, line 39: i have added "as any" for now to ensure compatibility with the rest of the codebase, as before the models were also interpreted as any

server/community/model.ts
```diff
@@ -1,129 +1,131 @@
...
-					Community.belongsTo(SpamTag, {
-						as: 'spamTag',
-						foreignKey: 'spamTagId',
-					});
-				},
+		/* Set by Associations */
+		organizationId: { type: dataTypes.UUID },
+	},
+	{
+		// @ts-expect-error ts(2345): Argument of type '{ classMethods: { associate: (models: any) => void; }; }' is not assignable to parameter of type 'ModelOptions<Model<any, any>>'. Object literal may only specify known properties, and 'classMethods' does not exist in type 'ModelOptions<Model<any, any>>'.
```
tefkah, line 82: since the sequelize define's now have proper typechecking, this raises an error, so we have to `ts-expect-error` it

server/discussionAnchor/__tests__/queries.test.ts
```diff
@@ -26,6 +26,8 @@ setup(beforeAll, async () => {
 	await models.resolve();
 });
 
+teardown(afterAll);
```
tefkah, line 29: ive added `teardown` everywhere `setup` is used to close the connection with `sequelize`

server/models.ts
```diff
@@ -128,3 +71,16 @@ Object.values(sequelize.models).forEach((model) => {
...
+
+    In previous versions of Sequelize, the models were defined by importing it
+    in the object using `Sequlize.import`. This is no longer the case, instead
+    we first create the sequelize object and then import it in the relevant 
+    `model.ts` file.
+
+    TD:DR if you get "Model is not defined" errors, make sure you import the
+    `sequelize` object from this file, not form `server/sequelize.ts`.
+*/
+export { sequelize };
```
tefkah, line 86: this re-export is here to make sure that if you import `sequelize` all the models have been defined already

server/activityItem/model.ts
```diff
@@ -1,27 +1,28 @@
...
+		pubId: { type: dataTypes.UUID },
+		payload: { type: dataTypes.JSONB },
+		timestamp: {
+			type: dataTypes.DATE,
+			defaultValue: NOW,
+			allowNull: false,
 		},
-		{
-			indexes: [
-				{ fields: ['communityId'], method: 'BTREE' },
```
tefkah, line 20: this is the one change to the `define`s. `method` is not a valid property, i don't think it ever was checking the sequelize v4 and v5 docs. But I think `BTREE` is the default index method anywan

server/routes/pubDocument.tsx
```diff
@@ -126,12 +127,19 @@ const getEnrichedPubData = async (options: EnrichedPubOptions) => {
...
+ * While testing, this is not very useful anyway.
+ */
+const speedLimiter: RequestHandler =
+	process.env.NODE_ENV === 'test'
+		? (req, res, next) => next()
+		: slowDown({
+				windowMs: 60000, // 1 minute for requests to be kept in memory. value of 60000ms is default but expressed here for clarity
+				delayAfter: 60, // allow 60 requests per minute, then...
+				delayMs: 100, // 60th request has a 100ms delay, 7th has a 200ms delay, 8th gets 300ms, etc.
+				maxDelayMs: 20000, // max time of request delay will be 20secs
```
tefkah, line 141: Important: I changed maxDelay to maxDelayMs: `maxDelay` is not a property you can set, so did not do anything. This might cause unexpected behavior now it is actually set, although i've not noticed anything

server/routes/pubDocument.tsx
```diff
@@ -126,12 +127,19 @@ const getEnrichedPubData = async (options: EnrichedPubOptions) => {
...
-	delayAfter: 60, // allow 60 requests per minute, then...
-	delayMs: 100, // 60th request has a 100ms delay, 7th has a 200ms delay, 8th gets 300ms, etc.
-	maxDelay: 20000, // max time of request delay will be 20secs
-});
+/**
+ * `express-slow-down` causes a memory leak in Jest while testing.
+ * While testing, this is not very useful anyway.
+ */
+const speedLimiter: RequestHandler =
+	process.env.NODE_ENV === 'test'
```
tefkah, line 135: Disable this on test, bc it (monkey-)patches node `http` and therefore causes a memory leak during tests. Since we don't test for this (maybe we should) i thought it could be okay to just ignore it

stubstub/global/setup.ts
```diff
@@ -0,0 +1,59 @@
...
+
+// HACK(ian): The PUBPUB_SYNCING_MODELS_FOR_TEST_DB flag tells the code that we're only going to use
+// it to sync Sequelize to a test database, and that it's safe to skip certain operations.
+// In practice, we do this for the stupidest possible reason: because I can't figure out how to
+// get any modules required() from here to acknowledge absolute import paths like "server/utils"
+// instead of "../../server/utils". Since we use this absolute style extensively, we have to
+// carefully limit what we require from here. Luckily, all we need is server/models.ts, and
+// the handful of utilities that it must import.
+//
+// This is absurd, but it works. If you know how to fix this, please do!
```
tefkah, line 19: the hack is still needd, but i just set it manually below. could reinstate the function wrapper if desired, probably should just remove it

stubstub/global/setup.ts
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable no-console, global-require */
+// require('ts-node').register({
```
tefkah, line 2: this is not required, `ts-jest` already loads `ts-node`

should probably remove it

stubstub/global/teardown.ts
```diff
@@ -0,0 +1,13 @@
+// (this global is defined next door in setup.js)
+
+import { ChildProcessWithoutNullStreams } from 'child_process';
+
+declare namespace global {
```
tefkah, line 5: just add lil typesafety.

stubstub/modelize/builders.ts
```diff
@@ -0,0 +1,281 @@
...
+ *
+ * type MyTypeWithNullAsAlsoUndefined = WithOptional<MyType>;
+ * // { a: string; b: number; c: boolean | null | undefined; d: string | null | undefined; }
+ *
+ * type MyTypeWithNullAsAlsoUndefinedAndOptionalA = WithOptional<MyType, 'd'>;
+ * // { a: string | undefined; b: number; c: boolean | null | undefined; d: string | null | undefined; }
+ * ```
+ *
+ */
+type WithOptional<
```
tefkah, line 56: small typehelper to make typescript happy

stubstub/modelize/builders.ts
```diff
@@ -0,0 +1,281 @@
...
+> = {
+	[PossiblyNullKey in keyof WithOptionalBase<T, K>]: WithOptionalBase<
+		T,
+		K
+	>[PossiblyNullKey] extends null
+		? WithOptionalBase<T, K>[PossiblyNullKey] | undefined
+		: WithOptionalBase<T, K>[PossiblyNullKey];
+};
+
+export const builders = {
```
tefkah, line 74: main change here is that this is from

```ts

const builders = {}

builders.User = ...
```

to 

```ts

const builders = {

User: //

}
```

for better type inference

stubstub/modelize/link.ts
```diff
@@ -2,41 +2,105 @@
...
 		throw new Error(`Modelize cannot find a model named ${modelName}`);
 	}
 	return model;
 };
 
-const walkAst = (ast, parameters) => {
-	const definitionsByBoundName = {};
-	const definitions = [];
-	const resolveIdentifiersCallbacks = [];
+export type ModelDefinition = {
```
tefkah, line 16: these types are inferred from the parser, which i did not change.
I think they are accurate!

tools/5to6/v5/models.js
tefkah, line 1: this should maybe not be changed?

custom-sequencer.js
3mcd: Could we create the feature flag in the setupFiles instead of using a custom sequencer?

jest.config.js
3mcd: Oooh, does this speed up the tests at all?

server/facets/create.ts
```diff
@@ -28,7 +29,7 @@ const getSequelizePropsDefinition = (props: FacetProps) => {
 };
 
 export const createSequelizeModelsFromFacetDefinitions = (sequelize: Sequelize) => {
-	const FacetBinding = sequelize.import('./models/facetBinding') as any;
+	// const FacetBinding = sequelize.import('./models/facetBinding') as any;
```
3mcd, line 32: Why was this commented out? Was it unused/unnecessary?

server/login/api.ts
```diff
@@ -109,7 +109,7 @@ app.post('/api/login', (req, res, next) => {
 			return updatedUserData[1][0];
 		})
 		.then((user) => {
-			// @ts-expect-error
+			// // @ts-expect-error
```
3mcd, line 112: Looks like a double comment snuck in here!

server/logout/api.ts
```diff
@@ -8,7 +8,7 @@ app.get('/api/logout', (req, res) => {
 		...(isProd() && req.hostname.indexOf('pubpub.org') > -1 && { domain: '.pubpub.org' }),
 		...(isDuqDuq() && req.hostname.indexOf('pubpub.org') > -1 && { domain: '.duqduq.org' }),
 	});
-	// @ts-expect-error
+	// // @ts-expect-error
```
3mcd, line 11: Another double comment.

server/routes/pubDocument.tsx
3mcd: Awesome, thanks for pointing this out. We'll keep an eye on it!

workers/queue.ts
```diff
@@ -108,15 +108,19 @@ const processTask = (channel) => async (message) => {
 	}, maxWorkerTimeSeconds * 1000);
 };
 
+const cloudAmqpUrl = process.env.CLOUDAMQP_URL;
+if (!cloudAmqpUrl) {
+	throw new Error('CLOUDAMQP_URL environment variable not set');
+}
```
3mcd, line 114: Nice!

tsconfig.json
```diff
@@ -67,6 +67,7 @@
 			"utils/*": ["./utils/*"],
 			"facets": ["./facets"],
 			"facets/*": ["./facets/*"]
+			//		"sequelize": ["./types/test.d.ts"]
```
3mcd, line 70: Should this just be removed?

stubstub/modelize/builders.ts
3mcd: Awesome!

3mcd (changes requested):
Looks great. I left a few questions and small nits. I'm going to go ahead and do some smoke testing while those are resolved. Then we can hopefully get this merged by EOD.

jest.config.js
tefkah: hmmm not sure honestly! let me check

server/facets/create.ts
tefkah: ah sorry, that should have been removed

you cannot use `sequelize.import` anymore in v6, so its now defined with sequelize.define in `models/facetBingding`

tsconfig.json
tefkah: yes, this was for the type inference things

3mcd (approved):

PR #2724, opened 2023-06-13
fix: few small tweaks to address db connection timeouts

3mcd:
A few potential fixes for deadlocks/timeouts w/ db connections.


gabestein (approved):
👍

PR #2730, opened 2023-06-15
fix:return 404 on absent zotero integration

### LINKED ISSUE #2728: Zotero call filed every time a citation is opened, even if Zotero not connected
1. Open a pub with an account that is not connected to Zotero and open dev console
2. Make a citation
3. Close citation
4. Re-open citation
5. Notice call to /api/zotero, which eventually failes

No call to zotero endpoint.

howebenhowe:
## Issue(s) Resolved
#2728 

## Test Plan
perform same steps as in issue above:

1. Open a pub with an account that is not connected to Zotero and open dev console
2. Make a citation
3. Close citation
4. Re-open citation
6. Notice call to /api/zoteroIntegration, which eventually fails

### DIFF
--- a/server/zoteroIntegration/api.ts
+++ b/server/zoteroIntegration/api.ts
@@ -6,7 +6,9 @@ app.get(
 	'/api/zoteroIntegration',
 	wrap((req, res) =>
 		getZoteroIntegration(req.user.id).then((integration) =>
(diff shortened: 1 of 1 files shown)


gabestein:
I think the test that failed is just flaky activity item, right? Can someone confirm before approving.

3mcd (approved):
LGTM. Failing test is the flaky activity items test.

PR #2733, opened 2023-06-15
fix: batch Cite.async calls to avoid timeouts in Pubs with many structured citations

### LINKED ISSUE #2729: Some structure citations error out unless a space is present at the end of the URI
1. Visit the pub below.
2. Search for an errored citation.
3. Toggle between putting a space (or line break) after the structured citation
4. Notice that with space/break, the citation loads. Without, it errors.

Citation should load without a trailing space/line break.

https://homeworld.pubpub.org/pub/aqueous-solvents/

3mcd:
## Issue(s) Resolved

Resolves #2729

## Test Plan

1. Open a Pub with many (100+) structured citations.
2. The citations should load properly (i.e. not display as "Error")

### DIFF
--- a/server/utils/citations/structuredCitations.ts
+++ b/server/utils/citations/structuredCitations.ts
@@ -1,3 +1,5 @@
+/* eslint-disable no-restricted-syntax */
+
 import fs from 'fs';
 import path from 'path';
 import crypto from 'crypto';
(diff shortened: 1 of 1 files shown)


gabestein (approved):
Thanks for the quick fix here. LGTM.

PR #2746, opened 2023-06-29
add noindex meta to all non-prod pages

gabestein:
Per [this](https://support.google.com/webmasters/answer/9689846?hl=en#make_permanent), we shouldn't rely on robots.txt to block indexing. Instead, per [this](https://developers.google.com/search/docs/crawling-indexing/block-indexing), we should put meta tags on all pages that shouldn't be followed.

This does that for all non-prod sites. It's a little brute force, so if there are better places to pass down unlisted, happy to entertain that.

## Issue(s) Resolved
n/a

## Test Plan
1. Visit on local or instance and make sure robots noindex meta is set
2. Set pubpub prod environment locally and make sure noindex is _not set_ on most pages, but _is_ set on, e.g., a dashboard page.

### DIFF
--- a/server/utils/__tests__/ssr.test.tsx
+++ b/server/utils/__tests__/ssr.test.tsx
@@ -14,6 +14,10 @@ const defaultProps = {
 
 const MetaComponents = (props) => <React.Fragment>{generateMetaComponents(props)}</React.Fragment>;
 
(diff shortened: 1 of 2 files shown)


kalilsn (approved):
I ran through the test plan and both steps look good!

PR #2747, opened 2023-06-29
fix: Impact iframe positioning

gabestein:
Metabase updated its embed styling to put the filter on the left instead of the right, which caused the header of the Impact tab to overlap it on some browser sizes. This fixes that.

Screenshot 2023-06-29 at 22 03 18

Screenshot 2023-06-29 at 22 03 48

_Test Plan_
Visit an impact tab at a few browser sizes and make sure it works ok?


kalilsn (approved):
Looks good!

PR #2766, opened 2023-07-12
dev: allow actorId as a parameter for sequelize.create etc, and fix activityHook related type errors

tefkah:
This PR creates an override for a `sequelize` Model's options, such that e.g. these are typescript legal

```ts
Pub.create({...}, {actorId})
Pub.destory({actorId})
``` 

This patch affects every single model we use, even though not every model is able to use this actorId.
However, this was also the case before, and this is a much easier way of fixing this type issue.

This also fixes some other type errors related to activityHooks (where the `actorId` is used)

From 399 -> 343 errors in 99 files.

### DIFF
--- a/types/sequelize-actorid-override.d.ts
+++ b/types/sequelize-actorid-override.d.ts
@@ -1,0 +1,37 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import {
+	BuildOptions,
+	Logging,
+	Silent,
+	Transactionable,
+	TruncateOptions,
+	Paranoid,
+	Hookable,
+} from 'sequelize';
+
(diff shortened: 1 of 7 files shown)


types/sequelize-actorid-override.d.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import {
+	BuildOptions,
+	Logging,
+	Silent,
+	Transactionable,
+	TruncateOptions,
+	Paranoid,
```
3mcd, line 8: That's a new one 😆

3mcd (approved):

PR #2768, opened 2023-07-12
dev: expect types

tefkah:
Fix a class of type errors which result from #2765, namely that `Model.findOne` returns `null | Model`, the null case not being accounted for.


3mcd (approved):

3mcd (approved):

tefkah:
Down to 271 errors in 79 files

client/components/AttributionEditor/AttributionDetailControls.tsx
```diff
@@ -98,7 +98,7 @@ const AttributionDetailControls = (props: Props) => {
 			<div className="right-details">
 				<InputGroup
 					placeholder="Affiliation"
-					defaultValue={affiliation}
+					defaultValue={affiliation ?? undefined}
```
tefkah, line 101: I do this in a number of places where I cannot change the definition from e.g. `string | undefined` to `string | null | undefined`. Kind of ugly, and not very nice to have to continue doing. If you have a better idea let me know!

client/components/AttributionEditor/AttributionEditor.tsx
```diff
@@ -58,7 +59,9 @@ class AttributionEditor extends Component<Props> {
 
 	handleAttributionAdd(user) {
 		const { attributions, onUpdateAttributions, onPersistStateChange } = this.props;
-		const maxOrder = attributions.length ? Math.max(...attributions.map((a) => a.order)) : 0;
+		const maxOrder = attributions.length
+			? Math.max(...attributions.map((a) => a.order ?? 0))
```
tefkah, line 63: `a.order` can be `null`, think setting it to `0` would be best, but should be smoke tested

client/components/AttributionEditor/AttributionEditor.tsx
```diff
@@ -142,14 +145,16 @@ class AttributionEditor extends Component<Props> {
...
 
 		onUpdateAttributions(newAttributions);
 	}
 
 	render() {
 		const { attributions, canEdit, listOnBylineText, hasEmptyState } = this.props;
-		const sortedAttributions = attributions.concat().sort((a, b) => a.order - b.order);
+		const sortedAttributions = attributions
+			.concat()
+			.sort((a, b) => (a.order && b.order ? a.order - b.order : 0));
```
tefkah, line 157: Again, these are bc `a.order` can be `null`. This should be tested

client/components/AttributionEditor/EditableAvatar.tsx
```diff
@@ -9,7 +9,7 @@ require('./editableAvatar.scss');
 
 type Props = {
 	attribution: {
-		avatar?: string;
+		avatar?: string | null;
```
tefkah, line 12: This is how i general tried to solve the `string | null cannot be assigned to type string | undefined. null cannot be assigned to undefined` errors. In most cases this is enough, sometimes I need to add some extra `avatar ?? undefined` somewhere in the component bc e.g. blueprint does not accept null

client/components/Avatar/Avatar.tsx
```diff
@@ -39,7 +39,7 @@ const Avatar = (props: Props) => {
 		borderColor,
 		borderWidth: borderColor ? borderWidth || Math.floor(width / 50) + 1 : 0,
 		fontSize: isBlock ? Math.floor(width / 1.5) : Math.floor(width / 2.5),
-		backgroundColor: communityData.accentColorDark,
+		backgroundColor: communityData.accentColorDark ?? undefined,
```
tefkah, line 42: `accentColorDark` is one that comes up particularly often, maybe should set be set to `string` in the types

client/components/Deposit/DataciteDeposit.tsx
```diff
@@ -85,7 +85,7 @@ export default function DataciteDeposit(props: Props) {
 	const pub = 'pub' in props ? props.pub : undefined;
 
 	useEffect(() => {
-		if (pub && pub.releases.length > 0) {
+		if (pub && (pub.releases?.length ?? 0) > 0) {
```
tefkah, line 88: i am somewhat inconsistent with this, sometimes I do this, sometimes I do `pub && pub.releases && ...`. Let me know if i should make that more consistent

client/components/Email/Section.tsx
```diff
@@ -47,7 +47,7 @@ export const Section = ({
 		width={width}
 	>
 		<tr>
-			<td valign="top" style={{ padding: innerPadding, color }}>
+			<td valign="top" style={{ padding: innerPadding, color: color ?? undefined }}>
```
tefkah, line 50: This comes up somewhat often, `style` props do not accept `null`

client/components/Layout/Layout.tsx
```diff
@@ -67,7 +67,9 @@ const Layout = (props: Props) => {
 					<LayoutPagesCollections
 						content={block.content}
 						pages={communityData.pages}
-						collections={communityData.collections}
+						collections={
+							communityData.collections as DefinitelyHas<Collection, 'title'>[]
```
tefkah, line 71: This is another one that comes up quite often: Collections can theoretically not have titles, as the column in the Collection table can be `null`, but in practice it's assumed to be always assigned. Maybe set `@AllowNull{ false }` on the db and set type to `string`?

client/components/PubEdge/PubEdge.stories.tsx
```diff
@@ -17,7 +17,10 @@ const pubEdge = {
 		contributors: ['Richard Jenkins', 'Bradley Whitford', 'Sigourney Weaver'],
 		publicationDate: new Date().toISOString(),
 		avatar: 'https://assets.pubpub.org/9fk06ei5/41589421981678.jpg',
-	},
+		createdAt: new Date().toISOString(),
+		updatedAt: new Date().toISOString(),
+		doi: null,
+	} satisfies ExternalPublication,
```
tefkah, line 23: PubEdge just needs these types.

client/containers/App/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -103,7 +104,7 @@ const Breadcrumbs = (props: Props) => {
 				<Avatar
 					avatar={avatar}
 					initials={title[0]}
-					// @ts-expect-error ts-migrate(2322) FIXME: Type '{ avatar: any; initials: any; communityData:... Remove this comment to see the full error message
+					// @ts-expect-error FIXME: CommunityData does not exist on Avatar?
```
tefkah, line 107: I think i removed this error bc it worked at some point, then broke?

Anyway, this prop does not even exist on `Avatar`,  `Avatar` finds `communityData` through a provider, think this can be removed safely.

client/containers/DashboardOverview/CollectionOverview/collectionState.ts
```diff
@@ -28,7 +33,7 @@ export const useCollectionPubs = (options: UseCollectionPubsOptions) => {
 	} = usePageContext();
 	const { pendingPromise } = usePendingChanges();
 	const [collectionPubs, setCollectionPubs] = useState(() => sortByRank(initialCollectionPubs));
-	const [scopeSummary, setScopeSummary] = useState(collection.scopeSummary);
+	const [scopeSummary, setScopeSummary] = useState<MinimalScopeSummary>(collection.scopeSummary);
```
tefkah, line 36: changed this to `MinimalScopeSummary` (see those edits later) bc the `subtractScopeSummaries` and `addScopeSummaries` method do not include `createdAt`, `updatedAt` and `id`.

server/activityItem/fetch.ts
```diff
@@ -121,9 +122,7 @@ const applyFiltersToWhereQuery = (whereQuery: any, filters: ActivityFilter[]) =>
 	return whereQuery;
 };
 
-const fetchActivityItemModels = async (
-	options: Required<FetchActivityItemsOptions>,
-): Promise<types.ActivityItem[]> => {
+const fetchActivityItemModels = async (options: Required<FetchActivityItemsOptions>) => {
```
tefkah, line 125: In a lot of cases, explicit return or parameter types are no longer necessary, in which case i removed them.

server/activityItem/__tests__/fetch.test.ts
```diff
@@ -111,7 +112,9 @@ const models = modelize`
 
 setup(beforeAll, models.resolve);
 
-teardown(afterAll);
+teardown(afterAll, async () => {
+	await finishDeferredTasks();
+});
```
tefkah, line 117: This is in some ways the solution to the memory leak problems in the tests I ran into before.

Some tests use methods which create or update ActivityItems, which are deferred by that util. Sometimes, they do not finish before the test is over, so they stick around and Jest does not clean them up properly, leading to leaked memory.

How I solved this is by just forcefully closing the connection to the DB every time (this is in the `teardown` function), but that sometimes leads to (non-test-ruining) errors of `DB was closed before transaction was completed`, because the deferred tasks are still running. 
To solve that and to prevent possible future errors, I just wait until all deferred tasks are done before closing the connection.

Now that I think of it, it's probably better to just do this in the `teardown` function in general rather than adding it where it causes problems huh

server/activityItem/hooks.ts
```diff
@@ -1,8 +1,7 @@
-import * as types from 'types';
 import { ActivityItem } from 'server/models';
 import { defer } from 'server/utils/deferred';
 import { createNotificationsForActivityItem } from 'server/userNotification';
 
-ActivityItem.afterCreate((item: types.ActivityItem) => {
+ActivityItem.afterCreate((item) => {
```
tefkah, line 5: I did this for all hooks

server/activityItem/model.ts
```diff
@@ -9,34 +9,44 @@ import {
...
 
 @Table
-export class ActivityItem extends Model<
-	InferAttributes<ActivityItem>,
-	InferCreationAttributes<ActivityItem>
+// export class ActivityItem<Kind extends ActivityItemKind = ActivityItemKind> extends Model<
+// 	InferAttributes<ActivityItem<Kind>>,
+// 	InferCreationAttributes<ActivityItem<Kind>>
+// > {
+export class ActivityItem<T extends InsertableActivityItem = InsertableActivityItem> extends Model<
```
tefkah, line 19: This was necessary to turn activityItems  into a discriminated union of `{kind: 'pub-created', payload: { pub: ... } } | { kind: 'collection-created', payload: { collection: ... } }` instead of 
`{ kind: 'collection-created' | 'pub-created' | ..., payload: { pub: ... } | { collection: ... } | ... }`

server/activityItem/model.ts
```diff
@@ -9,34 +9,44 @@ import {
...
+	InferAttributes<ActivityItem<T>, { omit: Extract<keyof ActivityItem<T>, keyof T> }> & T,
+	InferCreationAttributes<ActivityItem<T>, { omit: Extract<keyof ActivityItem<T>, keyof T> }> & T
 > {
+	// this overrides the default Date type to be compatible with existing code
+	declare createdAt: CreationOptional<string>;
+	declare updatedAt: CreationOptional<string>;
```
tefkah, line 25: Ive added this to all Models. Maybe cleaner to create a new Model class that already takes care of this.

Ive chosen `string` because that is how the values are used in the code (almost always).
And, because when serialized that is the correct value.

If we left it at the default (`Date`), then `ActivityItem.toJSON()` would still have `updatedAt: Date`. 
Now, we override `toJSON` anyway, so we could also just make it turn `Date`s to `string`s properly, but this seemed simpler. Let me know if you think the other way is better though, I kind of think it is but will run into more edge cases.

server/activityItem/model.ts
```diff
@@ -9,34 +9,44 @@ import {
...
+
+	public declare toJSON: <M extends Model>(this: M) => RecursiveAttributes<M>;
```
tefkah, line 27: I also override `toJSON` on every model, because the default version does not serialize the associations. 
This helper `recursiveAttributes` does do that.

I tried to do it differently as well, e.g. by just setting the type of the associations to e.g. `Attributes<Pub>` instead of `Pub`, but that caused TypeScript to throw a lot of `Type instantiation is possibly infinitely deep` errors, which ya hate to see

server/activityItem/queries.ts
```diff
@@ -118,9 +117,9 @@ export const createCommunityCreatedActivityItem = async (
 export const createCommunityUpdatedActivityItem = async (
 	actorId: null | string,
 	communityId: string,
-	oldCommunity: types.Community,
+	oldCommunity: Community,
```
tefkah, line 120: I felt comfortable changing these because only full models are passed to these functions, never the serialized one.
Also, the serialized one threw errors, because the `getDiffsForPayload` function generates a `Diff`ed type between the result of `Community.findOne` and `types.Community` (different types obv), which cannot get passed to `createActivityItem`.

server/activityItem/queries.ts
```diff
@@ -217,16 +216,16 @@ export const createCollectionActivityItem = async (
...
 ) => {
-	const collection: types.Collection = await Collection.findOne({ where: { id: collectionId } });
+	const collection = expect(await Collection.findOne({ where: { id: collectionId } }));
 	const { title } = collection;
 	return createActivityItem({
 		kind,
 		collectionId,
 		actorId,
-		communityId: collection.communityId,
+		communityId: expect(collection.communityId),
```
tefkah, line 225: These `expect`s were very tricky to do, because in some cases `null` is allowed to be passed and _is_ passed in the tests, causing the test to fail. 
This won't per se be a problem in prod because of how the `expect` function is designed to only throw in not-`production`, but still, failing tests are annoying.

In some cases (see `types/activity`) i elected to change the type of the ActivityItem itself to be e.g. `title | null`. I think there were only two cases where this was necessary though, at the very least for the review title.

tefkah: Another solution is to modify the tests here to e.g. correctly assign a title to the review, but I'll leave that up to you!

server/activityItem/queries.ts
```diff
@@ -346,13 +349,17 @@ export const createPubReviewCommentAddedActivityItem = async (
...
+	const promises = await Promise.all([
 		ReviewNew.findOne({
 			where: { id: reviewId },
 		}),
 		ThreadComment.findOne({ where: { id: threadCommentId } }),
 	]);
-	const pub = await Pub.findOne({ where: { id: review.pubId } });
+
+	const review = expect(promises[0]);
+	const threadComment = expect(promises[1]);
```
tefkah, line 360: I do this un-destructuring in order to `expect` the result properly, otherwise I need to `expect` every instance where `review` is used. This way I only need to do it once.

server/activityItem/queries.ts
```diff
@@ -379,10 +386,10 @@ export const createPubReviewCommentAddedActivityItem = async (
 export const createPubReviewUpdatedActivityItem = async (
 	actorId: null | string,
 	reviewId: string,
-	oldReview: types.Review,
+	oldReview: ReviewNew,
 ) => {
-	const review: types.Review = await ReviewNew.findOne({ where: { id: reviewId } });
-	const pub: types.Pub = await Pub.findOne({ where: { id: review.pubId } });
+	const review = expect(await ReviewNew.findOne({ where: { id: reviewId } }));
+	const pub = expect(await Pub.findOne({ where: { id: expect(review.pubId) } }));
```
tefkah, line 392: Just to make sure this is reviewed at least once: I very often remove the hand typed return type like this, because it throws an error and is wrong, `findOne` does not return a serialized Model.

I could have also done `Review.findOne(...) as types.Review`. Because of the way the types are defined now this will not error, but it will still be incorrect.

server/activityItem/utils.ts
```diff
@@ -1,8 +1,10 @@
 import * as types from 'types';
 
 import { ActivityItem } from 'server/models';
+import { CreationAttributes } from 'sequelize';
 
-export const createActivityItem = (ai: types.InsertableActivityItem) => ActivityItem.create(ai);
+export const createActivityItem = <M extends ActivityItem>(ai: CreationAttributes<M>) =>
```
tefkah, line 6: Add type argument here to get better type inference.

server/collection/model.ts
```diff
@@ -75,7 +82,7 @@ export class Collection extends Model<
 
 	@Default('choose-best')
 	@Column(DataType.ENUM('none', 'minimal', 'medium', 'choose-best'))
-	readNextPreviewSize!: CreationOptional<ReadNextPreviewSize | null>;
+	readNextPreviewSize!: CreationOptional<ReadNextPreviewSize>;
```
tefkah, line 85: I do this once or twice: if the column has a default value I think it's pretty safe to remove the `| null`. 
Technically it can still be set to null manually, e..g `Collection.update({ readNextPreviewSize: null }, { where: ... }), while throwing a type error, will still run, but I couldn't find anywhere where that could possibly happen.

server/collection/permissions.ts
```diff
@@ -1,6 +1,27 @@
...
+	userId = null,
+	communityId = null,
+	collectionId = null,
+}: {
+	userId: string | null;
+	communityId: string | null;
+	collectionId: string | null;
+}): Promise<Permissions> => {
 	if (!userId) {
 		return {};
```
tefkah, line 26: This is something I do a number of times, explicitly type the permissions.

This was not to prevent any type errors, if anything it introduced more, but this proves to be useful when typing the queries more accurately, which is something we want to do eventually.

server/collection/queries.ts
```diff
@@ -141,7 +142,7 @@ const findCollectionOptions = {
...
 };
 
-export const findCollection = (
-	collectionId: string,
-): Promise<types.DefinitelyHas<types.Collection, 'community' | 'attributions'>> =>
-	Collection.findOne({ where: { id: collectionId }, ...findCollectionOptions });
+export const findCollection = (collectionId: string) =>
+	Collection.findOne({ where: { id: collectionId }, ...findCollectionOptions }) as Promise<
+		types.DefinitelyHas<Collection, 'community' | 'attributions'>
+	>;
```
tefkah, line 148: Not sure why it is typed like this other than convenience, but I decided to leave it to break as little as possible.

server/collectionAttribution/queries.ts
```diff
@@ -1,28 +1,34 @@
...
-export const createCollectionAttribution = (inputValues) => {
-	return CollectionAttribution.create({
+export const createCollectionAttribution = async (inputValues: {
+	userId?: string | null;
+	collectionId: string;
+	name: string;
+	order: number;
+	communityId?: string | null;
+}) => {
+	const newAttribution = await CollectionAttribution.create({
```
tefkah, line 12: This is another example of being a bit too enthusiastic: turning `then`d functions into async functions. The reason is almost always that it is easier to `expect` things like this, like below.

server/discussion/queries.ts
```diff
@@ -124,7 +125,14 @@ export const createDiscussion = async (options: CreateDiscussionOpts) => {
...
 	if (commenterId) {
-		await Discussion.update({ commenterId });
+		await Discussion.update(
+			{ commenterId },
+			{
+				where: {
+					id: newDiscussion.id,
+				},
+			},
+		);
```
tefkah, line 135: `Model.update` needs a `where`. But I don't think this ever gets called, running `jest` with `coverage` shows that at least in tests this never gets called.

server/doi/__tests__/api.test.ts
```diff
@@ -109,6 +109,7 @@ it('lets community admins create a DOI for pubs in their community', async () =>
 	});
 
 	expect(dois.pub).toEqual(expectedPubDoi);
+	// @ts-expect-error FIXME: Property 'deposit' does not exist on type 'object'.
```
tefkah, line 112: `depositJson` is not typed (was `any` first, now `object`). I don't know what the type is supposed to be, so just `ts-expect-error`ed it for now. If there is a type for this somewhere I can update it!

server/doi/queries.ts
```diff
@@ -85,9 +88,16 @@ export const persistCrossrefDepositRecord = async (ids, depositJson) => {
...
+	// this is just to make typescript happy, update cannot be called on the union
+	if (targetModel instanceof Pub) {
+		await targetModel.update({
+			crossrefDepositRecordId: crossrefDepositRecord.id,
+		});
+	} else {
+		await targetModel.update({
+			crossrefDepositRecordId: crossrefDepositRecord.id,
+		});
+	}
```
tefkah, line 100: This happens once or twice more somewhere, because `update` uses the `this` type of the model you cannot call it on the union, very annoying.

server/externalPublication/model.ts
```diff
@@ -41,5 +48,5 @@ export class ExternalPublication extends Model<
 	avatar!: string | null;
 
 	@Column(DataType.DATE)
-	publicationDate!: Date | null;
+	publicationDate!: string | null;
```
tefkah, line 51: Again, `Date` to string

server/landingPageFeature/propose.ts
```diff
@@ -11,7 +11,7 @@ type GetProposedFeatureOptions = {
 
 type ProposalResult = null | { pubId: string } | { communityId: string };
 
-const extractCommunityQueryFromHostname = async (hostname: string) => {
+const extractCommunityQueryFromHostname = (hostname: string) => {
```
tefkah, line 14: doesn't do anything async, wraps the type in a promise for no reason.

server/page/sanitizePageHtml.ts
```diff
@@ -1,6 +1,6 @@
 import sanitizeHtml from 'sanitize-html';
 
-export const sanitizePageHtml = (html) => {
+export const sanitizePageHtml = (html: string) => {
```
tefkah, line 3: compulsive typing

server/review/queries.ts
```diff
@@ -140,19 +143,19 @@ export const updateReview = async (inputValues, updatePermissions, userData) =>
...
+			const prevStatus = (await previousReview)?.status;
 			const wasClosed = prevStatus !== 'closed' && nextValues.status === 'closed';
 			const wasCompleted = prevStatus !== 'completed' && nextValues.status === 'completed';
 			if (wasClosed) {
-				return Promise.all[createClosedThreadEvent(userData, nextValues.threadId)];
+				return Promise.all([createClosedThreadEvent(userData, nextValues.threadId)]);
 			}
 			if (wasCompleted) {
-				return Promise.all[createCompletedThreadEvent(userData, nextValues.threadId)];
+				return Promise.all([createCompletedThreadEvent(userData, nextValues.threadId)]);
```
tefkah, line 158: this was realy weird, don't think this ever got called to begin with, but obviously `Promise.all[<randomObject>]` won't work.

server/routes/passwordReset.tsx
```diff
@@ -21,7 +21,11 @@ app.get(['/password-reset', '/password-reset/:resetHash/:slug'], (req, res, next
 			if (userData && userData.resetHash !== req.params.resetHash) {
 				hashIsValid = false;
 			}
-			if (userData && userData.resetHashExpiration < Date.now()) {
+			if (
+				userData &&
+				userData.resetHashExpiration &&
+				userData.resetHashExpiration < Number(new Date())
```
tefkah, line 27: Force cast date to number, `resetHashExpiration` is one of the few `Date`s that are typed as `number` instead of `string`

server/rss/queries.ts
```diff
@@ -253,16 +255,17 @@ export const getCommunityRss = async (communityData, query) => {
 
 	const feed = new RSS({
 		title: communityData.title,
-		description: communityData.description,
+		description: communityData.description ?? undefined,
 		feed_url: `${communityUrl}/rss.xml`,
 		site_url: communityUrl,
-		image_url: communityData.favicon,
+		image_url: communityData.favicon ?? undefined,
```
tefkah, line 261: Rss pretends to hate null

server/scopeSummary/__tests__/scopeSummary.test.ts
```diff
@@ -135,24 +136,30 @@ const expectSummaryFor = async (scopeIds, expectedValues) => {
 	const fetchModel = async () => {
 		const { pubId, collectionId, communityId } = scopeIds;
 		if (pubId) {
-			const { scopeSummary } = await Pub.findOne({
-				where: { id: pubId },
-				include: [{ model: ScopeSummary, as: 'scopeSummary' }],
-			});
+			const { scopeSummary } = assertExpect(
```
tefkah, line 139: rename `expect` here to prevent conflict with jest

server/submission/model.ts
```diff
@@ -28,7 +35,7 @@ export class Submission extends Model<
 	status!: SubmissionStatus;
 
 	@Column(DataType.DATE)
-	submittedAt!: Date | null;
+	submittedAt!: string | null;
```
tefkah, line 38: nother `Date` to `string`

server/threadEvent/permissions.ts
```diff
@@ -30,3 +40,9 @@ export const getPermissions = async ({ userId, communityId, pubId, threadId }) =
 		destroy: false,
 	};
 };
+
+export type ThreadEventPermissions = {
+	create?: boolean;
+	update?: false;
```
tefkah, line 46: so `update` can only ever be `false` here, but this runs into problems below

server/threadEvent/queries.ts
```diff
@@ -1,50 +1,61 @@
...
+export const updateThreadEvent = async (
+	inputValues: types.ThreadEvent & { threadEventId: string },
+	updatePermissions: ThreadEventPermissions['update'],
+) => {
 	// Filter to only allow certain fields to be updated
 	const filteredValues = {};
 	Object.keys(inputValues).forEach((key) => {
-		if (updatePermissions.includes(key)) {
+		// @ts-expect-error FIXME: updatePermissions can only be `false`, this shouldn't be here
+		if (updatePermissions && updatePermissions?.includes(key)) {
```
tefkah, line 39: `updatePermissions` can only be `false`, so the part after `&&` is `never`, which cannot have `includes`.

In general this pattern is badly typed (filteredValues is always `{}`), but I restrained myself from messing with it too much haha

server/user/model.ts
```diff
@@ -31,69 +33,73 @@ import {
...
-	declare setPassword: (password: string, cb: (err: any, user?: User) => void) => void;
+class ModelWithPassport<T extends {} = any, C extends {} = T> extends Model<T, C> {
+	declare setPassword: <M extends Model>(
+		this: M,
+		password: string,
+		cb: (err: any, user?: M) => void,
+	) => void;
 
-	declare setActivationKey: (cb: (err: any, user?: User) => void) => void;
+	declare setActivationKey: <M extends Model>(this: M, cb: (err: any, user?: M) => void) => void;
```
tefkah, line 43: I  updated the types like this to give better type inference.

server/user/model.ts
```diff
@@ -186,7 +194,7 @@ export class User extends ModelWithPassport<
 	googleScholar!: string | null;
 
 	@Column(DataType.DATE)
-	resetHashExpiration!: Date | null;
```
tefkah, line 189: `resetHasExpiration` is always treated as `number`, so I made it `number` instead of `string` like with the others.

server/userNotification/model.ts
```diff
@@ -45,7 +52,7 @@ export class UserNotification extends Model<
 	@Column(DataType.BOOLEAN)
 	manuallySetIsRead!: CreationOptional<boolean>;
 
-	@BelongsTo(() => ActivityItem, {
+	@BelongsTo(() => ActivityItem<InsertableActivityItem>, {
```
tefkah, line 55: This typing leaves something to be desired, leads to some annoying types. I think UserNotification can only have like `pub-commented` or `review-commented` or something, so this could be more precise.

server/utils/ssr.tsx
```diff
@@ -30,6 +30,9 @@ type MetaProps = {
 
 const contributorRoles = ['Writing – Review & Editing', 'Editor', 'Series Editor'];
 const sortAttributions = (foo: Attribution, bar: Attribution) => {
+	if (!foo.order || !bar.order) {
```
tefkah, line 33: did this  a number of times, bc again `order` can be `null`

server/utils/ssr.tsx
```diff
@@ -98,19 +101,19 @@ export const generateMetaComponents = (metaProps: MetaProps) => {
 			<meta
 				key="t2"
 				property="og:title"
-				content={useCollectionTitle ? collection.title : title}
+				content={useCollectionTitle ? collection.title ?? undefined : title}
```
tefkah, line 104: `content` cannot be null, so a lot of these silly things.
Again, maybe `collection.title` should just be `string`

tefkah:
brought up to speed with master

3mcd (changes requested):

3mcd (approved):

PR #2769, opened 2023-07-12
dev: add types for `passport-local-sequelize`

tefkah:
Types the User model with methods from `passport-local-sequelize`

From 399 -> 397 errors in 109 files.

### DIFF
--- a/server/passwordReset/queries.ts
+++ b/server/passwordReset/queries.ts
@@ -69,7 +69,7 @@ export const updatePasswordReset = (
 	return User.findOne({
 		where: whereQuery,
 	})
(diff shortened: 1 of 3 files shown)


server/passwordReset/queries.ts
```diff
@@ -78,7 +78,7 @@ export const updatePasswordReset = (
 			}
 
 			/* Promisify the setPassword function, and use .update to match API convention */
-			const setPassword = promisify((userData as any).setPassword.bind(userData));
```
tefkah, line 81: there are a few more of these strewn around, but they require some more effort to fix, should be done after other sub-prs are merged

3mcd (approved):
LGTM!

PR #2776, opened 2023-07-18
Remove "You may reply to this email thread to reach us." 

### LINKED ISSUE #2764: Emails with non-functional reply address say "You may reply to this email thread to reach us."
1. When someone submits via the submission workflow, they get an email beginning "You may reply to this email thread to reach us."
2. When you hit reply, the email is directed to "submissions@mg.pubpub.org", which bounces.
...

That the reply-to would be set to the email address associated with the pubpub community, or that the email would otherwise find its way to them.

Or alternatively, have the full email editable here:

image

yaxu:
This isn't true, unless a CC: is set and the recipient uses reply-all, so might well result in lost messages.

It would be nicer if the reply-to was set rather than cc:, so replying would work, but perhaps this is a useful quick fix, as it just brings this email in line with the others.

### DIFF
--- a/client/components/SubmissionEmail/SubmissionEmail.tsx
+++ b/client/components/SubmissionEmail/SubmissionEmail.tsx
@@ -42,8 +42,7 @@ const SubmissionEmail = (props: Props) => {
 		if (kind === 'received') {
 			return (
 				<>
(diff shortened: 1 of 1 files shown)


gabestein:
Fixed in #2827

yaxu:
Excellent, thanks!

PR #2785, opened 2023-07-25
Fix relative URL for "who we work with" to make it absolute

BlueReZZ:
## Issue(s) Resolved
Quick fix to an incorrect link in the footer on eLife


gabestein (approved):
Well, well, well, look who's out of retirement.

👍

PR #2796, opened 2023-08-09
feat: add api validation and OpenAPI doc generation using ts-rest

### LINKED ISSUE #2786: Add validation to API routes
Migrating to TS-Rest

### LINKED ISSUE #2788: Spike: decide on API routes to validate & validation strategy

tefkah:
## Issue(s) Resolved

#2788 
#2786 

## Docs

This PR adds validation and OpenAPI doc generation for API routes using `ts-rest` and `zod`.

### What it do

- Adds `zod` validation to API routes
- Declarative routing with `ts-rest` at `utils/api/contract.ts`
- Automagically generate a typed api client in `utils/api/client.ts`
- Create a full OpenAPI spec at `/spec.json` and a nice looking api documentation at `/apidocs` using Stoplight Elements.
- Does this for all the routes under
    - `/api/collections`
    - `/api/collectionAttributes`
    - `/api/collectionPubs`
    - `/api/export`
    - `/api/facets`
    - `/api/import`
    - `/api/members`
    - `/api/pages`
    - `/api/pubs`
    - `/api/pubAttributes`
    - `/api/pubEdges`
    - `/api/releases`
    - `/api/workerTasks`
    - `/api/login`
    - `/api/logout`
    - `/api/uploadPolicy`

These routes were chosen because they were either
a. Already used by `pubpub-client`
b. Had some recurring errors that could be solved with validation (e.g. `/api/pubEdges/:pubId` often crashed bc `pubId` was not a valid uuid)

> [!NOTE]
> Read the full documentation here

https://github.com/pubpub/pubpub/blob/fdab5bea0561f10de86b3dedbe2a9965c0f63641/utils/api/README.md?plain=1#L1-L426

<details>
<summary>Previous iteration</summary>

This PR adds validation middleware for select API routes (to be expanded to potentially all) using `zod`. 
It then uses that middleware to generate OpenAPI docs using Elements, available at `/apidocs` (could be anywhere of course).

The method used is taken from [this repo](https://github.com/omri-a/express-zod-openAPI-middleware/tree/master).

### How to use

First import the `validate` middleware from `utils/api`
(Note: this might not be the best name to use, as there's also a `validate` function in `utils/validation`. Open to suggestions!)

```ts
import { validate } from 'utils/api'
```

Then import `zod` and this thing that extends `zod` so you can add more properties to the schema that will be used for OpenAPI docs.

```ts
import { z } from 'zod'
import { extendZodWithOpenApi } from '@anatine/zod-openapi';

extendZodWithOpenApi(z)
```

Then, stick it in your route definition like so:

```ts
app.post('api/pubs',
    validate({
        summary: 'Create a new pub', // acts like the 'title' in the OpenAPI docs
        description: 'This route allows you to create a new pub',
        // true by default, indicates whether you need a cookie to access this route
        // does not actually do any validation on the cookie, this is just for docs
        // although this could be modified to do so if we wanted
        security: true, 
        // the allowed body schema, using zod
        body: z.object({
            // .uuid() will check if the string is a uuid, which solves some issues
            // sequelize we were having
            collectionId: z.string().uuid().optional(),
            communityId: z.string().uuid(),
            // you can add extra documentation using the .openapi method
                          .openapi({ description: 'The community to create the pub in' }),
        })
        // you can stick the openapi method almost anywhere!
        .openapi({ title: 'This will show up as the title of the body' }),
        statusCodes: {
            201: pubSchema, // complex schemas can be defined elsewhere and imported
            // 209: other statuscodes can also be documented here!
        }
        // response: pubSchema // you can also use the `response` property, but that will result in a 200 status code in the api docs
    }),
    wrap(async (req, res) => {
        // do stuff
        const pub = await createPub(req.body)
                                    //  ^? { collectionId?: string, communityId: string } // its got types!
        
        return res.status(201).json(pub) // <- also gets typechecked against the response schema you defined!
        // the response does not get parsed, but could be added if desired
      }
    ))
    ```
)
```

That's basically it!

### Some suggestions

#### ♻️ Reusing schemas

It's probably best to first define the base schema for an object, like a pub, and then derive the put/post/delete schemas from that. That way if you add some documentation to the main schema (or have some complexish validation), it will cascade down to the others.

Example:
```ts
const pubSchema = z.object({
    id: z.string().uuid(),
	slug: z
		.string({
			description: 'Slug',
		})
		.regex(/^[a-zA-Z0-9-]+$/)
		.min(1)
		.max(280)
		.openapi({
			uniqueItems: true,
			example: 'some-slug',
		}),
	title: z.string().openapi({
		example: 'A beautiful title',
	}),
    // etc
})

const pubPutSchema = pubSchema.omit({
    id: true
  }).partial()
```

#### ❓ Validating other things

You can also validate query and path params using this method. Header validation could also be added, but AFAIK that's not really used in our API.

```ts
app.get(
	'/api/pubEdges/:id',
	validate({
		tags: ['PubEdges'],
		description: 'Get a pubEdge by id',
		security: false,
		params: {
			id: z.string().uuid(), // only allow uuid params
		},
        query: z.object({
            collectionId: z.string().uuid().optional() // make sure to add optional, otherwise you must include the query param!
        }), 
		response: pubEdgeSchema,
	}),
	wrap(async (req, res) => {
        const id = req.params.id 
                           // ^? string
        const collectionId = req.query.collectionId
                                     // ^? string | undefined    
        //
    })
)
```

### How does it work?

It does something roughly like:
1. Create a middleware function that does validation and allows you to add some OpenAPI related documentation.
2. Set some properties on that function
3. Start app
4. Backtrace through the app to find all the routes with such middleware
5. Loop over those properties you defined and generate OpenAPI docs

Quite clever! This gives us both the ease of defining the validation as middleware compatible with our existing routes, but is still able to collate them all into a single OpenAPI doc.

#### 👍 Good things

Allows us to easily slot validation/documentation into our existing route definitions, something which I could not find a way to do with existing libraries.

#### 👎 Bad things

1. Requires us to start the app in order to generate the OpenAPI docs.
2. (see further suggestions) Probably not possible to (declaratively) generate a typesafe client to use in PubPub itself from this. Could be done from the OpenAPI docs, but with 1. that would require us to start the app first, plus would end up duplicating a lot of types.
3. Custom library code :(

### Next steps

Major next step is to bring this information to `pubpub-client`.

I think there's two ways to do that

#### 1. Leverage the OpenAPI docs to generate (parts of) the client using some existing library (surely one out there).

Would probably work! Might end up duplicating the types, but I think it's better than say exporting a library of types from `v6`, as that would probably be very large.

#### 2. Create a more declarative route definition that can be used to generate both the routes and a typesafe client.

I'm very interested in this approach (very inspired by `ts-rest`), and created an example (actually already works, but only has two routes implemented) in this sub-pr (not sure if that's a word) that shows how it could work! 

#2822 

### Other steps

- [ ] Add more validation to routes
- [ ] Test validation more
- [ ] Add more documentation to docs
- [ ] Add some more `GET` routes for the client, such as `collections` and `firebaseToken` to make the client rely less on the nasty hacks it does now.

</details>


deposit/resource.ts
```diff
@@ -7,16 +9,19 @@ export type ResourceLicense = {
...
+	'Journal',
+	'JournalIssue',
+	'JournalArticle',
+	'Conference',
+	'ConferenceProceeding',
+	'ConferencePaper',
+	'Other',
+] as const;
+
+export type ResourceKind = (typeof resourceKinds)[number];
```
tefkah, line 24: ive done this a number of times in order to reuse this enum in a schema somewhere

deposit/resource.ts
```diff
@@ -152,3 +177,86 @@ export const getIdentifier = (resource: AnyResource, identifierKind: ResourceIde
 export const getIdentifierValue = (resource: AnyResource, identifierKind: ResourceIdentifierKind) =>
 	resource.identifiers.find((identifier) => identifier.identifierKind === identifierKind)
 		?.identifierValue;
+
+export const partialResourceSchema = z.object({
```
tefkah, line 181: i made this before i had the `/utils/api/schemas/` convention, maybe should be moved there.

Also, might make more sense to define the types based on the schema now, but its not that important

3mcd (approved):
I've done a good bit of testing locally. Found a few issues but @tefkah patched them up. We should go ahead and get this in dev and do some additional testing on duqduq!

PR #2798, opened 2023-08-14
Prevent downloads when suggestions exist

### LINKED ISSUE #2734: Reject suggestions before converting pub
After a user clicks _Download_ and selects a different format, automatically reject all suggestions in a pub (if it has any) before sending the contents to a worker for conversion.

kalilsn:
## Issue(s) Resolved
Resolves #2755
Resolves #2734

## Test Plan
On the review instance:
1. Create a pub or visit an existing one
2. Enter suggestion mode and add suggestions
3. Attempt to download the pub
4. Resolve suggestions
5. Download the pub and confirm the download works
6. Create a release
7. Confirm the pub can be downloaded from the release as well

Then visit a community without the SE flag enabled and confirm that downloads work

image

### Notes/Context/Gotchas

@gabestein could use your approval/advice on the wording and the button names


gabestein (approved):
Functionality only: LGTM!

kalilsn:
Since I had a fix a merge conflict anyways, I rebased this PR and updated it to have the buttons in the right order and make the pub release dialog text consistent.

howebenhowe (approved):
Apart from the eslint issue you pointed out, this looks great.
I dunno if it's a real problem. The squiggles are yellow rather than red. From your link to documentation, it sounds like it can task the DOM with more work, but we haven't been seeing sluggish rerenders, so I don't feel worried about it myself.

kalilsn:
>Apart from the eslint issue you pointed out, this looks great.
I dunno if it's a real problem. The squiggles are yellow rather than red. From your link to documentation, it sounds like it can task the DOM with more work, but we haven't been seeing sluggish rerenders, so I don't feel worried about it myself.

Makes sense! And it's not like there's tons of children for this component, the impact is probably small.

PR #2799, opened 2023-08-14
Ignore transactions inside of math nodes

### LINKED ISSUE #2689: See if it's possible to disable math node suggestions
### What
Math nodes have some issues in SE mode, let's see if it's easy to disable suggestions in them for now. Half-day spike, we'll just release  with a caveat in the tooltip otherwise

kalilsn:
## Issue(s) Resolved
Resolves #2689 
## Test Plan
On the review instance
1. Create a math node
2. Turn on suggestion mode
3. Edit the math node. The edits should appear directly, not as suggestions


gabestein (approved):
Functionality only: LGTM!

howebenhowe (approved):
code looks good

PR #2814, opened 2023-08-18
Allow discussions of suggestions by combining the inline menus

kalilsn:
Here's a draft of combining the menus so users can add discussions on suggestions. It technically works at the moment, but as you can see in the screenshots there are a few issues:

- Highlight colors override suggestion styling
- Reanchoring doesn't seem to have any effect
- Discussions can end up in pretty weird locations

New menu style:
image

Adding a discussion on a suggestion:
image

Viewing a discussion on a suggestion:
image


gabestein:
Thanks for checking this out!

The discussion highlighting I think is fine? It only shows when you hover or open the discussion, which is more or less what I would expect, I think.

Discussion placement is indeed weird, but I'm having trouble replicating in my own test. Can you link to the doc (or try to create) one with that issue?

kalilsn:
@gabestein https://pubpub-pipel-kalilsn-di-dd4ikx.herokuapp.com/pub/jz6c9av3/draft has these weird comment placements. I wonder if your comment didn't cause them because it was in a shorter doc?

kalilsn:
>The discussion highlighting I think is fine? It only shows when you hover or open the discussion, which is more or less what I would expect, I think.

Coming back to this with fresh eyes, I do think the highlighting makes more sense than I did on Thursday. But it feels kind of like clicking into a form input where the label is in the placeholder text and disappears when you focus the input - I find myself losing track of what the suggestion is once I can't see the color. Doesn't seem like a huge issue as you said, and I think the best fix is ultimately generating descriptions of the suggestions instead of relying on colors and icons.

gabestein:
> Doesn't seem like a huge issue as you said, and I think the best fix is ultimately generating descriptions of the suggestions instead of relying on colors and icons

Agreed. That and probably auto-creating discussions from those descriptions.

> @gabestein https://pubpub-pipel-kalilsn-di-dd4ikx.herokuapp.com/pub/jz6c9av3/draft has these weird comment placements. I wonder if your comment didn't cause them because it was in a shorter doc?

Ah, yup. That is quite strange. Unless you think it's a quick tweak to the placement alg, let's postpone until after September and then get it in?

PR #2816, opened 2023-08-21
fix: throw not found error on /dash/.../impact for invalid collections/pubs instead of crashing

tefkah:
Fixes errors like these: https://kfg.sentry.io/issues/4405829345/?project=1505439&query=is%3Aunresolved+cannot+read+properties+of+null&referrer=issue-stream&statsPeriod=14d&stream_index=0

## Test Plan
1. Go to random community
2. Go to /dash/collection/some-name/impact
3. See if you get "Not found"  instead of Application Error
4. Go to /dash/pub/some-name/impact
5. See if you get "Not found" instead of Application Error

Ideally also check for
/dash/.../some-name/settings
.../members
.../facets
.../submissions
.../reviews
.../activity

### Before
image

### After
image

### DIFF
--- a/server/routes/dashboardPage.tsx
+++ b/server/routes/dashboardPage.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import Html from 'server/Html';
 import app from 'server/server';
-import { handleErrors, ForbiddenError } from 'server/utils/errors';
+import { handleErrors, ForbiddenError, NotFoundError } from 'server/utils/errors';
 import { getInitialData } from 'server/utils/initData';
 import { hostIsValid } from 'server/utils/routes';
 import { generateMetaComponents, renderToNodeStream } from 'server/utils/ssr';
@@ -15,10 +15,7 @@ app.get(['/dash/pages/:subMode'], async (req, res, next) => {
 		}
 		const initialData = await getInitialData(req, { isDashboard: true });
 
-		if (
-			!initialData.scopeData.activePermissions.canView ||
-			!initialData.scopeData.elements.activeTarget
-		) {
+		if (!initialData.scopeData.activePermissions.canView) {
 			throw new ForbiddenError();
 		}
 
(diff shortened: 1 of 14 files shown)


gabestein:
Mind adding for these routes, too:
`/settings`
`/members`
`/facets`
`/submissions`
`/reviews`
`/activity`

server/routes/dashboardPage.tsx
```diff
@@ -15,7 +15,10 @@ app.get(['/dash/pages/:subMode'], async (req, res, next) => {
 		}
 		const initialData = await getInitialData(req, { isDashboard: true });
 
-		if (!initialData.scopeData.activePermissions.canView) {
+		if (
+			!initialData.scopeData.activePermissions.canView ||
+			!initialData.scopeData.elements.activeTarget
+		) {
```
kalilsn, line 21: I think this should be `NotFoundError` like the others to avoid leaking whether a page exists to a logged out (or otherwise unauthorized) user.

Also I'm still seeing a crash at [/dash/pages/some-bad-name](https://pubpub-pipel-fix-collec-sudbuf.herokuapp.com/dash/pages/some-bad-name) but only while logged in.

kalilsn:
I don't actually see crashes for the /dash/pub/some-name/impact (or settings, members etc.) routes on duqduq or prod right now. Was that just caution or is there a different route I should look at?

The collections fixes look good too!

server/routes/dashboardPage.tsx
tefkah: Agree, that should be separated!

Also that was not properly catching the error (a page is never a target/scope, the activeTarget here is the community so it would pass), but should be fixed now!

tefkah:
> I don't actually see crashes for the /dash/pub/some-name/impact (or settings, members etc.) routes on duqduq or prod right now. Was that just caution or is there a different route I should look at?
> 
> The collections fixes look good too!

Hmm I thought I saw that at some point, but can't reproduce it either! Sorry for making you do double checks!

kalilsn:
Looks good but there's now an unused import which I will fix for you!

kalilsn (approved):

PR #2817, opened 2023-08-22
fix: bulk import inlineFIle and downloads

gabestein:
Makes two fixes to bulk imports to account for some db and prosemirror schema drifts over time: downloads, which require a JSON wrapper, and file cards, which require some attribute data.

## Test Plan
It's complicated to test, but ping Gabe for an example directive and directory if you'd like to.

### DIFF
--- a/workers/tasks/import/bulk/directives/pub.ts
+++ b/workers/tasks/import/bulk/directives/pub.ts
@@ -157,9 +157,12 @@ const resolveDirectiveValues = async (directive, sourceFiles, rawMetadata, doc):
 			if (key === 'downloads') {
 				resolvedDirective[key] = [{ url: resolvedValue, type: 'formatted' }];
 			} else if (key === 'inlineFile') {
+				const { $sourceFile } = value as any;
+				const resolvedFileName =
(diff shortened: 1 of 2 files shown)


workers/tasks/import/bulk/directives/pub.ts
```diff
@@ -154,7 +154,26 @@ const resolveDirectiveValues = async (directive, sourceFiles, rawMetadata, doc):
...
+			} else if (key === 'inlineFile') {
+				const { $sourceFile } = value as any;
+				const resolvedFileName =
+					`${rawMetadata.title}${resolvedValue
+						.split('/')
+						[resolvedValue.split('/').length - 1].substring(
+							resolvedValue.indexOf('.'),
+						)}` ||
+					$sourceFile ||
+					resolvedValue.split('/')[resolvedValue.split('/').length - 1];
```
3mcd, line 168: Could you pull `resolvedValue.split('/').length - 1` into a const above? Would make this a lot more readable I think

3mcd (changes requested):

3mcd (approved):

PR #2827, opened 2023-08-29
fix: move workflow target emails to cc; add Reply-To for first target address

3mcd:
## Issue(s) Resolved

Implements expected behavior addressed by #2776.

## Test Plan

1. Enable submissions for a collection.
2. Add a personal email address to the list of addresses to be CC'd when a Pub is submitted (target email addresses).
3. Submit a Pub to the collection.
4. Confirm the submitter's email address recieves a message with all target email addresses CC'd.
5. Reply to the email from the submitter's email address. The first target email address should receive this email.


gabestein (approved):
Sorry this sat for so long. Approving and will merge!

PR #2830, opened 2023-08-30
Prevent archiving empty discussions + sentry/sourcemap fixes

kalilsn:
## Issue(s) Resolved
This PR fixes a minor bug where a user could try to archive a discussion before they'd actually posted it, which caused an error. That button is now hidden until the comment is posted.

Since I was using that error to test out some monitoring improvements and the fix was small, I've combined some of the sentry changes into this PR:
- We now upload sourcemaps to sentry for the frontend and backend on each build
- The assert function was a no-op in prod, although it would still narrow the type, which made it easy to hide potential errors while using it! That's still true, but now we'll capture those failures in sentry.
- I removed the slow request warning since it's mostly noise at this point (and eating up our events quota)
- I made sure review instances get the commit hash (they don't have access to the heroku provided env var at run time, only build time) so they can set a release too

The only change in here that seems controversial/risky is that I've made our `wrap` function use async/await instead of `Promise.resolve` for compatibility with the node/v8 async tracing, [which only works with async/await and Promise.all](https://v8.dev/docs/stack-trace-api#async-stack-traces). Since this hasn't actually resolved the issue (stack traces still end at `process.processTicksAndRejections (node:internal/process/task_queues:95:5)`), I don't mind reverting this part, but it does seem like a step in the right direction.

## Test Plan
- Create a new discussion and observe that you cannot archive it until posting
- Make sure you can still archive it after comments are posted


client/webpack/webpackConfig.dev.js
```diff
@@ -102,6 +103,8 @@ module.exports = {
 		new ManifestPlugin({
 			publicPath: '/dist/',
 		}),
+		// Allow shared utils to import the sentry/node package by replacing it in the webpack build
+		new webpack.NormalModuleReplacementPlugin(/@sentry\/node/, '@sentry/react'),
```
kalilsn, line 107: Without this, the webpack build fails when it tries to import the node sentry sdk rather than the react one in the new assert implementation. Since they have the same API, this works fine even if it feels like a hack!

3mcd (approved):

PR #2843, opened 2023-09-18
Allow the SDK to use PubPub as a submodule

### LINKED ISSUE #2828: Use PubPub as sub-module for SDK

tefkah:
Terminology
- SDK: the thing at https://github.com/tefkah/pubpub-client
- "the client": the thing generated by `initClient` in `utils/api/client.ts` 
___

This PR allows PubPub v6 to be used as a submodule in the SDK, see this PR for in depth details on how that works: https://github.com/tefkah/pubpub-client/pull/7

Relevant for the current PR, there are a couple of things that needed to be changed on v6's side in order to accommodate this. Most of these changes where necessary in order to spin up a version of v6 to be used in the tests of the sdk, that way the sdk

## 🔥 Replaced the `webpackConfig.dev.js` loader from `ts-loader` to `esbuild-loader`, leading to `8x` faster compile times

In order to run the tests properly, the client needs to be compiled. Current compilation is really slow, taking 45 seconds on my machine. 

To speed this up, I swapped out the transpiler from `ts-loader` to `esbuild-loader`, which as the name might suggest uses `esbuild` instead of `tsc` to load the typescript/js files, which is significantly faster.

This, surprisingly, worked almost immediately! The only changes I had to make were to re-export typescript types, like here:

https://github.com/pubpub/pubpub/blob/c7c8c7033224896523c786bc042f639b657a371b/client/components/Editor/plugins/code/index.ts#L29-L39

`esbuild` does not do typechecking, so unless explicitly told to ignore these by adding `type` in front, it will assume these are javascript functions/objects/etc., and will complain (warn) that it cannot find these.
Simply adding the `type` annotation in front solves this.

### Downsides

There is one downside about this approach, namely that `esbuild` does not do any typechecking.

I do not think this is a problem, because the whole codebase should be typechecked anyway before publishing. That's why I instead included `tsc --noEmit` at the end of `npm run lint`, which always gets run in CI. This way we know for sure that the types are alright, and still save a lot of time locally developing.

Also, I did not change this in the prod config anyway, just to be safe!

## Moving some stuff around/changing imports

In order to have imports resolve correctly when running the app from a different directory and to avoid importing things incredibly deeply in the SDK, I moved some things around a bit. 

These fall into X categories:

1. Moving inline types from e.g. `client` to `types`
2. Copied some enums (which were actual JS, not types) from `types` to the schema where I need them. This allows `tsup` to kind of ignore `types` for bundling the code, which lessens the possibility of accidentally including a lot v6 code in the SDK.
3. Changed some explicit `dist` imports. The were like `require(process.cwd() + '/dist')`. If you run the app from the sdk (which is one dir up), this would obviously not work, so I changed it to something like `require(__dirname + '/dist')`, so it's relative to the file instead. This does not always work, because sometimes the files are themselves (assumed to be) in `dist`, so if that was the case I did something not very pretty, like

https://github.com/pubpub/pubpub/blob/c7c8c7033224896523c786bc042f639b657a371b/server/Html.tsx#L7-L11

## Changed all `sequelize-typescript` class definitions to use `declare` for the fields instead of just putting them in there.
See e.g. here: https://github.com/pubpub/pubpub/blob/c7c8c7033224896523c786bc042f639b657a371b/server/collection/model.ts#L48-L65

To be quite honest I'm not entirely sure why this was needed. When running tests in v6, everything is fine, but when running them through the submodule in the sdk, I get errors like these:
```log
  console.warn
      (sequelize) Warning: Model "Pub" is declaring public class fields for attribute(s): "id", "slug", "title".
      These class fields are shadowing Sequelize's attribute getters & setters.
      See https://sequelize.org/main/manual/model-basics.html#caveat-with-public-class-fields
```

This indicates that sequelize is overwriting already declared fields somehow. This not only leads to these warnings, but to the whole app not working.

Changing all fields to use `declare` instead does work, and does not seem to negatively impact the core app, so that's why I just ran with it. This might be because I am running the tests in the `sdk` with `swc` to speed them up, idk.

I do not think this will pose a problem in the future, e.g. when someone would forget to add `declare` somewhere. As stated, the core app works fine without it, it might break a test in the sdk maybe, but the `sdk` does not need the entire app to be able to run to work, only to test it.
  

## Issue(s) Resolved

#2828 
#2851 

### To test SDK

1. Pull this PR: https://github.com/tefkah/pubpub-client/pull/7 with `--recurse-submodules`
2. `pnpm i`
3. `pnpm run ci` to see the tests work.

### To test that the app has no regressions

I think a smoke test is in order. I'm quite confident nothing of note has changed, but it's good to check.


tsconfig.json
```diff
@@ -65,10 +65,10 @@
...
 		} /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */,
 		// "rootDirs": [],                        /* List of root folders whose combined content represents the structure of the project at runtime. */
-		"typeRoots": [
-			"./types",
-			"node_modules/@types"
-		] /* List of folders to include type definitions from. */,
+		// "typeRoots": [
+		// 	"./types",
+		// 	"./node_modules/@types"
+		// ] /* List of folders to include type definitions from. */,
```
tefkah, line 71: So for some reason this was necessary to do, but I'm not entirely sure why.

utils/api/contracts/community.ts
```diff
@@ -0,0 +1,21 @@
...
+		path: '/api/community',
+		method: 'PUT',
+		summary: 'Update a community',
+		description: 'Update a community',
+		body: communityUpdateSchema,
+		responses: {
+			200: communitySchema,
+		},
+	},
+});
```
tefkah, line 21: I have only added updating communities here, not creating or deleting. I don't see the need to be able to do that through the API, seems like a possible abuse vector.
I would add the GET endpoint here when I finish that.

utils/api/contracts/customScript.ts
```diff
@@ -0,0 +1,22 @@
...
+		method: 'POST',
+		summary: 'Set a custom script',
+		description:
+			'Set a custom scripts, e.g. the CSS or JS (if you have access) for this community',
+		body: createCustomScriptSchema,
+		responses: {
+			200: customScriptSchema,
+		},
+	},
+});
```
tefkah, line 22: This allows users to update scripts, for most only the CSS.
In the client, I've only explicitly added a method to update the CSS. People who have JS scripts activated can probably figure out how to use the raw client (e.g. `PubPub.client.customScript.set`, instead of `PubPub.community.css`)

utils/api/schemas/collection.ts
```diff
@@ -26,9 +27,9 @@ export const collectionSchema = z.object({
 	viewHash: z.string().nullable(),
 	editHash: z.string().nullable(),
 	metadata: z.record(z.any()).nullable(),
-	kind: z.enum(types.collectionKinds).nullable(),
+	kind: z.enum(collectionKinds).nullable(),
 	doi: z.string().nullable(),
-	readNextPreviewSize: z.enum(types.readNextPreviewSizes).default('choose-best'),
+	readNextPreviewSize: z.enum(readNextPreviewSizes).default('choose-best'),
```
tefkah, line 32: Using non-types from `types` really increases the bundlesize, better to manually specify them.

utils/api/schemas/collection.ts
```diff
@@ -73,6 +74,7 @@ export const collectionUpdateSchema = collectionSchema
 		layout: true,
 		layoutAllowsDuplicatePubs: true,
 		avatar: true,
+		doi: true,
```
tefkah, line 77: It wasn't possible to update the doi before, whoops.

utils/api/server.ts
```diff
@@ -19,7 +19,11 @@ import { contract } from './contract';
 
 const s = initServer();
 
-export const server = s.router(contract, {
+// FIXME: Add these routes back in
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const { customScript, community, ...contractWithScriptsAndCommunity } = contract;
+
+export const server = s.router(contractWithScriptsAndCommunity, {
```
tefkah, line 26: Note: I did not implement any server routes for the scripts or contract, that would warrant a lot more testing. As such, these routes do not have validation.

I will convert these routes to ts-rest routes when I do #2850

package.json
```diff
@@ -77,16 +79,16 @@
 		"@lezer/python": "^1.1.1",
 		"@lezer/rust": "^1.0.0",
 		"@lezer/xml": "^1.0.0",
-		"@monaco-editor/react": "^4.1.1",
+		"@monaco-editor/react": "4.1.1",
```
tefkah, line 82: This is really silly, but npm install this as 4.1.1, pnpm as 4.5.0, which leads to some bugs with the types. I removed the caret to make it more consistent.

package.json
```diff
@@ -14,16 +15,17 @@
 		"generate-sequelize-typediff": "npx tsx tools/toSequelizeTypescript/type-differ.ts",
 		"api-dev": "concurrently \"npm run watch-server\" \"NODE_PATH=./dist/server/client:./dist/server:./dist node --enable-source-maps init.js --watch ./dist/server --watch ./node_modules/@pubpub\"",
 		"api-prod": "NODE_PATH=./dist/server/client:./dist/server:./dist node --enable-source-maps init.js",
+		"build-client-dev": "webpack --config ./client/webpack/webpackConfig.dev.js",
```
tefkah, line 18: This is the new fast way to build the cilent using esbuild-loader. I separated it out as its own command to be able to use it in the sdk, as the client needs to be compiled in order for the app to work.

server/collectionPub/api.ts
```diff
@@ -57,7 +57,8 @@ export const collectionPubServer = s.router(contract.collectionPub, {
 		});
 		return {
 			status: 201,
-			body: collectionPub,
+			// FIXME: Do some checking of whether the contextHint is valid
+			body: collectionPub as any,
```
tefkah, line 61: I tried to make collectionPubs and Collections more strict, by matching the types of contextHints and shape of the metadata with the Kind of the Collection, e.g. Issue collections can't have ISBNs, its pubs can only have Article contextHints etc. 
This turned out to be quite difficult type wise, especially because collection kind information is not available when updating collectionPubs or Collections (you cannot change a collection's kind afaik, or should not be able to at least).

While I do think it would be good to not allow people to e.g. set ISBNs on Issues, this would require some rewriting of the routes (adding extra checks for kinds). I might try to add this to the routes PR, but atm it's not very important.

3mcd (approved):
LGTM. Once we get this merged I'll do some additional testing in duqduq.

PR #2844, opened 2023-09-20
hotfix: Add cache cookie and cachebuster to user signup

gabestein:
## Issue(s) Resolved
Previously, newly created users were not receiving the cache cookie but were being logged in. When this coincided with a cache refresh, it meant logged in pages were sometimes cached and displayed to other users. This fixes that problem by adding the cache cookie on user signup and a cachebreaker to the redirect.

## Test Plan
1. Create new user locally.
2. Make sure the request to the users endpoint returns a `pp-no-cache` cookie.
3. Make sure the redirect request contains `pp-no-cache`
4. Repeat on duqduq, making sure that after new user creation, accessing duqduq from incognito doesn't load as if you are that new user.

### DIFF
--- a/client/containers/UserCreate/UserCreate.tsx
+++ b/client/containers/UserCreate/UserCreate.tsx
@@ -69,7 +69,8 @@ const UserCreate = (props: Props) => {
 			}),
 		})
 			.then(() => {
-				window.location.href = '/';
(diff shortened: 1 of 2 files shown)


3mcd (approved):
Tested, works great!

PR #2849, opened 2023-09-25
fix: correct validation for customPublishedAs and collection-pages blocks

tefkah:
- fix: properly coerce custompublishedat in zod schema
- fix: allow full pages to be sent as collections-pages items
- fix: allow doi previews/deposits to be null/empty objects


3mcd (approved):
LGTM

PR #2855, opened 2023-10-02
fix: allow null backgroundImage and 'none' button

tefkah:
## Test Plan

1. Go to HDSR
2. Open network tab
3. Try to modify homepage
4. Should not cause a problem.

### Notes/Context/Gotchas
Two issues

1. banner button with no text is "none", not `undefined`
2. allow backgroundImage to be `null`


gabestein (approved):
LGTM! Thanks!

PR #2866, opened 2023-10-09
Hotfix: allow multiple custom labels for the same reference in PDF exports

gabestein:
## Issue(s) Resolved

Currently, if you have two of the same references in a document with two different custom inline citation labels, and export to PDF/HTML, it will always export the PDF/HTML so that every inline citation label for that reference is the last custom label, rather than using the custom label for each one. 

The reason is that during the process of creating the HTML document which gets turned into a PDF, we are giving every inline citation with the same "fingerprint" -- e.g. the same reference -- the same ID, which when run through prosemirror-reactive, results in every node with that ID being replaced with the last version of the node. The result is a PDF/HTML export that deviates from the intent in the pub. Minimal example:

**Pub with multiple citations of the same reference with different custom labels**
Screenshot 2023-10-09 at 16 18 48

**Exported PDF version**
Screenshot 2023-10-09 at 16 19 14

This PR fixes that by adding the original prosemirror node ID to the ID generated during the HTML/PDF conversion process. This means that if there are multiple citations in HTML/PDF, the "return" link in the footer reference list will always only go to the first such citation. Ideally, we'd probably list multiple return arrows? Maybe? But there's actually not a great UX solution for that, and this hotfix seems acceptable for now.

## Test Plan
1. Export a Pub with multiple citations with custom labels for the same reference to PDF/HTML. I have two examples below.
2. Make sure all the custom references show as expected in the PDF, e.g., match the Pub.
3. Make generated footer list is not duplicated (note: maximal example has two footer lists due to HDSR replicating footer list in text).

Minimal example (HDSR local/duqduq community): http://localhost:9876/pub/tlaq5hmc/draft

Maximal example (HDSR local/duqduq community): http://localhost:9876/pub/cuqeguoe/draft

### DIFF
--- a/workers/tasks/export/html.tsx
+++ b/workers/tasks/export/html.tsx
@@ -94,7 +94,7 @@ const addAttrsToNodes = (newAttrs, matchNodeTypes, nodes) =>
 const getCitationLinkage = (unstructuredValue, structuredValue, nodeId = null) => {
 	const digest = digestCitation(unstructuredValue, structuredValue);
 	return {
-		inlineElementId: `citation-${digest}${nodeId && '-' + nodeId}-inline`,
+		inlineElementId: `citation-${digest}${nodeId ? '-' + nodeId : ''}-inline`,
 		bottomElementId: `citation-${digest}-bottom`,
 	};
 };
--- a/workers/tasks/export/html.tsx
+++ b/workers/tasks/export/html.tsx
@@ -91,10 +91,10 @@ const addAttrsToNodes = (newAttrs, matchNodeTypes, nodes) =>
 		return node;
 	});
 
-const getCitationLinkage = (unstructuredValue, structuredValue) => {
+const getCitationLinkage = (unstructuredValue, structuredValue, nodeId = null) => {
 	const digest = digestCitation(unstructuredValue, structuredValue);
 	return {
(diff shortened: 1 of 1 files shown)


workers/tasks/export/html.tsx
```diff
@@ -91,10 +91,10 @@ const addAttrsToNodes = (newAttrs, matchNodeTypes, nodes) =>
 		return node;
 	});
 
-const getCitationLinkage = (unstructuredValue, structuredValue) => {
+const getCitationLinkage = (unstructuredValue, structuredValue, nodeId = null) => {
 	const digest = digestCitation(unstructuredValue, structuredValue);
 	return {
-		inlineElementId: `citation-${digest}-inline`,
+		inlineElementId: `citation-${digest}${nodeId && '-' + nodeId}-inline`,
```
kalilsn, line 97:
```suggestion
		inlineElementId: `citation-${digest}${nodeId ? '-' + nodeId : ''}-inline`,
```

I can't tell if this is just a cosmetic fix for links or if it's an issue but this version will avoid printing `'null'`

gabestein: No, this is needed -- there are some cases where we may not want to pass in the ID.

kalilsn (approved):

PR #2874, opened 2023-10-16
feat: Add /api/upload route

### LINKED ISSUE #2850: Implement v6 API Routes
- Get routes
- Upload routes
- Allow post routes to take all variables
- Firebase routes

tefkah:
## Issue(s) Resolved

#2850 

## Test Plan

- run `npm run test-file-upload`

## Docs

This PR implements an `/api/upload` endpoint, which takes a single `File` as an argument and streams it to our S3 bucket.

This is implemented using `busboy` to stream the formdata, and `@aws-sdk/lib-storage` to stream the file included in the form to S3 in multiple parts.

#### Community admin only

https://github.com/pubpub/pubpub/blob/5a1f9ae7abb1df6f68cabfc42bfccfc483792f9a/server/upload/api.ts#L112

#### Mime/filename validation

https://github.com/pubpub/pubpub/blob/5a1f9ae7abb1df6f68cabfc42bfccfc483792f9a/server/upload/api.ts#L120-L126

In case you don't properly set the `filename` or `contentType` properly on the file upload, I do some logic to make sure they're set properly.

In particular, if you upload a form in node like this

```ts
const form = new FormData();

const pdf = fs.readFileSync('pdf.pdf');

form.append('file', new Blob([image], { type: 'application/pdf' }));

```
using our current naming scheme the pdf will be named some random hash without an extension.
Because of this, if you download the file on e.g. MacOS, it will not recognize it as a PDF, and you won't be able to open it until you rename it.

By requiring either the filename or mimeType to be set, we can determine the one from the other.

#### Streaming to S3

I implemented a splitFileUpload method to the existing S3 client.

https://github.com/pubpub/pubpub/blob/5a1f9ae7abb1df6f68cabfc42bfccfc483792f9a/server/upload/api.ts#L129-L138

The way I use it in the actual route defaults to uploading 3 chunks at the same time. 
This is a completely arbitrary number, and we'll have to look at memory usage to see whether this is sustainable, but it speeds up uploads considerably.

https://github.com/pubpub/pubpub/blob/5a1f9ae7abb1df6f68cabfc42bfccfc483792f9a/server/upload/api.ts#L129-L138

### Considerations

Maybe this should be a worker? I found workers kind of annoying personally as an external dev, but I'm not sure what the possible memory impact might be just running this on the same server.
Same with #2880


tefkah:
I once again cannot check Heroku to see what's wrong with the CI, could anyone check for me?

3mcd (approved):

PR #2868, opened 2023-10-10
Implement new GET routes

tefkah:
Blocked by #2828 

## Issue(s) Resolved
#2850 (partially) 

Only implements the GET routes for Communities and Collections, to keep the PR manageable.

Is blocked by #2828, and kind of dependent on it.

PR #2899, opened 2023-11-06
fix: remove permission check for get user endpoint

### LINKED ISSUE #2861: User API suggestion route returning 403 even when user is logged in in some cases
1. Visit a pub with suggested edits as a non-superadmin, but admin of the community
2. Hover over suggestion
3. Name of user who made suggestion doesn't appear
4. Console network tab returns 403 response from `/api/user?suggestion...`

User who made suggestion's name shows.

tefkah:
## Issue(s) Resolved
#2861 

## Test Plan
Modified existing test to work properly.

Otherwise:
1. Go to Draft with suggested edits enabled (to be disclosed URL) as a non-super admin user
2. Create some suggested edits
3. Hover over them checkmark
4. Should say "Approve edit suggested by <name>" instead of just "Approve edit"

## Notes/Context/Gotchas

This was caused by the permissions never returning `{ read: true}` except for superadmins, not even for looking at your own user information, because that data was never passed to the permissions.

The "user you are looking at" is supposed to be extracted by this `getRequestIdsFunction`, the `submittedUserId`
https://github.com/pubpub/pubpub/blob/bbf520c08d7e7ccc2908b7d38bc05dd35bb8b9b7/server/user/api.ts#L10C1-L18C3

However, it extracts that from the `body`, which a `GET` request isn't supposed to have, so no `submittedUserId` is used during the `GET` request.

Even if it did though, you would only have permission to read if the suggested user is you, or you are a superadmin.
https://github.com/pubpub/pubpub/blob/bbf520c08d7e7ccc2908b7d38bc05dd35bb8b9b7/server/user/permissions.ts#L45-L51

Since this is not really the behavior that is desired from this endpoint (you would only be able to see your own name), I changed it (consulting @kalilsn and @3mcd) to _remove_ the permissions check here.
This way, anyone is able to request the suggestedEdit user data.

This was deemed acceptable as we only expose the name and avatar, information which isn't private to begin with on PubPub as anyone can search users and get this information. The only change is that this can now be done if you know the user's id.

Hope that's okay with you @gabestein !

### Changing the path

I also changed the API route structure to match the new [GET routes](#2868), so from

```
GET /api/users?suggestedUserId=<userId>
```
to

```
GET /api/users/<userId>
```

and updated all the uses of this route to match that.

Not a super important change given that this route is only used for the suggested edits, but stillo

### Why did the test work before

There was a test that was supposed to check whether a user is able to get another user's info, however it sent a `GET` request with a body of `{userId: <the current user>}`, which would pass the check, as the requesting user is allowed to request their own info.
However, that probably isn't how the API is intended to be used, I didn't even know you could send `GET` requests with a body!

I changed the test to fit the new API route structure and to not send the `userId` in the body.

### DIFF
--- a/client/containers/Pub/PubDocument/PubInlineSuggestedEdits.tsx
+++ b/client/containers/Pub/PubDocument/PubInlineSuggestedEdits.tsx
@@ -50,9 +50,7 @@ const PubInlineSuggestedEdits = () => {
 	const fetchSuggestedEditsUserInfo = useCallback(async () => {
 		if (selectedSuggestion) {
 			const suggestionUser: SuggestedEditsUser = await apiFetch.get(
-				`/api/users?suggestionUserId=${encodeURIComponent(
-					selectedSuggestion.suggestionUserId,
-				)}`,
+				`/api/users/${encodeURIComponent(selectedSuggestion.suggestionUserId)}`,
 			);
 			if (suggestionUser) {
 				setSuggestionAuthor(suggestionUser);
--- a/server/user/__tests__/api.test.ts
+++ b/server/user/__tests__/api.test.ts
@@ -46,7 +46,7 @@ describe('/api/users', () => {
 	it('allows an exisiting user to read another users profile info', async () => {
 		const { user, suggestionUser } = models;
 		const agent = await login(user);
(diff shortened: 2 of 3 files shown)


gabestein (approved):
Thanks! Looks good, decision makes sense.

PR #2904, opened 2023-11-09
fix(#2903): fix connections not being able to be updated

### LINKED ISSUE #2903: Can't save edited connections
1. Add a new connection to a pub
2. Edit the connection (type, direction, etc.) and attempt to save
3. Get Zod validation error

Pub connection edits should save.

Megan / Arcadia

tefkah:
## Issue(s) Resolved
#2903

## Test Plan
1. Create a pub
2. Add a connection
3. Try to update the connection

Also added a test to prevent regression

### Notes/Context/Gotchas

Cause was the connection schema not allowing for `null` values for the `targetPubId` or `externalPublication`, but the frontend sends that when updating.


gabestein (approved):
Thanks! Tested and works for me on local.

PR #2905, opened 2023-11-09
fix(#2902): rollback blueprintjs version to 3.26 to fix inability to edit media width through numeric input

### LINKED ISSUE #2902: Can't edit media width numbers
1. Add an image, for example to a Pub
2. Attempt to modify % width number.
3. Will not edit.

Screenshot 2023-11-08 at 14 46 01

Number should be editable.

Megan / Arcadia

tefkah:
## Issue(s) Resolved
#2902 

## Test Plan
1. Go to draft
2. Click on image
3. Try to manipulate the width using the numeric input (not the slider)
4. Should work

### Notes/Context/Gotchas
I stared at this for longer than I wanted to, could not figure out why it wasn't working. Just rewinded to an older version of blueprint (newer version happened because of my regenerating the lockfile in #2843), which seemed to restore the expected behavior.
Might be a regression in blueprint or something they changed, idk.


gabestein (approved):
Thanks for digging into this, and sorry it took so long. Yeah, Blueprint updates have always caused problems — we need to be really careful about making sure the versions stay consistent.

PR #2906, opened 2023-11-09
feat: new API routes

tefkah:
This PR adds GET routes for
- pubs
- pages
- member
- pubAttributions
- collectionAttributions

using the same method used in #2868.

It also incorporates the following PRs, which add routes which allow for imports and uploads. See them for more detail.

#2880 
#2874 

## Test Plan

Added a lot of tests, but routes are API only.

## Follow up
I would like to add some GET routes for discussions, but that would also involve moving creating schemas and `ts-rest` contracts for them. 

I think this would end up making the PR even larger than it already is, plus I need some time to understand how discussions are structured and which models need additional endpoints, so I think it's better suited for a follow up PR.

### Adding `batch` create to `CollectionAttributions`

`PubAttribution` has a `batch` create endpoint, but `CollectionAttribution` doesn't. Why?
I evened this out and also created one for `CollectionAttribution`

https://github.com/pubpub/pubpub/blob/c0a51e394ddbc87db0d05f1e2303e813d69e0f02/server/collectionAttribution/api.ts#L51-L86

I also added tests for this endpoint in both.

### Tests

I added tests for `CollectionAttribution` back in #2796, but not for `PubAttribution`. 
I basically just copied those over here:
https://github.com/pubpub/pubpub/blob/c0a51e394ddbc87db0d05f1e2303e813d69e0f02/server/pubAttribution/__tests__/pubAttribution.test.ts#L86-L210

### Changing 201s to 200 in `/api/pubAttributions`

The `PUT` and `DELETE` endpoints for `/api/pubAttributions` responded with `201 Created` instead of `200` like `/api/collectionAttributions` does.

I didn't see any place in the code where this is checked, so I just changed it to return a `200`.

### Removing custom error catcher around `/api/pubAttributions` CRUD routes

See here how it was:
https://github.com/pubpub/pubpub/blob/54ed278dc85bd527ddcd8ae12b8f8217f35bef65/server/pubAttribution/api.ts#L134-L152

This meant that everything was returned as 500 errors, which is a bit weird.

I just removed them, we don't have this anywhere else.

## Query Changes

3 relevant changes are made to the querying/filtering mechanism introduced in  #2868 , in addition to simply adding new endpoints for the above mentioned.

### 1. Ability to specify filters as their own query parameters, instead of being in `?filter`

Previously, to e.g. get all collections which are of kind `issue`, you had to do

```ts
const {body: issueCollections} = await fetch(`/api/collection?filter=${
	encodeURIComponent(
		JSON.stringify({
				kind: 'issue'
        }))
	}`).then(res=>res.json)
```

This is kind of tedious if you want to do quite simple filters.

Now, you are able to just specify `?kind=issue`

```ts
const {body: issueCollections} = await fetch(`/api/collection?kind=issue`).then(res=>res.json)
```

much cleaner!

This did however cause some collisions. `PubAttributions` and `CollectionAttributions` have an `order` field, which would conflict with the `?order` query parameter. To mitigate this, the `order` and `sort` query parameters have been renamed to `orderBy` and `sortBy`

### 2. Ability to use associations than `Community` to limit access by community

(Almost) all the new GET routes are only allowed to be accessed by the admin of the community the request is made to, or by superadmins.

For `Collection`, `Page`, `Pub` and `Member` it is quite easy to determine whether they belong to the Community the request is made to. We just do a `Pub.findAll({where: { communityId: community.id }})` to only return those `Pub`s in the current community.

For `Member`, `PubAttribution`, and `CollectionAttribution` this is not so easy, because they do not (always) have a `communityId` to scope the query to. 
Instead of giving up on access checking, I added the ability to add more complex querying functionality for those routes that need it. Luckily, this does not increase the number of queries that need to be made, although the resulting query is a bit ugly.

Let's see how this works for `Member`.

#### Example

A `Member` either belongs to a `Community`, a `Pub`, or a `Collection`, but never to more than one of them.

By passing an optional `customScope` key to the query implementation here
https://github.com/pubpub/pubpub/blob/f7993c2580416e54a84a619b14006b6005c8e790/server/member/api.ts#L45-L66

and the type here
https://github.com/pubpub/pubpub/blob/f7993c2580416e54a84a619b14006b6005c8e790/utils/query/queryMany.ts#L14-L42

This gets turned into an `Op.or` where clause + includes here
https://github.com/pubpub/pubpub/blob/f7993c2580416e54a84a619b14006b6005c8e790/utils/query/queryMany.ts#L44-L73

So in order to check whether e.g. the `pubId` that is mentioned on the `Member` belongs to a `Pub` that belongs to the current `Community`, a `where` clause like this is generated

```ts
{ 
   where: {
       [Op.or]: [
			{
			[Op.and]: [
				{
					pubId: {
						[Op.not]: null, // to check the pubId actually exists, otherwise the next part is unnecessary
					},
				},
				{
// this references the `pub` field. this requires us to `include` the `Pub` association, otherwise we get an error.
					[`$pub.communityId$`]: communityId, 
				},
			],
		}
  	}
}
```

which results in roughly the following SQL

```sql
SELECT ...
LEFT OUTER JOIN "Pubs" AS "pub" ON "Member"."pubId" = "pub"."id"
WHERE ("Member"."pubId" IS NOT NULL AND "pub"."communityId" = '<SOME-ID>')
```

Good to note is that for `Member`, this also requires a `LEFT OUTER JOIN` on `Communities` and `Collections`, so it's quite a heavy query. We might be able to speed it up if we added some indices, but it might be worth waiting to see some actual usage in sentry before making that decision.

### 3. Changed how admin access to the community is checked to be faster

Before, the "auth" flow was
1. find community based on hostname
2. check to see if there is Member with the userId of the current user with admin permissions
3. if not, check to see if the current user is superadmin

So for a `fail` we need to make 3 separate queries, which seems a bit wasteful.

Now, we do two checks.

1. If the user is superadmin (determined by looking at the `User` object that is included in the request, skipping the need for a query), just fetch the community and return that.
https://github.com/pubpub/pubpub/blob/f7993c2580416e54a84a619b14006b6005c8e790/utils/ensureUserIsCommunityAdmin.ts#L55-L57

Although this seems sketch, this User is determined based on the session cookie. This does not pose any additional security concerns, as we still lookup the User based on the session. If people were able to fake sessions, they would also succeed in getting around the previous implementation.

2. If not, run this query and return the community if we find one, otherwise throw a 403.
https://github.com/pubpub/pubpub/blob/f7993c2580416e54a84a619b14006b6005c8e790/utils/ensureUserIsCommunityAdmin.ts#L59-L86

This way we always only make 1 request for "auth", instead of 3.


3mcd (approved):
Looks great, thank you!

PR #2913, opened 2023-11-16
fix: allow facets to revert to higher scope value

tefkah:
## Issue(s) Resolved

Facets were not updating if you reverted to the higher facets setting.

## Test Plan

1. Go to a Pub
2. Change a facet setting (e.g. setting the labels for tables to "Strange Tables")
3. Save
4. Click "Revert to community/collection value"
5. Save
6. Reload
7. Change should be saved.


isTravis (approved):
Working for me on the PR deploy! Looks good. Thank you for the lightning fast fix!

PR #2936, opened 2023-12-06
fix: math duplication

### LINKED ISSUE #2925: Math nodes repeating in captions
1. Create a math node in a caption and update it
2. Close caption and re-open
3. Math will appear doubled

Expected: one maths only!

3mcd:
## Issue(s) Resolved

Resolves #2925

## Test Plan

1. Add a math node (display or inline) to a caption.
2. Close the formatting bar controls and click on the captioned element again.
3. The math node should look the same as before, and not contain duplicates!

### DIFF
--- a/client/components/Editor/schemas/math.tsx
+++ b/client/components/Editor/schemas/math.tsx
@@ -12,6 +12,7 @@ const inlineMathSchema = {
 	atom: true,
 	parseDOM: [
 		{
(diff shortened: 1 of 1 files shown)


tefkah:
Can confirm it works here and not on duqduq, so LGTM!

tefkah (approved):

PR #2941, opened 2023-12-11
feat: add purgeMiddleware for purging cache on CRUD operations 

### LINKED ISSUE #2924: Caching improvements
- Sub/domain-wide cache clearing calls to fastly when pub, page, settings, etc. updated.
- Sending logged in cache directives from server response instead of using cookies.

tefkah:
## Issue(s) Resolved

#2924

### DIFF
--- a/.jest/setup-env.js
+++ b/.jest/setup-env.js
@@ -25,6 +25,9 @@ process.env.ZOTERO_CLIENT_SECRET = 'def';
 process.env.FASTLY_PURGE_TOKEN_PROD = 'token';
 process.env.FASTLY_SERVICE_ID_PROD = 'prod';
 
+process.env.FASTLY_PURGE_TOKEN_DUQDUQ = 'token_duqduq';
+process.env.FASTLY_SERVICE_ID_DUQDUQ = 'duqduq';
+
 if (process.env.INTEGRATION) {
 	try {
 		require('../config.js');
--- a/.jest/setup-env.js
+++ b/.jest/setup-env.js
@@ -22,6 +22,9 @@ process.env.FIREBASE_TEST_DB_URL = 'http://localhost:9875?ns=pubpub-v6';
 process.env.ZOTERO_CLIENT_KEY = 'abc';
 process.env.ZOTERO_CLIENT_SECRET = 'def';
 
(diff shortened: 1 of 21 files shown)


gabestein:
This is awesome. After code review, I'd like to do some smoke testing on duqduq since I think it's really the only way to properly test it.

tefkah:
> This is brilliant. If I understand it correctly:
> 
> Anytime a post/put/delete request is sent, we purge the cache for that hostname? If so, I think that's going to satisfy most needs. A few edge cases to consider:
> 
> * What about CORS/HEAD/etc. requests? How will these be handled?
> * Is there a way to debounce these to some degree? I think we have 100,000 purge requests per hour, and I'm not all that concerned about hitting the limit, but it would be nice to have some thinking about how to handle it.
> * PDF/exports? These are kicked off on new releases and when you generate a format. I'm not sure if the links for these are updated via api request or via something on the server. I think the latter — they're just added to the database and the next query to the page returns the new URLs. But I'm not sure.
> * Global user profiles (e.g. [www.pubpub.org/user/](http://www.pubpub.org/user/)*). I don't think your Fastly mods will tag those pages with hostnames that apply, unfortunately. Those are also scraped relatively frequently and are sometimes quite large queries, so I think it would be best not to just, e.g., tag them all as user and purge them all anytime a post request is made.
> * Are there any API routes that apply to www? I think everything goes through the hostname, not directly to pubpub.org, but worth thinking through.
> * Purging www on new releases. Very low priority, since we can always purge manually when we make updates, which is infrequent. But are requests to www tagged with your fastly update so we can do that?

I think i addressed every one of these now, I can make a complete writeup tomorrow of how this works!

- exports now purge cache
- debouncing added
- tests added
- www.pubpub.org/user/* page tagging
- do not trigger on CORS/HEAD etc.

Through a tip of @3mcd  I don't need to patch the `res.json` function either, I can just listen to `res.on('finish')`, which is much cleaner.

3mcd (approved):
Looks great! Awesome work @tefkah

gabestein (approved):
Lgtm! Let's do some smoke testing on duqduq and get this launched!

PR #2948, opened 2023-12-14
fix: edge case where pubs with sibling edges of different domains would 500

tefkah:
- fix: include community in pub edge includes in for sibling edges to prevent weird connection 500 edge case
- fix: revert base hostname for pubdoc to be req.hostname as that has been proven to work

### DIFF
--- a/server/routes/pubDocument.tsx
+++ b/server/routes/pubDocument.tsx
@@ -209,7 +209,7 @@ const setSurrogateKeysHeadersForPubEdges = async (
 
 			const { subdomain, domain } = maybePub.community ?? {};
 
-			const hostname = getCorrectHostname(subdomain, domain);
+			const hostname = subdomain ? getCorrectHostname(subdomain, domain) : req.hostname;
 
 			if (acc.includes(hostname)) {
 				return acc;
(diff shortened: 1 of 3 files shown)


gabestein (approved):
Thanks! Looks good.

gabestein (changes requested):
Found a small issue here that prevents merging, unfortunately. On custom domain pages with any connections at all, it will set the surrogate-key to the community subdomain rather than the hostname, which will prevent purging.

E.g., on `https://duqduqdomaintest.underlay.org/pub/zkme9ef3/release/9`, surrogate-key is gabestein.duqduq.org rather than `duqduqdomaintest.underlay.org`

tefkah:
Hmm i think this should work, I reverted this problem from the previous commit. AFAIK the one on duqduq is the previous one right?
This PR has the same behavior in that regard as the one on prod rn

server/routes/pubDocument.tsx
```diff
@@ -207,15 +207,18 @@ const setSurrogateKeysHeadersForPubEdges = async (
...
+
+			if (acc.includes(hostname)) {
+				return acc;
+			}
 
 			acc.push(hostname);
 			return acc;
 		},
-		[getCorrectHostname(subdomain, domain)] as string[],
+		[req.hostname] as string[],
```
tefkah, line 221: i mean this change here: above is how it is on duqduq, below is how it is on prod

gabestein (approved):

PR #2952, opened 2023-12-15
fix: prevent api route not returning by wrapping actually throwing in…

### LINKED ISSUE #2951: Server never finished responding for /api/citations/zotero requests while not logged in
1. Send GET request to `/api/citations/zotero` locally without logging in
2. See that no response ever gets sent, you'll have to wait till timeout.
...

Get a 403

https://demo.duqduq.org/api/citations/zotero

(be sure to be logged out)

## Optional
An archive scraper bot was spamming this, which lead to ~~outage~~ Heroku notification of too many failed requests.

This bot was causing the 503s

image

Reverse IP lookup show that it's apparently an Archive.org bot. 
image

They are known to make weirdly formatted requests, no clue why it would be just going ham on this specific API you can only normally access in the editor view.

tefkah:
## Issue(s) Resolved

#2951 

## Test Plan
Added tests

Make sure the api still works as expected otherwise (smoke test)

## Notes/Context/Gotchas

This fixes a weird issue where a bot was calling `/api/citations/zotero=q` a bunch, which lead to errors.

Reason why this was erroring out is because
a) the route was not wrapped in `wrap`
b) an error was returned instead of thrown.

Not sure this lead to actual downtime for the rest of the app (see screenshot), but good to fix anyway.

image

### DIFF
--- a/server/citation/api.ts
+++ b/server/citation/api.ts
@@ -1,11 +1,13 @@
 import zoteroClient from 'zotero-api-client';
 
-import app from 'server/server';
+import app, { wrap } from 'server/server';
 import { CitationStyleKind } from 'utils/citations';
 
 import { ZoteroStyleKind } from 'types';
 
 import { expect } from 'utils/assert';
+import { ForbiddenError, BadRequestError, NotFoundError } from 'server/utils/errors';
+
 import { ZoteroIntegration, IntegrationDataOAuth1 } from '../models';
 
 // mapping our citation style keys to zotero's keys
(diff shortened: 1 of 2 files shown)


server/citation/api.ts
```diff
@@ -27,29 +29,40 @@ const zoteroStyleKindMap: Record<CitationStyleKind, ZoteroStyleKind> = {
 
 // the parameters which can be passed to the get() method
 // can be found at https://www.zotero.org/support/dev/web_api/v3/basics
-app.get('/api/citations/zotero', (req, res) => {
-	const userId = req.user?.id;
-	const { q, include, style = 'apa' } = req.query;
-	const zoteroStyle = zoteroStyleKindMap[style as CitationStyleKind];
-	if (!userId) return new Error('Log in to request citations');
```
tefkah, line 34: this was the problem, the error was not thrown.

but if the error was thrown, it wouldn't have been handled properly because it's not `wrap`ed and it's just a generic error, so you get a ton of scary 500s

gabestein:
This is a great find, thank you. It may not have caused an outage at that particular moment, but because it's not handled and times out in 30 seconds, it could easily exhaust our db connections if hit by a bot repeatedly.

We should be on the lookout for more routes like this for sure.

gabestein (approved):
Tested on test instance, looks good!

PR #2960, opened 2023-12-19
feat: logged in user caching + better purging for user related activities

tefkah:
- fix: new caching for logged in users and fixes for user pages
- fix: bust user cache on notification
- fix: purge user pages after user is added or deleted as a member somewhere
- fix: its only pubattributions, not collections
- fix: modify tests


3mcd (approved):

gabestein:
@tefkah -- this functions great. But I have one concern. On the user profile pages, we're using the user ID for the surrogate key rather than the hashed user ID. Any reason not to use the hashed ID here? I'd like to avoid exposing the user ID if we can.

PR #2961, opened 2023-12-20
Adds biophysics colab layout

gabestein:
As described. To test, visit `biophysicscolab` community on a dev environment.


3mcd (approved):
LGTM! No custom footer right?

gabestein:
Thanks! I don't think so, no.

Issue #2958, opened 2023-12-18
Purge session ID surrogate key on login and when new messages arrive


gabestein:
### What?
We can implement caching for logged-in visits by creating a separate cache hash and surrogate key within those hashes for each session. To avoid issues when users logout as one user and login as another in one browser profile, and to make sure new messages always show to users, we need to do a few things:

- Append a hashed user ID to the no cache cookie (e.g. `pp-no-cache-<id>`) on both login and new user creation. We need to keep the no-cache text itself so we can target it effectively in Varnish; otherwise we'll have to target absence, which is never predictable given cookie behavior.
- Update the Fastly config to add this cookie's contents to both the content hash and the surrogate key header
- Purge this surrogate key in fastly when the user receives new notification
- Remove the code that purges the surrogate key on logout, since it's unique to each user and thus no longer needed!

### Why?
Make the logged in experience much faster. Reduce logged in load on server.

gabestein:
Okay @tefkah I think here's what's left after #2960:

- I don't think we should be using hashed user IDs on profile pages. It exposes the user ID to anyone who can use Fastly.
- If you create a Pub in a community and add a user attribution to it, that user's logged in profile, both on the community and on www, logged in and out, is not purged. Looking behind the scenes, this appears to be because on profile pages we're no longer sending (or are overwriting?) the community domain surrogate keys.

These both, unfortunately, delay launch. If we can get them fixed tomorrow, however, I feel confident moving forward.

tefkah:
Great, I'll get on it!

1. I agree, if we take the effort to hide them in the cookie it doesn't make sense to expose them there. I propose we use the users slug, as that is already in the url anyway.
2. Whoops, that is kind of silly, i'll try to figure out why that happens. I probably just need to add the community hostname manually.

tefkah:
2. That problem does exist, but I think it's actually related to releases.

Here is the problem workflow

1. Add user which has no pubs in this community to a pub
2. This sends a purge request to the user page 
3. Go to `www.pubpub.org/user/<user-slug>`. Cache miss, but no pub, because it isn't released.
4. Release Pub
5. Go back to `www.pubpub.org/user/<user-slug>`. Still no Pub, because this page isn't purged on release.
6. Remove and re-add user
7. Go back to `www.pubpub.org/user/<user-slug>`. Now the pub shows up.

So the solution is to also purge all the user page on release of a Pub, or any edit to a Pub that affects that little card, soo

Pub: update, delete
[...]

PR #2963, opened 2023-12-21
feat(hooks.ts): add release hooks to create purge hooks for PubAttrib…

tefkah:
## Issue(s) Resolved
https://github.com/pubpub/pubpub/issues/2958

## Test Plan
Same as https://github.com/pubpub/pubpub/pull/2960, except do this also

### Added Pubs
We want to make sure a user's profile page updates if they are added to a Pub.

1. Login
2. Go to `www.duqduq.org/user/<your-user-slug>`
3. Confirm it is cached by checking for X-Cache: HIT on reload
4. Open new tab, go to a community you have no Pubs in.
5. Add yourself as an attribution to that Pub.
6. Refresh `wwww.duqduq.org/user/<your-user-slug>`, confirm X-Cache: HIT (nothing should have changed)
7. Release the Pub
8 Refresh `wwww.duqduq.org/user/<your-user-slug>`, confirm X-Cache: MISS (you should see the pub now)

## Notes/Context/Gotchas

Main changes are to make sure the comments mentioned here work: https://github.com/pubpub/pubpub/issues/2958#issuecomment-1865135864

The PR also introduces a somewhat more robust way to do purges outside of the standard middleware purge.

I define a `createPurgeHooks` function that works similarly to the `createActivityHooks` function, which works like shown here

https://github.com/pubpub/pubpub/blob/7d4774db96ef9e73bbb8ed023a48d7eb596a768c/server/pubAttribution/hooks.ts#L23-L37

Basically, you define a function that runs after a model is created/updated/destroyed, which returns either a string, and array of strings, or something nullish.
If given an array of strings or a string, it will schedule purges for those Surrogate Keys.

I moved the previously kind of haphazardly placed purges for `PUT /api/users` into such a hook, see comment below.

### DIFF
--- a/server/user/queries.ts
+++ b/server/user/queries.ts
@@ -103,6 +103,7 @@ export const updateUser = (
 
 	return User.update(filteredValues, {
 		where: { id: inputValues.userId },
+		individualHooks: true,
 	}).then(() => {
 		if (req.user.fullName !== filteredValues.fullName) {
 			updateUserData(req.user.id);
--- a/utils/caching/purgeMiddleware.ts
+++ b/utils/caching/purgeMiddleware.ts
@@ -1,11 +1,6 @@
 import { Request, Response, NextFunction } from 'express';
 import { defer } from 'server/utils/deferred';
-import { sequelize } from 'server/models';
-import { QueryTypes } from 'sequelize';
 import type { createCachePurgeDebouncer } from './schedulePurge';
-import { uniqueCommunitiesFromMembersQuery } from './uniqueCommunitiesFromMembersQuery';
-import { getCorrectHostname } from './getCorrectHostname';
-import { getPPLic } from './getHashedUserId';
 import { shouldntPurge } from './skipPurgeConditions';
 
 const ALLOWED_METHODS = ['POST', 'PUT', 'DELETE', 'PATCH'] as const;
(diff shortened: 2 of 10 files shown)


utils/caching/purgeMiddleware.ts
```diff
@@ -126,29 +110,11 @@ export const purgeMiddleware = (
...
-					];
-
-					await Promise.all(allPurges.map(async (tag) => schedulePurge(tag)));
-				} else {
-					/**
-					 * this await is here so you can wait on the purge requests
-					 * using `finishDeferredTasks` in tests or other short lived processes
-					 */
-					await schedulePurge(surrogateTag);
-				}
```
tefkah, line 151: this was moved here: https://github.com/pubpub/pubpub/pull/2963/files#diff-091fe153990dde352868e0d97e0be2e9ebff522544644952f1348961f6cceaab

server/user/queries.ts
```diff
@@ -103,6 +103,7 @@ export const updateUser = (
 
 	return User.update(filteredValues, {
 		where: { id: inputValues.userId },
+		individualHooks: true,
 	}).then(() => {
 		if (req.user.fullName !== filteredValues.fullName) {
 			updateUserData(req.user.id);
```
tefkah, line 109: this is necessary in order for the `User.afterUpdate` hook to run.

by default, those hooks only run for `user.update`, not `User.update({ where: { ...}})`

gabestein (approved):
Tested on the instance to make sure user profile stuff etc. all works and it checks out. Promoting now to check out purging on dev.

PR #2968, opened 2024-01-08
fix: handle pub downloads from base pubpub

### LINKED ISSUE #2964: visits to www.pubpub.org/pub/slug/format/pdf unhandled
See: 
- https://kfg.sentry.io/issues/4755769590
- https://kfg.sentry.io/issues/4443834700

1. Visit eg https://www.pubpub.org/pub/r2s2rfo7/download/pdf
2. Internal server error

Error should be handled in some way. Either PDF should be downloadable from this URL, forwarded to the community first, or return 404.

https://www.pubpub.org/pub/r2s2rfo7/download/pdf

tefkah:
## Issue(s) Resolved
#2964 

## Test Plan
1. Go to \<base\>/pub/r2s2rfo7/download/pdf
2. See 404 instead of unhandled

### Notes/Context/Gotchas
I decided to handle this the same way as releases are handled, i.e. check whether the community is valid and then just forward it

https://github.com/pubpub/pubpub/blob/c2c9d9b880892ac2068b29f0299a7eebfa8f4443/server/routes/pubDocument.tsx#L227-L230

### DIFF
--- a/server/routes/pubDownloads.ts
+++ b/server/routes/pubDownloads.ts
@@ -9,10 +9,14 @@ import { getPubForRequest } from 'server/utils/queryHelpers';
 import { getBestDownloadUrl } from 'utils/pub/downloads';
 import { defer } from 'server/utils/deferred';
 import { createPubExportsForLatestRelease } from 'server/export/queries';
+import { hostIsValid } from 'server/utils/routes';
 
 app.get(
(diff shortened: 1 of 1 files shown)


gabestein (approved):
Thanks! LGTM!

PR #2977, opened 2024-01-18
perf: Add infinite scroll to collection dropdown/minimize calls to /api/collectionPubs

### LINKED ISSUE #2979: Refactor Pub Collection and read next queries to limit Pubs loaded
## Motivations/Context
Because Pubs load all the pubs in their primary collection in the collection dropdown and the read next button, scrapers can bring down PubPub by visiting lots of Pubs in very large primary collections (e.g. BAAS, Criminology) where the query is not cached.

## In Scope
- Limit collection dropdown query to 10 pubs at a time
- For collection dropdown button, add infinite scrolling so next pubs in collection are paginated 10 at a time and load asynchronously
- Replace read next query with simpler one that only loads next pub in collection rather than all pubs

## Out of Scope
- Making sure pub you're on is loaded first
- Finding the perfect number of pubs to load for performance
- Optimizing the query in ways other than pagination
- Adding other query params to align with other API routes

## Requirements
- For read next section, prefer using pagination developed for collection dropdown query via API call over moving it to the server, to keep surface area lower.
- Subjective duqduq testing to make sure spinners load quicker.

tefkah:
- feat: allow GET /api/collectionPubs to be paginated and make it slightly more efficient
- feat: find next pub to read on serverside
- feat: add infinite scroll to CollectionBrowser

## Issue(s) Resolved
#2979 

Previously, repeated fetching of very large collections over `/api/collectionPubs` through the Collection Dropdown could cause the DB to become overwhelmed.
This was especially problematic when crawlers crawled a ton of Pubs in very large collections at the same time, as the query would never succeed, therefore never cache, and would continue to bring too high of a load on the DB.

This PR changes the way that the Pubs in the Collection Dropdown are fetched. Instead of fetching everything all at once, now only 10 pubs are fetched at a time, the rest being loaded in on scroll.
Pubs will also only start being fetched on interaction, which will save a lot of requests, especially those due to both that aren't even interacting with the page in the first place.

The same fetching also happened at the "Read Next" section at the bottom of the page. This has been replaced by a much more efficient server side fetch.

### Read next section
1. Go to `/pub/2021n1i101p01/release/1?readingCollection=27a4f9bb` (locally/duqduq)
2. Go to https://baas.aas.org/pub/2021n1i101p01/release/1?readingCollection=27a4f9bb
3. Check if Read Next section matches
4. Check a number of other Pubs against prod to see if the read next matches

### Dropdown
1. Go to `/pub/2021n1i101p01/release/1?readingCollection=27a4f9bb` (locally/duqduq) on BAAS
2. Click on the collection dropdown
3. Observe spinner
4. Scroll down
5. Observe loading

Ideally repeat with internet throttled to fast 3G to see if the loading speed is tolerable.

### Points of feedback
I'd like thoughts on the following

1. In the spec we wanted to keep the same client fetching mechanism for the Read Next section. I found it easier to just remove that fetch and pass this info through the server on load. Do you agree with this?
2. I chose to only fetch the collection pubs when the user interacts with the dropdown. This is possibly slightly worse UX, as you will always see a spinner when you click the dropdown. However it saves us a request as (most) users don't click the dropdown, especially bots. The query was also running twice every page load, possibly due to a rerendering bug, or possibly due to react strict mode running effects twice (i tried to investigate whether this was the case and am leaning towards this not being the case). Only running the query on open saves me from having to debug this duplicate fetching.

### DIFF
--- a/client/containers/Pub/PubHeader/collections/CollectionBrowser.tsx
+++ b/client/containers/Pub/PubHeader/collections/CollectionBrowser.tsx
@@ -7,11 +7,11 @@ import { usePageContext } from 'utils/hooks';
 import { createReadingParamUrl, useCollectionPubs } from 'client/utils/collections';
 import { pubUrl, collectionUrl } from 'utils/canonicalUrls';
 import { getSchemaForKind } from 'utils/collections/schemas';
+import { useInfiniteScroll } from 'client/utils/useInfiniteScroll';
 import { Collection, Pub } from 'types';
 
 import { usePubContext } from '../../pubHooks';
 import CollectionsBarButton from './CollectionsBarButton';
-import { useInfiniteScroll } from 'client/utils/useInfiniteScroll';
 
 require('./collectionBrowser.scss');
 
--- a/client/containers/Pub/PubHeader/collections/CollectionBrowser.tsx
+++ b/client/containers/Pub/PubHeader/collections/CollectionBrowser.tsx
@@ -11,6 +11,7 @@ import { Collection, Pub } from 'types';
 
 import { usePubContext } from '../../pubHooks';
 import CollectionsBarButton from './CollectionsBarButton';
+import { useInfiniteScroll } from 'client/utils/useInfiniteScroll';
 
 require('./collectionBrowser.scss');
 
@@ -23,7 +24,19 @@ const CollectionBrowser = (props: Props) => {
 	const { collection, currentPub } = props;
 	const { updateLocalData } = usePubContext();
 	const { communityData } = usePageContext();
-	const { pubs, error, isLoading } = useCollectionPubs(updateLocalData, collection);
+	const menuRef = React.useRef<HTMLUListElement | null>(null);
+	const { pubs, error, isLoading, hasLoadedAllPubs, requestMorePubs } = useCollectionPubs(
+		updateLocalData,
+		collection,
+	);
+	const [visible, setVisible] = React.useState(false);
+
+	useInfiniteScroll({
+		enabled: !isLoading && !hasLoadedAllPubs && !error && visible,
+		element: menuRef.current,
+		onRequestMoreItems: requestMorePubs,
+		scrollTolerance: 0,
+	});
 	const { bpDisplayIcon } = getSchemaForKind(collection.kind)!;
 	const readingPubUrl = (pub) => createReadingParamUrl(pubUrl(communityData, pub), collection.id);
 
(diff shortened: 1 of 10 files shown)


server/routes/pubDocument.tsx
```diff
@@ -121,12 +122,17 @@ const getEnrichedPubData = async (options: EnrichedPubOptions) => {
 		};
 	};
 
-	const [docInfo, edges, subscription] = await Promise.all([
+	const currentCollectionPub = pubData.collectionPubs.find((collectionPub) =>
```
tefkah, line 125: this works, but feels slightly sloppy. also did not write any tests or anything to confirm this matches what the "normal" api method finds, just manually tested it.

For reference, here is how it is normally found (on the frontend)
https://github.com/pubpub/pubpub/blob/5041cb89f9b9c8b99402887be5f1ed40be203e87/client/utils/collections.ts#L34-L43

client/containers/Pub/PubHeader/collections/CollectionBrowser.tsx
```diff
@@ -82,6 +88,15 @@ const CollectionBrowser = (props: Props) => {
...
 				))}
+			{isLoading && (
+				<MenuItem
+					disabled
+					className="loading-menu-item"
+					textClassName="menu-item-text"
+					icon={<Spinner size={30} />}
+					text="Loading..."
+				/>
+			)}
```
tefkah, line 99: i moved the spinner down, as loading and showing pubs aren't mutually exclusive anymore.

tefkah:
brought it up to speed with master, should be easier to test on qubqub now

utils/collections/getNextCollectionPub.ts
```diff
@@ -0,0 +1,31 @@
...
+		order: [['rank', 'ASC']],
+		include: [
+			{
+				model: Pub,
+				as: 'pub',
+			},
+		],
+	})) as types.DefinitelyHas<CollectionPub, 'pub'> | null;
+
+	return nextCollectionPub;
```
3mcd, line 30: Very nice! `findOne` + `Op.gt` is clean.

utils/collections/getNextCollectionPub.ts
```diff
@@ -0,0 +1,31 @@
+import { Op } from 'sequelize';
+import { CollectionPub, Pub } from 'server/models';
+import * as types from 'types';
+
+/**
+ * Retrieves the next CollectionPub in a collection based on the current rank
+ *
+ * @param collectionPub - The current CollectionPub
+ * @returns The next CollectionPub, or null if there is no next publication.
+ */
```
3mcd, line 10: Thanks for the JSdoc comment!

3mcd (approved):
Looks and works great!

PR #2985, opened 2024-01-22
feat: allow authentication with Bearer Authorization header for api

tefkah:
## Issue(s) Resolved
-

## Test Plan
1. Run tests

## Supporting Docs

This PR adds a `Bearer <token>` authentication and an API to create/delete these tokens to PubPub.

### Motivation

Using the API without the client (a likely common use-case among those users interested in using the API) is quite difficult at the moment, as our login flow is quite convoluted.
Users would need to

1. Hash their password in a very specific manner.
2. Send a login request with that password and email.
3. Get the specific set-cookie from the response of that request
4. Use that cookie going forward in every request

While fine for the `sdk`, this is kind of annoying for one-off/experimental use of the API, and possibly discourages users for trying out the API.

It is also slightly difficult to properly document these steps on the API documentation. It would be much easier to just include a message that says "you need to send a header with `Bearer <token>`, you can create this token here.", than say give a link to some other piece of documentation that says

1. Pull your password through this hasher (probably some website)
2. Make this specific request
3. Store the answer of that request
4. Include the `pp-lic` from that `Set-Cookie` in every future request.

Logging in also creates a ton of sessions. When using the API/SDK, it's more common to keep logging in/creating new sessions rather than logging out each time you are done/reuse existing sessions.
This creates some unnecessary strain on the DB, and I have had at least one instance of me creating a ton of sessions and no longer being able to edit certain content.

### Workings

This PR adds a table for the tokens, 3 endpoints and some middleware.

#### Model

I chose to have tokens both
- scoped to a user
- scoped to a community

A user can create a token for a specific community if and only if they are an admin in that community, or are a superadmin.

Tokens have an expiration date of
- 1 day
- 1 week
- 1 month
- 3 months
- 1 year
- never

In the DB this is represented as a `Date` or `null`, so theoretically any date could be set as expiration.

#### API

- `POST /api/authTokens` allows a user to create a token for a community they are an admin of.
- `DELETE /api/authTokens/:id` allows a user to delete a previously deleted token, even if they are no longer an admin of that community.
- `DELETE /api/authTokens` allows superadmins to revoke any authtokens. 

#### Middleware

The middleware uses the well-established  [`passport-http-bearer` package](https://www.passportjs.org/packages/passport-http-bearer/)

The middleware is instantiated here
https://github.com/pubpub/pubpub/blob/60ffc89e4572bb7d710b32ddd427c5fb37009572/server/server.ts#L204-L230

and here
https://github.com/pubpub/pubpub/blob/60ffc89e4572bb7d710b32ddd427c5fb37009572/server/server.ts#L164

and is implemented here: https://github.com/pubpub/pubpub/blob/609d6af97e6bc4b6c1b433d0092e529bc1803209/server/authToken/strategy.ts#L1-L65

The strategy is roughly as follows

1. skip this whole thing if the user is already authenticated through the normal cookie, or if the route is not an API route
2. check the `Authorization` header for a `Bearer <token>`
3. find the token in the DB
4. check if the user has admin access for this community.
5. check if the token is not expired
6. authenticate the user as usual

Because only the authentication part of the flow is affected, not the authorization, all existing permission checks can just operate in the same way as before. This saves a lot of rewriting.

## Feedback

I'd like feedback on the following.

### General philosophy

I had the following rules in mind for token creation/deletion and general authentication

- Users should create tokens per community
- Users need to be admins of the community they create tokens for
- Authorization should not be affected by token usage (we should use the normal authorization flow in the api routes)
- Users should not be able to delete each others tokens
- Tokens should stop working once the user is no longer an admin
- Tokens are scoped to the user first, to the community second
- Users should be able to delete their tokens even if they are no longer admins (this is just clean)

#### Expiration

Should we have a `never` option? Might not be the safest.

#### Scope

I chose to have one token per User/Community combination.

I also considered the option of allowing users to create tokens which give them access to multiple communities, but decided to just go for the simpler option for now. 

Do you agree with this?

#### Name

I didn't allow an option to set a name for now, but maybe that would be good. Since tokens are scoped to communities, we can just name them after them. However, atm users can create an unlimited number of tokens per community, so that would get confusing.

#### Format

I just made the token a UUIDv4 for simplicity's sake. I did not really see any benefit of e.g. turning it into a JWT or hashing something, as we just need "something".

If we decide that a UUID is enough, we could also simplify the model further by collapsing the `id` and the `token` field.

### Where should tokens be created

At first, I thought of tokens as being created in the community dashboard, but after reflection I think it would be best if tokens were created on either the "User Profile" page or, on a new "Developer options" page.

My reasoning for this is roughly as follows
- Tokens are used to authenticate a user, not a community
- The community dashboard otherwise has information that anyone could theoretically see, if they have the proper permissions. AFAIK there is no "private" space in the Dashboard that is only accessible to a specific user (except maybe submissions?). Since I don't want other users to be able to see/remove existing auth tokens, auth tokens should not be configured in the dashboard.
- It is convenient to see all your tokens in one place, rather than having to go to each and every community in order to configure a token.

Do you agree with this?

## What is not included

- UI. That can be done later.
- Token names.
- Some tests


types/user.ts
```diff
@@ -1,7 +1,10 @@
 import { User as UserModel } from 'server/models';
 import { SerializedModel } from './serializedModel';
 
-type UserType = SerializedModel<UserModel>;
+type UserType = SerializedModel<UserModel> & {
+	// this is useful to prevent repeated checks for admin status
+	isAdminOfThisCommunity?: boolean;
+};
```
tefkah, line 7: we need to do admin checks for auth tokens (bc what if someone gets removed as admin for a community)

however, a lot of routes also do these kinds of auth checks, which is redundant. this acts as a sort of "cache" for those checks.

this works if api requests never cross community boundaries, which i'm pretty sure they don't

tefkah:
One thing I'd like feedback on specifically: what should the token look like? Atm it's just a UUID, but maybe it should be/look different? Don't really have strong opinions on it tbh.

3mcd (approved):
Sorry this took me so long to review/approve!

PR #2988, opened 2024-01-23
spike: analytics

### LINKED ISSUE #2976: Spike: Analytics
- Schema for sending data
- UI for switching
- What GA stuff to support
- How to fingerprint (if at all) our own analytics

tefkah:
## Issue(s) Resolved

#2976 

## Description

This PR adds custom analytics to the site, replacing Heap.io.

### Page views
1. In turn, go to
  - A Pub
  - A Collection
  - A Page (landing page, about page, etc.)
2. Open Network Tab
3. See one and only one request to `/api/analytics/track`, with `"type": "page"` and  `"properties": { "type": "pub" | "collection" | "page"}` respectively 

### Downloads
1. Go to a Pub with a formatted download
2. Open network tab
3. Click, in turn, on download for
  - Formatted
  - `docx`
4. Observer one and only one request to `/api/analytics/track`, with `"type": "download"` and `"properties": { "type": "formatted" | "docx" }` respectively
5. Refresh page
6. Repeat steps 3-4
7. Observe the same result on as 4 (to confirm that caching the download does not affect analytics) 

### Feature flagging

Only communities with the `newAnalytics` feature flag should have events tracked.

At the moment these are `demo` and `jtrialerror`.

1. Go to `demo`
2. See request made to `/api/analytics/track`
3. Go to `baas`
4. See no such event

### Google

`jtrialerror` is set up to use GTM

Communities using GTM/GA should show a cookie banner, and should not load GA code until that is accepted.

1. Go to `jtrialerror` & open network tab
2. Clear cookies & reload
3. See cookie banner
4. Check network tab, see no requests to any google domain
5. Click "Accept"
6. See some requests to googletagmanager.com and a request to some ngrok/collect endpoint. (i set that up for testing, no longer active)


client/utils/heap.ts
tefkah, line 1: begone heap

client/containers/App/App.tsx
```diff
@@ -104,40 +123,42 @@ const App = (props: Props) => {
...
+							</div>
+							{showFooter && footer}
 						</div>
-						{showFooter && footer}
-					</div>
-				</RKProvider>
-			</FacetsStateProvider>
+					</RKProvider>
+				</FacetsStateProvider>
+			</AnalyticsProvider>
```
tefkah, line 161: not entirely sure this is the appropriate depth, could be lower perhaps

client/components/Layout/Layout.tsx
```diff
@@ -28,10 +30,44 @@ type Props = LayoutOptions & {
...
+					collectionSlug: collection.slug,
+			  }
+			: pageData
+			  ? {
+						type: 'page' as const,
+						communityId: communityData.id,
+						pageSlug: pageData.slug,
+						title: pageData.title,
+			    }
+			  : undefined;
```
3mcd, line 62: Might be worth a comment describing when this `undefined` case would occur. And better yet if there's a way to isolate that case to an `if` statement above it would simplify this ternary a little.

server/server.ts
```diff
@@ -109,8 +108,20 @@ if (process.env.NODE_ENV === 'production') {
 app.use(deduplicateSlash());
 app.use(noSlash());
 app.use(compression());
-app.use(bodyParser.json({ limit: '50mb' }));
-app.use(bodyParser.urlencoded({ limit: '50mb', extended: true }));
+app.use(express.json({ limit: '50mb' }));
+app.use(express.urlencoded({ limit: '50mb', extended: true }));
+// needed to parse analytics events from the client sent with `navigator.sendBeacon`
+app.use(express.text(), (req, res, next) => {
```
3mcd, line 114: Should this be route-level middleware instead? I'm not sure any other cases where we send text to the server but could there be a performance penalty here?

3mcd (approved):
Left a couple of optional suggestions. Otherwise looks great.

server/server.ts
tefkah: good point!

3mcd (approved):

tefkah:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **TypeError: Cannot read properties of null (reading 'type')** `GET /pub/:pubSlug/release/:releaseNumber` [View Issue](https://kfg.sentry.io/issues/4915791255/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

This is bc I changed the DB schema, #2991 should fix that

Issue #2994, opened 2024-01-25
analytics: create feature flag to gatekeep communities from using other analytics providers


tefkah:
### What?

Currently, there is a feature flag for gating communities from using the new analytics system, defaulting back to the old Heap system.

We also want a feature flag that prevents communities from using e.g. the GA/Simple plugins, such that we can enable the new analytics suite without having everyone using GA willy nilly

### Why?
-

### Initial estimate

0.5

## Issue Readiness
ready

Issue #2993, opened 2024-01-25
analytics: make other analytics providers track in addition to, instead of replace, our custom analytics solution


tefkah:
### What?

Currently, you either pick between our custom analytics solution or between GA/Simple. 

Instead, we always want to do our custom tracking and allow users to also have GA/Simple tracking.

### Why?

Good to have internal metrics.

### Initial estimate

0.5

### Implementation checklist

- [ ] Make sure all tracking etc events fire at the same time
- [ ] Have internal tracking fire even if cookie banner is not accepted for GA
- [ ] Internal tracking should not refire when cookie banner is accepted
- [ ] Remove `{ type: 'default', credentials: null }` setting for using default analytics.

## Issue Readiness

ready

Issue #2992, opened 2024-01-25
analytics: add Simple analytics plugin


tefkah:
### What?

Add Simple Analytics as an option for analytics. 

### Why?

Simple Analytics is a privacy friendly analytics provider which has an existing plugin for `analytics`, making it trivial to add support for.

### Initial estimate

0.5 days

## Issue Readiness

Ready

PR #2999, opened 2024-01-25
feat: remove gtm and add simple analytics, and gatekeep communities from using custom analytics behind a feature flag

### LINKED ISSUE #2991: analytics: remove GTM plugin
### What?

Remove the GTM plugin as an option for analytics.

### Why?

GTM could allow for way too invasive tracking. Maybe add it back later if we can get users to sign a contract limiting what they can track using GTM.

### Initial estimate

0.5 days

## Issue Readiness

ready

tefkah:
## Issue(s) Resolved
#2991 
#2992 
#2993 
#2994 

The communities on dev have been set up as follows

1. `demo`: access to new analytics, does not use custom provider
2. `baas`: no access to new analytics, has google analytics set as custom provider
3. `jtrialerror`: access to new analytics, has google analytics set as custom provider

## Test Plan

Before each of these, clear your cookies and open your network tab.

### Non-enrolled

1. Go to `hdsr`
2. Check that no events are sent to `/api/analytics/track` nor to `googletagmanager`
3. Check that cookie banner shows as normal

### Using new but no custom analytics

1. Go to `demo`
2. Check that no cookie banner is popped up
3. Check that an event is sent to `/api/analytics/track` when visiting a page

### No access to custom analytics

1. Go to `baas`
2. Check that no cookie banner pop up
3. Check that event is sent to `/api/analytics/track`, but not to `googletagmanager`

### Using new and custom analytics

`jtrialerror`

#### Default flow
1. go to `jtrialerror`
2. Check that cookie banner opens (do not click anything)
3. Check that event is sent to `/api/analytics/track`
5. Check that no event is sent to `googletagmanager.com`

Then

6. Click Accept
7. Check that a request is sent to googletagmanager.com to download the code and that another event is sent (you may need to wait a second) to `.../collect`. Make sure a second request is NOT sent to `/api/analytics/track`.
8. Reload page (don't clear cookies)
9. Check that request is sent to `googletagmanager.com` and tracking events to `/api/analytics/track` and `somegoogledomain/collect`

#### Decline flow

Clear cookies and reload page

1. Check event sent to `/api/analytics/track`
2. Click decline on the cookie banner
3. Check that no further events sent
4. Reload
5. Check only event to `/api/analytics/track` is sent, and cookie banner does not pop up.

#### Logged in flow

1. Log in
2. Go to privacy settings and set to "allow third party cookies"
3. Go to home page
4. Reload
5. Check that events are sent to both google and `/api/analytics/track`
6. Go back to privacy settings and set to "do not allow third party cookies"
7. Go to home page
8. Reload
10. Check that only event is sent to `/api/analytics/track`


utils/analytics/useAnalytics.ts
```diff
@@ -61,3 +71,71 @@ export const useAnalytics = () => {
...
+ *
+ * 1. User visits page
+ * 2. Page view event is sent to only our analytics service
+ *
+ * @param payload - The payload containing the page view data or a function that returns the
+ *   payload.
+ * @param gdprConsent - Optional boolean indicating whether GDPR consent has been given.
+ */
+export const usePageOnce = (
+	payload: PageViewPayload | (() => PageViewPayload | null),
```
3mcd, line 102: What's the reasoning for having `payload` be either the payload or a function that returns a payload (invoked immediately with no arguments)? It appears to be used only in the former case.

utils/analytics/useAnalytics.ts
```diff
@@ -61,3 +71,71 @@ export const useAnalytics = () => {
...
+ *
+ * 1. User visits page
+ * 2. Page view event is sent to our analytics service
+ * 3. User clicks "Accept" on the GDPR banner
+ * 4. Page view is sent to Google Analytics.
+ *
+ * Flow for Google Analytics + consent refused:
+ *
+ * 1. User visits page
+ * 2. Page view event is sent to only our analytics service
```
3mcd, line 95: Very nice documentation.

tefkah: thank you :)

3mcd (approved):
Ran through the test plan. Everything worked great!

PR #3008, opened 2024-02-05
fix: limit the amount of memory jest consumes

tefkah:
## Issue(s) Resolved

CI often fails because Jest goes OOM.

## Test Plan

Run tests, should not fail due to OOM errors.

### Notes/Context/Gotchas

This runs jest with the `--expose-gc` node flag as per this stockoverflow answer: https://stackoverflow.com/questions/62885390/my-jests-tests-are-leaking-memory-how-can-i-fix-this

I swore I tried this when I first starting to debug the tests last year, but apparently not! 

With this, the tests do not go above 1200mbish, while before they would go to about 2500 and crashed often.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
 		"storybook": "start-storybook -p 9001 -c .storybook -s ./",
 		"test": "npm run lint && npm run test-no-lint",
 		"test-dev": "jest --forceExit",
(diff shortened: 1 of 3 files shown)


tefkah:
I am merging this as its a change that only affects the CI and allows me to actually set up the other PRs to be merged! Apologies!

PR #3012, opened 2024-02-08
feat: add additional fields to the analytics schema

tefkah:
- refactor: require community name/subdomain for track events
- feat: add country code to data
- feat: add isProd flag to analytics event payload

## Issue(s) Resolved

This semi-finalizes the schema for analytics.

Explanation of the extra flags

### Community name/id/subdomain for every event

This makes grouping events easier in metabase.
Before I only required this info for pageviews, but I now also do for track events.

### Country code

This allows us to use a different map in metabase that is slightly nicer.
Specifically, this map has much easier to see/click countries. To pick a completely random and non-biased example, the Netherlands is quite hard to see on the default one.

### isProd

I decided that it would be much easier to store Prod and Dev analytics in the same db. 

The way metabase works means we basically have to duplicate every view/model/dashboard if we use two different databases, making it very hard to keep things in sync. 

By using the same database, we only have to create two versions of every dashboard, which is much easier to do.

The downside is that schema experimentation in dev is slightly more risky, as it will also affect the prod db.
I do not think this is a significant issue though, since most schema changes moving forward would probably be additive.

## Test Plan

1. Go to `demo`
2. Check whether analytics event sent to `/api/analytics/track` contains the `isProd` flag. It should be false

You could also try to do this locally

### DIFF
--- a/client/components/Layout/Layout.tsx
+++ b/client/components/Layout/Layout.tsx
@@ -41,16 +41,21 @@ const Layout = (props: Props) => {
 			return null;
 		}
 
+		const base = {
+			communityId: communityData.id,
+			communityName: communityData.title,
+			communitySubdomain: communityData.subdomain,
+		};
+
 		if (collection) {
 			return {
 				event: 'collection' as const,
-				communityId: collection.communityId,
 				title: collection.title,
 				collectionId: collection.id,
 				collectionTitle: collection.title,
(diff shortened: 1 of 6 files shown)


tefkah:
@gabestein  Would like your thoughts on merging the prod and db databases!

PR #3017, opened 2024-02-19
fix: add client and server side validation for orcids

### LINKED ISSUE #3014: Front-end validation of contributor ORCID
In Pub (and collection) contributor screens, it's possible to fill out invalid ORCIDs, which can cause DOI deposits to fail. It would be good to validate ORCIDs here on the frontend, similarly to how we do on user profiles (e.g. stripping out the orcid.org domain so that only the ID itself is submitted -- even though, yes, technically the ID is the full URL).

Prevent invalid ORCIDs from being added to contributors.

tefkah:
## Issue(s) Resolved

Closes #3014 

## Test Plan

1. Run tests

1. Open network tab
2. Create new attribution for a pub
3. Try to add the following orcids:
	- 0000-0000-0000
	- https://orcid.org/0000-0000-0000-0000
	- http://orcid.org/0000-0000-0000-0000
	- https://www.orcid.org/0000-0000-0000-0000
	- orcid.org/0000-0000-0000-0000
	- www.orcid.org/0000-0000-0000-0000
4. Observe that the response of all of these is `{"orcid": "0000-0000-0000-0000"}`
5. Also do this for collection attributions
6. Check that when creating an orcid with `https://orcid.org`, this is gone when reloading 
7. Try to create some invalid orcids and observe an error message popping up, AND that the API response 404s.

### Notes/Context/Gotchas

I decided to allow `https://orcid.org/ID` to be filled in on the frontend and to be accepted by the backend, but have these be transformed into just the ID part with Zod.

https://github.com/pubpub/pubpub/blob/23258ba3bc4ee3b7c46daff500b99e411114e5e9/utils/api/schemas/attribution.ts#L18-L23

This makes the user experience slightly better i think.

Another change to note is when the change is submitted to the API, see the comments below for that.
TLDR: no more submission on Enter, instead autosubmission if correct orcid is sent.

### DIFF
--- a/client/components/AttributionEditor/AttributionDetailControls.tsx
+++ b/client/components/AttributionEditor/AttributionDetailControls.tsx
@@ -3,8 +3,10 @@ import { Checkbox, Classes, InputGroup, MenuItem, Position, Tag } from '@bluepri
 import { MultiSelect } from '@blueprintjs/select';
 
 import { AttributionWithUser } from 'types';
+import { ORCID_ID_PATTERN } from 'utils/orcid';
 
 import { getFilteredRoles } from './roles';
+import InputField from '../InputField/InputField';
 
 type Props = {
 	attribution: AttributionWithUser;
@@ -18,6 +20,8 @@ const AttributionDetailControls = (props: Props) => {
 	const { attribution, isShadowAttribution, listOnBylineText, onAttributionUpdate, roles } =
 		props;
 	const { affiliation, id, isAuthor, orcid } = attribution;
+
+	const isOrcidInvalid = Boolean(orcid && !orcid.match(ORCID_ID_PATTERN));
 
 	return (
 		<div className="detail-controls">
@@ -112,7 +116,7 @@ const AttributionDetailControls = (props: Props) => {
 					}
 				/>
 				{isShadowAttribution && (
-					<InputGroup
+					<InputField
 						// @ts-expect-error ts-migrate(2322) FIXME: Type '"" | Element | undefined' is not assignable ... Remove this comment to see the full error message
 						rightElement={orcid && <Tag minimal>ORCID</Tag>}
 						placeholder="ORCID"
@@ -127,6 +131,11 @@ const AttributionDetailControls = (props: Props) => {
 								id,
 								orcid: evt.target.value.trim(),
 							})
+						}
+						error={
+							isOrcidInvalid
+								? 'Invalid ORCID. Please enter a valid ORCID or leave the field blank.'
+								: undefined
 						}
 					/>
 				)}
--- a/client/components/AttributionEditor/AttributionDetailControls.tsx
+++ b/client/components/AttributionEditor/AttributionDetailControls.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { Checkbox, Classes, InputGroup, MenuItem, Position, Tag } from '@blueprintjs/core';
 import { MultiSelect } from '@blueprintjs/select';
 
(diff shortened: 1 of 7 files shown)


server/pubAttribution/api.ts
```diff
@@ -25,58 +22,6 @@ const getRequestIds = createGetRequestIds<{
...
-			z.object({
-				userId: z.undefined().optional(),
-				name: attributionSchema.shape.name.unwrap(),
-				orcid: attributionSchema.shape.orcid.unwrap().optional(),
-			}),
-		]),
-	) satisfies z.ZodType<
-	Omit<types.PubAttributionCreationParams, 'order' | 'pubId'> & { order?: number }
->;
-
```
tefkah, line 79: this was a vestigial schema not in use anymore

package.json
```diff
@@ -349,7 +349,7 @@
 		"ts-loader": "^8.0.11",
 		"ts-morph": "^21.0.1",
 		"ts-node": "^10.9.1",
-		"typescript": "^5.2.2",
+		"typescript": "^5.3.3",
```
tefkah, line 352: i ran into a really weird bug with typescript where it would yell at me saying the output schema of `GET /api/pubs` was incompatible for some reason. Never happened in the editor, only when running `tsc`. Could not figure out what caused it, seemed like a false alarm, upgrading solved this.

client/components/AttributionEditor/AttributionDetailControls.tsx
```diff
@@ -105,28 +121,41 @@ const AttributionDetailControls = (props: Props) => {
...
-						onAttributionUpdate({
+						onAttributionUpdateWithValidation({
 							id,
 							affiliation: evt.target.value.trim(),
 						})
 					}
 				/>
 				{isShadowAttribution && (
-					<InputGroup
+					<InputField
```
tefkah, line 131: I needed to use `InputField` bc `InputGroup` does not have the `onError` thingy.

client/components/AttributionEditor/AttributionDetailControls.tsx
```diff
@@ -105,28 +121,41 @@ const AttributionDetailControls = (props: Props) => {
...
+							const orcidInvalidity = isOrcidInvalid(input);
+							setInvalidOrcid(orcidInvalidity);
+
+							if (!orcidInvalidity) {
+								onAttributionUpdateWithValidation({
+									id,
+									orcid: input.trim(),
+								});
 							}
 						}}
```
tefkah, line 147: `onKeyDown` is not available for `InputField`, so I removed it. This is somewhat annoying bc this removes the ability to "save" on Enter for this field.

I can't seem to capture `Enter` events on `onChange`, so instead, if the value matches the RegEx, we just submit the change. I see no real harm in this, as we used to and still submit changes on blur.

utils/api/schemas/attribution.ts
```diff
@@ -14,7 +15,15 @@ export const attributionSchema = baseSchema.extend({
...
-	orcid: z.string().nullable(),
+	orcid: z
+		.string()
+		// allow both url or id to be submitted
+		.regex(ORCID_ID_OR_URL_PATTERN)
+		// grab just the id part of the pattern
+		.transform((orcid) => orcid.match(ORCID_PATTERN)?.[0]!)
+		.nullable()
+		// for historical reasons, we allow empty strings to be submitted
+		.or(z.literal('')),
```
3mcd, line 26: Very nice! Zod is so expressive

3mcd (approved):

gabestein (approved):

PR #3042, opened 2024-03-14
feat: allow heap and new analytics to be used simultaneously, and add AWS endpoint instead of local one

tefkah:
- fix: allow stitch even with new analytics
- chore: add analytics lambda endpoint

## Issue(s) Resolved
Currently Heap and New analytics cannot both be turned on at the same time. Specifically, heap would not get initialized when new analytics is turned on. This PR gets rid of that flag.

It also sends the events to the new AWS endpoint instead of the old /api/analytics/track endpoint.

## Test Plan
(on demo, on either qubqub or duqduq)
1. Open network tab
2. Go to home
3. See event to `https://yhzkpvgmsji7wkkmeplrtgoj5y0qbzpp.lambda-url.us-east-1.on.aws` with a payload and `event: 'page'` and an event to `heap.io`
4. Go to a collection
5. See event to AWS endpoint with `event: 'collection'` and an event to `heap.io`
6. Go to a pub
7. See event to AWS endpoint with `event: 'pub'` and an event to `heap.io`

### Notes/Context/Gotchas
I hardcoded the URL to AWS, as I don't know how to pass environment variables to the client.

### Follow ups
After this PR is merged, a `shouldUseNewAnalytics` feature flag should be created in prod and turned on for everyone.

### DIFF
--- a/client/utils/hydrateWrapper.tsx
+++ b/client/utils/hydrateWrapper.tsx
@@ -3,7 +3,6 @@ import * as Sentry from '@sentry/react';
 import { hydrate } from 'react-dom';
 import { FocusStyleManager } from '@blueprintjs/core';
 
-import { shouldUseNewAnalytics } from 'utils/analytics/featureFlags';
 import { setEnvironment, setAppCommit } from 'utils/environment';
 
 import { getClientInitialData } from './initialData';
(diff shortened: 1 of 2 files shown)


gabestein (approved):
Tested on duqduq, looks great. Love the rewrite on fastly as a way of keeping all the traffic local. We should demo that to the devs, super useful thing we could make use of in the future.

PR #3049, opened 2024-03-21
fix: show banner again, remove no longer needed new analytics flag, and refactor usepageonce to be able to be used higher up and for more pages

tefkah:
## Issue(s) Resolved
- [x] Cookie banner was not showing
- [x] Could not track events on e.g. `www.pubpub.org/explore`

## Test Plan

Start of with no cookies each time, devtools open

1. Go to `jtrialerror`
2. Check that event is sent (will fail on qubqub, not important) & cookie banner pops up
3. Accept cookie banner
4. See that google analytics loads, and some time later an event to `googlesomething.com/collect` is sent
5. Repeat ideally for a Pub, Collection, and `/search`

_____

1. Go to `jtrialerror`
2. Decline cookie banner
3. Check that google analytics does not load

________

1. Go to `demo`
2. Check that event is sent and cookie banner opens

________

1. Go to `www.pubpub.org/explore`
2. See that event is sent


gabestein (approved):
Thanks. LGTM!

PR #3063, opened 2024-03-27
Feat: condensed body contributor

gabestein:
## Issue(s) Resolved
Adds a condensed contributor to the Pub body behind a feature flag to match Arcadia's current manual version.

## Test Plan
- Load arcadia-research locally or on dev, where feature flag is already set.
- Load an extent pub, e.g. http://localhost:9876/pub/idea-defining-actin/release/2?collectionSlug=functional-annotation
- Check that new section matches, with addition of avatar and links to profiles.
- Check that it breaks similarly on mobile (1024 and 728)

## Launch plan
Launching will require setting existing contributor section to display none.

After that's done, add `bodyContributors` feature flag and arcadia-research community to flag.

Screenshot 2024-03-27 at 14 56 46

### Notes/Context/Gotchas
- Could use community primary/secondary colors in the future rather than hard-coding
- Could be extended in the future to automate alphabetic sorting of contributor roles

### DIFF
--- a/client/components/ContributorsListCondensed/ContributorCondensed.tsx
+++ b/client/components/ContributorsListCondensed/ContributorCondensed.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Avatar, Icon } from 'components';
+
+require('./contributorCondensed.scss');
+
+const propTypes = {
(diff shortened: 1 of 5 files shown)


tefkah (approved):
Works as expected! Would indeed be nice if the color of the separator would use a community color, but that can probably be added when this element is needed in a different community.

PR #3064, opened 2024-04-02
Fix discussion email sending

### LINKED ISSUE #3035: Discussion reply and digest emails not sending

1. Make sure your email settings are set to receive replies to comments in your preferences: https://www.pubpub.org/legal/settings
2. Create a new pub
3. From another account, add a discussion to that pub
4. Notice that no email is sent, and no attempts show up in mailgun logs

It appears that emails may have stopped for both in August 2023!

Email notification of discussion is sent.

https://notes.knowledgefutures.org/pub/jph8ooz1/draft

## Optional
Repeat for activity digests (make sure subscribe is turned on in e.g. https://notes.knowledgefutures.org/dash/activity)

kalilsn:
Discussion emails are sent by an `afterCreate` hook on `UserNotification` creation. However, these hooks don't run by default on bulk methods like `UserNotification.bulkCreate`.

https://sequelize.org/docs/v6/other-topics/hooks/#model-hooks:
>Note: methods like bulkCreate do not emit individual hooks by default - only the bulk hooks. However, if you want individual hooks to be emitted as well, you can pass the { individualHooks: true } option to the query call. However, this can drastically impact performance, depending on the number of records involved (since, among other things, all instances will be loaded into memory).

## Issue(s) Resolved
Resolves #3035

## Test Plan
Create a pub, comment on it, then reply from another account. You should receive an email.

### Notes/Context/Gotchas
~I'm not quite sure how this ever worked, since it doesn't seem like sequelize changed the bulkCreate api recently (this behavior is in the v5 docs too).~ I see now that this was deleted in https://github.com/pubpub/pubpub/pull/2960/commits/06f99929a2b231d417eb3387f3019ca047e3cb15 hopefully just by mistake, though let me know if that's wrong and this is breaking something else.

I'm not sure how the activityDigest email script gets run, but if those aren't being sent it seems like a separate issue since they don't use hooks.

### DIFF
--- a/server/userNotification/create.ts
+++ b/server/userNotification/create.ts
@@ -93,6 +93,7 @@ const createNotificationsForThreadComment = async (
 				activityItemId: item.id,
 			};
 		}),
+		{ individualHooks: true },
 	);
 
 	// purge cache for users who are receiving notifications
--- a/server/userNotification/__tests__/hooks.test.ts
+++ b/server/userNotification/__tests__/hooks.test.ts
@@ -8,6 +8,11 @@ import { DocJson } from 'types';
 const models = modelize`
     User rando {}
     User chattyUser {}
+	User craveEmailUser {
+		UserNotificationPreferences {
+			receiveDiscussionThreadEmails: true
+		}
+	}
     Community {
         Member {
             permissions: "admin"
@@ -26,12 +31,19 @@ const models = modelize`
                 permissions: "view"
                 User sickOfThisThreadUser {} 
             }
+			Member {
+				permissions: "view"
+				user: craveEmailUser
+			}
             UserSubscription {
                 user: pubSubscriber
             }
             UserSubscription {
                 user: sickOfThisThreadUser
             }
+			UserSubscription {
+				user: craveEmailUser
+			}
             Discussion membersDiscussion {
                 number: 0
                 author: chattyUser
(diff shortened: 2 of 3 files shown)


tefkah:
I can confirm that I got emails both from people adding comments to the Pub and from replying to a discussion I started!

server/userNotification/create.ts
```diff
@@ -93,6 +93,7 @@ const createNotificationsForThreadComment = async (
 				activityItemId: item.id,
 			};
 		}),
+		{ individualHooks: true },
```
tefkah, line 96: I'm _pretty_ sure I removed this by mistake. I came across this bug/"feature" in other places as well, I remember adding this somewhere, but don't think I ever intentionally removed it. Very sorry!

tefkah:
I added a test to make sure the emails are sent correctly.
Normally I would just ask for the test to be added, but since mocking and checking if something has been called is a bit tricky in v6 and I happened to have done that before, I just added it myself real quick! (also I was the one that broke it and want to make sure i did not break it again 😅)

tefkah (approved):

server/userNotification/create.ts
kalilsn: not a problem! thanks for adding the test!

PR #3065, opened 2024-04-02
Add ts-node to prod build so heroku scheduler can use it

kalilsn:
Our scheduled tasks have been failing for a while since ts-node isn't in the prod build. See https://my.papertrailapp.com/groups/14651822/events?q=program%3Ascheduler
This PR adds it back, and adds sentry to the `npm run tools` script so that we'll get some kind of alert if these tasks fail in the future.
## Issue(s) Resolved
#3064 (the activity digest portion)
## Test Plan
- I don't know of a way to test this without merging it (or potentially deploying from this branch) and waiting for the task to run. Should be low risk though!
- I tested the sentry handler locally by adding an explicit `throw new Error()` into the tools script and produced [this alert ](https://knowledgefutures.slack.com/archives/CLF7E2V61/p1712173041667359)


tefkah:
Looks good!

I agree that there isn't really any way to check this except for deploying it, but I don't think there's any risk in trying it that way.

If we wanted to be super careful we could quickly add the scheduler to DuqDuq after this is merged, but I don't really think it's necessary.

tefkah (approved):

tools/index.js
```diff
@@ -8,6 +8,13 @@ if (process.env.NODE_ENV !== 'production') {
 	// FIXME: Weird eslint issue where either with or without the extension is an error
 	// eslint-disable-next-line import/extensions
 	require('../config.js');
+} else {
+	const Sentry = require('@sentry/node');
+	Sentry.init({
+		dsn: 'https://abe1c84bbb3045bd982f9fea7407efaa@sentry.io/1505439',
+		environment: 'prod',
+		tracesSampleRate: 1.0,
+	});
```
tefkah, line 17: nice, i appreciate you always being on top of adding error reporting everywhere!

PR #3070, opened 2024-04-09
Fix type error preventing activity digest email sending

kalilsn:
#3035 
Fixes a new error in the activity digest script: https://my.papertrailapp.com/groups/14651822/events?q=program%3Aapp%2Fscheduler.3606&focus=1714219360833110031&selected=1714219360833110031

Also modifies the script to explicitly send these errors to sentry since they're being caught right now.

## Test Plan
Comment out the mailgun sending lines (L52-58) in `emailActivityDigest.ts`, then run `npm run tools -- emailActivityDigest` and it shouldn't error when rendering a digest email. Or just wait until this is merged and check that it succeeds via papertrail.


tefkah (approved):
That seems to have been the issue, LGTM!

PR #3071, opened 2024-04-10
Capture errors from backups script and update google dependencies

kalilsn:
## Issue(s) Resolved
`google-auth-library` was updated without updating `@google-cloud/storage` leaving them at incompatible versions and causing the backups script to fail with the following error

https://knowledgefutures.slack.com/archives/CGWSQ4CV7/p1712732463644249
>Error: Unable to detect a Project Id in the current environment. 
To learn more about authentication and Google APIs, visit: 
https://cloud.google.com/docs/authentication/getting-started
    at /app/node_modules/@google-cloud/storage/node_modules/google-auth-library/build/src/auth/googleauth.js:96:31
    at processTicksAndRejections (node:internal/process/task_queues:95:5)

## Test Plan
`npm run tools backup`


tools/backup/firebase.ts
```diff
@@ -41,13 +41,19 @@ const downloadGcpFile = async (file: File): Promise<BackupFile> => {
...
+					size = parseInt(metadata.size, 10);
+				} else {
+					size = metadata.size;
+				}
+				return resolve({
 					localPath,
 					remotePath: file.name,
-					size: parseInt(metadata.size, 10),
-				}),
-			)
```
kalilsn, line 50: `metadata.size` went from `string` to `string | number | undefined`.  I don't really know what IRL situation those types correspond to and couldn't find documentation but I think this should be fine

tefkah (approved):
seems to work, in that it does not seem to fail in the same way as before and successfully allows you to download files from google cloud storage. I did not finish running the entire backup script, as I a) don't really know if that's a good idea to run locally and b) it takes very long haha. I do think it should work as intended now!

PR #3086, opened 2024-05-13
fix: slow loading speeds for large collections with lots of attributions/members

### LINKED ISSUE #3078: Error in loading collection overview

1. User wants to access collection via Community Overview Dashboard
2. User selects the collection from the dashboard
3. Collection Overview Dashboard does not load (returns a 503 error).

Collection Overview Dashboard would load, displaying collection overview. 

https://openrewi.pubpub.org/dash/collection/verwaltungsrecht-in-der-klausur

Firefox

## Optional
Given the size of the collection (121 pubs), it looks like the system is timing out the request. Not sure if this is true of large collections in general?

Screen Shot 2024-04-26 at 9 30 34 AM
Screen Shot 2024-04-26 at 9 31 03 AM

Nikolas//OpenRewi

tefkah:
- **fix: add getPubs configuration to not overfetch collections**
- **dev: add tests for new configuration**

## Issue(s) Resolved
Resolves #3078

## Test Plan

Test overview is BAAS `/dash/collection/vol-54-issue-8/members`
(don't go there on duqduq, it will crash it)

1. Go to it on QubQub/locally, observe it load as intended.

___

Regression check

1. Create two collections that are not tags
2. Create a pub
3. Add the Pub to both collections, on of them will be the Pub's primary collection.
4. Go to both collection overviews
5.  observe that in one of them the Pub has a filled out star (the primary collection), in the other it has an empty star.
6. In the non-primary collection overview, click the empty star next to the Pub, and observe that the name of the other collection is correct.
7. Click on the empty star again, and make this collection the primary collection.

____

### Notes/Context/Gotchas
I think i fixed this, i think it's due to two things

rather large collection
relatively lot of members
by default, when fetching the pubs for the overview, we fetch every collection pub for every pub, and for every collection pub we fetch the entire collection, and for every collection we fetch every member/attribution

this collection had something like 140 pubs and like 40ish attributions + members
this ends up fetching ~5-6000 members/users/attributions, which i think is why this takes so long to complete.

i put up a PR that adds extra configuration to getPubs, which allows you to just fetch the collection for a collectionPub, which is a lot lighter.

we can't get away with not fetching the collection for each collectionpub for each pub, as it is needed to show what other primary collection a pub is in, if the current collection is not the primary collection. otherwise it always shows the current collection to be the primary collection.

### DIFF
--- a/server/utils/queryHelpers/__tests__/pubOptions.test.ts
+++ b/server/utils/queryHelpers/__tests__/pubOptions.test.ts
@@ -1,0 +1,124 @@
+import {
+	CollectionPub,
+	Page,
+	Member,
+	CollectionAttribution,
+	includeUserModel,
+} from 'server/models';
+import { Collection } from 'yaml/types';
+import buildPubOptions from '../pubOptions';
+
+describe('buildPubOptions', () => {
+	test('should handle getCollections = true', () => {
+		const { include } = buildPubOptions({ getCollections: true });
+
+		expect(JSON.stringify(include.find((x) => x.as === 'collectionPubs'))).toEqual(
+			JSON.stringify({
+				model: CollectionPub,
+				as: 'collectionPubs',
+				separate: true,
+				order: [['pubRank', 'ASC']],
+				include: [
+					{
+						model: Collection,
+						as: 'collection',
+						include: [
+							{
+								model: Page,
+								as: 'page',
+								attributes: ['id', 'title', 'slug'],
+							},
+							{
+								model: Member,
+								as: 'members',
+							},
+							{
+								model: CollectionAttribution,
+								as: 'attributions',
+								include: [includeUserModel({ as: 'user' })],
+							},
+						],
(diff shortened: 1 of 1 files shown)


gabestein (approved):
It tests fine.

But, there is an odd error that happens when trying to create a pub in BaaS specifically on qubqub (in multiple review apps) that does not occur in other communities:

`FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`

I think this is likely an issue with qubqub + baas, but worth checking on before committing to see if something deeper is going on here.

tefkah:
Damn, good catch! I'll have a look

tefkah:
Yeah I think this is just a BAAS thing, even after removing all the members from the problematic collection it still displays this behavior on different deployments. On DuqDuq creating a Pub works fine, even if I add back all the members, so I'm ascribing this to QubQub being on too small of a box for this.

PR #3089, opened 2024-05-15
Update: Terms, cookie banner

gabestein:
## Issue(s) Resolved
n/a. Updates privacy policy and privacy page to reflect removal of third-party tracking and ability for communities to add GA or simple. Updates cookie banner language to reflect same. Finally, 

## Test Plan
1. Visit `/legal/privacy` and make sure new GA language is present.
2. Visit `/legal/settings` and make sure new GA language is present
3. Clear cookies (or use incognito) and visit demo. You should see a privacy policy banner and then a cookie banner. Accept both, then refresh. You should see neither.
4. Clear cookies (or use incognito) and visit jtrialerror. You should only see the privacy policy banner.


tefkah:
Seems to work as intended!

I quite like the fact that the banner lead into each other (instead of popping up after each other), neat UX.

I updated this branch just to double check this was working as intended, the previous deployment was a bit out of date, otherwise LGTM!

tefkah:
Works as intended!

PR #3092, opened 2024-05-22
fix: small type changes to get the sdk to compile master again

tefkah:
## Issue(s) Resolved
Just shifts arounds some types in order to make `@pubpub/sdk` pass `tsc` checks again. Very annoying, would liked to have been able to solve it a different way, but this is the quickest.

## Test Plan
No change, is only a type change.

### DIFF
--- a/client/components/AttributionEditor/AttributionDetailControls.tsx
+++ b/client/components/AttributionEditor/AttributionDetailControls.tsx
@@ -129,8 +129,6 @@ const AttributionDetailControls = (props: Props) => {
 				/>
 				{isShadowAttribution && (
 					<InputField
-						// @ts-expect-error ts-migrate(2322) FIXME: Type '"" | Element | undefined' is not assignable ... Remove this comment to see the full error message
-						rightElement={orcid && <Tag minimal>ORCID</Tag>}
 						placeholder="ORCID"
 						defaultValue={orcid ?? undefined}
 						onChange={(evt) => {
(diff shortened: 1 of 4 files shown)


client/components/AttributionEditor/AttributionDetailControls.tsx
```diff
@@ -129,8 +129,6 @@ const AttributionDetailControls = (props: Props) => {
 				/>
 				{isShadowAttribution && (
 					<InputField
-						// @ts-expect-error ts-migrate(2322) FIXME: Type '"" | Element | undefined' is not assignable ... Remove this comment to see the full error message
-						rightElement={orcid && <Tag minimal>ORCID</Tag>}
```
kalilsn, line 133: The other type changes make sense to me, but is this one intentional too? And do you understand what was happening here before you removed it because I do not (why pass this unexpected `rightElement` prop)?

Also just a note that this throws an error now bc of the unused `Tag` import

tefkah: You're right, I incorrectly thought this was like a vestigial prop as maybe this was changed from being an `InputGroup` to an `InputField`. 

turns out i am the one who changed that lmao haha.
I wanted to use the `error` field on the `InputField`, as `InputGroup` does not have that, in doing so unknowingly removing that `Tag`, which ofc did not throw an error bc it was being `ts-expect-error`ed, although for completely different reasons after changing the type of component.

You can see rn on prod that it does not display a `Tag`
image

whereas it "should" look like this (this branch now)

image

I reverted it to being an `InputGroup` instead and fixed the type error another way, and manually implemented some error showing logic for this component.

Errors look like this on prod
image

And now look like this on this branch
image

I guess the font size is slightly different, other than that it behaves the same!

Again, thanks for mentioning this. I should have pointed it out in the first place, but this way I actually double checked why I made that change haha

kalilsn: thank you for the detailed explanation!! gabeN bless

PR #3098, opened 2024-06-03
fix: disallow dash routes from robot.txt

tefkah:
- **GitButler Integration Commit**
- **fix: added disallow rule for /dash/* path in robots.txt**

## Issue(s) Resolved

Amazon bots were sometimes somewhat aggressively crawling `/dash` pages, some of which are not cached very frequently.

This addition tells them: "please do not"


gabestein (changes requested):
I think this needs to be rebased against main, it has some icon changes that I'm not sure are correct?

gabestein (approved):

PR #3109, opened 2024-06-06
Hotfix: Customize Cite.js user agent for Crossref polite pool

gabestein:
## Issue(s) Resolved
Appends a mailto to our user agent string to move to Crossref's "polite" pool, per https://api.crossref.org/swagger-ui/index.html and thanks to @larsgw: https://github.com/citation-js/citation-js/issues/228

## Test Plan
I need help figuring out how to test this to make sure it's working.

### DIFF
--- a/server/utils/citations/generateCitationHtml.ts
+++ b/server/utils/citations/generateCitationHtml.ts
@@ -97,6 +97,13 @@ export const generateCitationHtml = async (
 		...getCollectionLevelData(primaryCollection),
 		publisher: communityData.publishAs || '',
 	};
+
+	/* Set the user agent with a mailto to use Crossref's "polite" pool. */
+	/* https://www.crossref.org/blog/rebalancing-our-rest-api-traffic/ */
+	Cite.util.setUserAgent(
+		'PubPub/6.0 (https://pubpub.org; mailto:dev@pubpub.org) Citation.js/0.7.11 Node.js/18.7.0',
+	);
+
 	const pubCiteObject = await Cite.async({
 		...commonData,
 		// @ts-expect-error ts-migrate(2339) FIXME: Property 'containerDoi' does not exist on type '{ ... Remove this comment to see the full error message
--- a/server/utils/citations/structuredCitations.ts
+++ b/server/utils/citations/structuredCitations.ts
@@ -72,8 +72,8 @@ const getSingleStructuredCitation = async (
 	citationStyle: CitationStyleKind,
 	inlineStyle: CitationInlineStyleKind,
 ) => {
+	const fallbackValue = generateFallbackHash(structuredInput);
 	try {
-		const fallbackValue = generateFallbackHash(structuredInput);
 		const citationData = await getSingleCitationAsync(structuredInput);
 		if (citationData) {
 			const citationJson = citationData.format('data', { format: 'object' });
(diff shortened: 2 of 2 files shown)


tefkah:
While this fix does seem to add the `user-agent` correctly, it does not seem to treat our requests any better

image

tefkah:
Through some trial and error testing, i've found that 5 requests at once is the upper limit of what crossref seems to tolerate from us, with or without the `User-Agent` string

tefkah:
With this knowledge that 5 requests at a time is somewhat okay, changing this `10` here to a `5` (and restoring the default timeout to `8000` again) seems to work

https://github.com/pubpub/pubpub/blob/ee91c82c1d25c2369d716dad01e1e01e72c06f24/server/utils/citations/structuredCitations.ts#L134

I don't really know whether this will work on prod though, i think there's still some chance of timeouts occurring.

tefkah:
I implemented a fix using `p-limit` to limit the global number of requests (per dyno) to 5 here: https://github.com/pubpub/pubpub/compare/master...pubpub:pubpub:tfk/citation-hotfix-crossref-down-p-limit

This seems to work alright. By still providing a 10s timeout we make sure that roughly no more than 10 seconds is spent on fetching citations.

This means that when two people request the same page with a lot of citations, the first person will probably see the correct ones, while the second has to wait somewhat longer (for the requests to "fail", e.g. spend 10 seconds waiting in the "queue" basically) and will not see any correct citations, only the fallback.

Another alternative is to use throttling using e.g. `p-throttle`. I also explored that but it's slightly less reliable. On my machine with fast internet, the limiting works better, but maybe on prod the throttling works better, as when you spend longer waiting on a response it makes increasingly more sense to already start sending other responses.

An improvement could be to, when trying to resolve all the promises, to first check how many are pending, and if the number is 100+ or something (because someone else is fetching a lot) to just give up prematurely and render fallbacks for the page. This prevents someone waiting for 10+ seconds to see nothing.

Another issue is caching, namely if someone visits a page with the fallbacks that page will get cached, which is not ideal.
A solution could be to track this failure during the request, and add some header to exclude the page from being cached. This has as a downside that it would ofc increase the load on the server if many people are visiting uncached pages, but I think it will be fine.

gabestein:
Thanks for taking a look at this!

One q: can we push the header fix now so we can start getting on their radar? I'd like to reach out an inquire about potential whitelisting or at least increasing our limit, and I think to do that we'll need to be in the polite pool.

Otherwise, I like the p-limit solution + cache invalidation if they fail, with one caveat, which is I think we should reduce the timeout for citation-fetching, or alternately do something like detect the first 429 and fail quickly. The reason is that if we don't cache pages where citations fail, we could easily get in a scenario where hundreds of people are requesting a page that takes 10s to return by default, and overrun the dynos, right?

tefkah:
I think pushing out the header fix now is fine, even though it will not solve anything right away!

And yeah, you're right, having a ton of people requesting a page where the citations fail is probably not a good idea. Probably best to lower the amount of time then, to maybe 5 seconds? Or/and do the check to see how many requests are queued atm, which is not exactly failing when the first 429 is seen, but should work better in these situations you're describing.

tefkah (approved):
this can be merged as is

PR #3125, opened 2024-07-11
updates bulkexport script to batch exports

gabestein:
## Issue(s) Resolved
- Batches collection exports so they don't take up all the available db connections
- Excludes pubs with no releases from exports by default

## Test Plan
- Export a large collection using the CLI tool
- See that when exports are missing and need to be generate, it only tries to do ~5 at a time


isTravis:
This PR is sufficiently out of date and the merge sufficiently complex that I think we'd have better luck approaching it fresh from the current codebase. I also think it's less of an issue now that the heroku pg database isn't in our way. Let's revisit if and when it comes up again!

PR #3133, opened 2024-07-31
Dev: Debounce user query and prevent empty queries

gabestein:
## Issue(s) Resolved
- Debounces the user search query so that our server isn't spammed when people perform lots of long searches.
- Does not send user query when the query value is blank

## Test Plan
- Visit a Pub
- With the network console open, attempt to add a user by searching
- See that the `/api/user/search?q=text` query is sent only once every 300ms at maximum (this felt like the right number between ~200 and ~500).
- Delete the text, wait 300ms, and make sure no empty query (`api/user/search?q=`) is sent.
- Search by a pasted email that has a user account and make sure the user is returned after ~300ms
- Make sure you can still add a returned user to the pub
- Make sure you can add a non-existing "shadow" user to the pub
- Repeat in pub settings
- Repeat for collections in the dashboard

### DIFF
--- a/client/components/UserAutocomplete/UserAutocomplete.tsx
+++ b/client/components/UserAutocomplete/UserAutocomplete.tsx
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { Classes, MenuItem, Position } from '@blueprintjs/core';
 import { Suggest } from '@blueprintjs/select';
-
+import debounce from 'debounce';
 import { Avatar } from 'components';
 import { apiFetch } from 'client/utils/apiFetch';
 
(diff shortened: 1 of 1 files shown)


client/components/UserAutocomplete/UserAutocomplete.tsx
```diff
@@ -28,26 +28,31 @@ type Props = OwnProps & typeof defaultProps;
...
 		const { queryValue } = this.state;
-		if (queryValue !== prevState.queryValue) {
-			apiFetch(`/api/search/users?q=${queryValue}`).then((result) => {
-				const { usedUserIds } = this.props;
-				this.setState({
-					items: result.filter((item) => !usedUserIds.includes(item.id)),
-				});
-			});
+		if (queryValue !== prevState.queryValue && queryValue !== '') {
+			this.runUserQuery(queryValue);
```
gabestein, line 55: It felt like grabbing the query value and comparing it with the previous state still needed to happen in componentDidUpdate? But it was just a feeling. Can move into `runUserQuery` if it's cleaner.

PR #3227, opened 2025-02-13
update: Add hideLabel value to output HTML

isTravis:
This adds the hideLabel value onto a `data-hide-label` attribute in the toDom function for images, videos, iframes, and audio.


3mcd (approved):
👍

PR #3236, opened 2025-02-19
feat: add connection pooling

tefkah:


gabestein (approved):
Checks out for me.

I think we should only read from the read DBs. We can pay the cost. I also would like to get this up quickly given the instability we've been experiencing.

tefkah:
Great! I actually already added that in the last push, so ill merge it and test it on duqduq

PR #3249, opened 2025-03-03
fix: make scopeSummaries hooks always use main write database

### LINKED ISSUE #3245: Bug: pubs and collections can be created without scope summaries
I think this may be the first/only read replica related issue. It may also be a caching thing? But anyway...

In some conditions (not sure which), it is now possible to create a collection and/or community without a scope summary. When this happens, it makes dashboard pages crash repeatedly. This seemed to occur only after about February 25.

It does not happen in every instance, but does seem to happen somewhat regularly.

New Pubs and Collections should always have scope summaries.

tefkah:
## Issue(s) Resolved

Might resolve #3245  ?

Hypothesis: when hooks are run, there might have been enough delay between syncing the write db that we end up reading unsynced data.  Instead, we force all `scopeSummary` queries to read from the master (write) db instead. that way we should be guaranteed to have the correct data.

All major write operations which have these hooks should probably run in a transaction instead, but that's quite some more work, and this should have a similar result for now.

https://sequelize.org/docs/v7/other-topics/hooks/#hooks-and-transactions

## Test Plan

1. 🙈

### DIFF
--- a/server/scopeSummary/queries.ts
+++ b/server/scopeSummary/queries.ts
@@ -28,8 +28,10 @@ const persistScopeSummaryForId = async (
 	return newScopeSummary.id;
 };
 
+type ModelWithScopeSummary = Community | Collection | Pub;
+
 const persistScopeSummaryForModel = async (
-	model: any,
+	model: ModelWithScopeSummary,
 	summary: CreationAttributes<ScopeSummary>,
 ) => {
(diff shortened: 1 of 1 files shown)


gabestein (approved):
Code looks good. Seems to work locally. ¯\_(ツ)_/¯

tefkah:
send it

PR #3260, opened 2025-03-18
feat: community export endpoint + ui

### LINKED ISSUE #3238: Basic Legacy backup script
### Motivation

So a developer or PubPub support person can easily create archives of an individual community
So a future developer could automate the generation of an archive to allow for user-requested archives

### Requirements

- A new private GH repo with a GitHub Action that allows user to specify:
   - Community URL
   - Email address (or gets user's email address from GitHub)
- When run, runs Travis' [getLegacyPubs script](https://github.com/knowledgefutures/2024-11-ArcadiaMigrationScripts/tree/main/src/getLegacyPubs)
- Modifies the script to upload the resulting files to an S3 bucket at (`archive.pubpub.org/communityurl`) - does not make any other modifications to the script, we're gonna do that all later
- When finished, sends an email to the specified email address

### Acceptance Criteria
- Uses environment variables to specify a superuser login to run the API script

### LINKED ISSUE #3290: Feature flag for Legacy backups

tefkah:
## Issue(s) Resolved

Resolves #3238 
Resolves #3290 

## Test Plan

As a superadmin (or regular admin), go to some community (like the test community). Go to settings, scroll down, see something like this

image

(the infinity is for superadmins, in the future when we open this to admins they have a limit of 5 exports per day)

Try to do an export. Click the link, see that it's a json array of the pubs in the community.

Bonus:
- try to do this for a huge community (baas) and see if it completes in a reasonable amount of time. currently we are fetching 100 pubs at a time, so this might take wayyyyy too long.
- try to do an export through the api at `/api/community/archive`

### Notes/Context/Gotchas

I replaced `jsonpath` with `jsonpath-plus` bc it was giving me errors locally. However i have not at all tested whether these two things behave the same, because they were only used in quite niche applications. if anyone knows how to tests these applications (see my comments) please do so


tools/figurelist.js
tefkah, line 1: jsonpath->jsonpath-plus swap

workers/tasks/import/bulk/directives/pub.ts
tefkah, line 1: jsonpath->jsonpath-plus swap

workers/tasks/import/transforms/extractEndnotes.ts
tefkah, line 1: jsonpath->jsonpath-plus swap

workers/tasks/archive.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const getTmpDirectoryPath = async () => {
+	const tmpDirPossiblySymlinked = await tmp.dir();
+	const tmpDir = fs.opendirSync(fs.realpathSync(tmpDirPossiblySymlinked.path));
+	return tmpDir.path;
+};
+
+export const archiveTask = async ({ communityId, key }: { communityId: string; key: string }) => {
+	let offset = 0;
+	const limit = 200;
```
tefkah, line 18: maybe this limit should be higher

workers/tasks/archive.ts
```diff
@@ -0,0 +1,76 @@
...
+
+		if (pubs.length > 0) {
+			await fs.promises.appendFile(tmpFilePath, stringifiedPubs);
+		}
+
+		offset += limit;
+		if (pubs.length < limit) {
+			break;
+		}
+	}
```
tefkah, line 68: basically i don't want to keep this thing in memory, but i want to create a single array in the end. if you have a better idea on how to do this, enlighten me! this should probably use a `writeStream` instead of repeated `appendFile` calls though

client/containers/DashboardSettings/CommunitySettings/ExportCommunityDataButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+					Download
+				</a>
+			)}
+			<Button
+				disabled={isExporting || disabled}
+				onClick={() => {
+					setIsExporting(true);
+					apiFetch('/api/communities/archive', {
+						method: 'POST',
+						body: JSON.stringify({ dontWait: true }),
```
kalilsn, line 23: immediately returning a URL here is kind of an awkward experience, since there's no way for a user to know when the link will work (you get an s3 access denied error if you click before the download finishes). since we're already using the `pingTask` helper here, couldn't we keep the button disabled until the download completes?

kalilsn:
Actually this runs without error on baas (about 1 min for a 330mb file) after I increased the worker instance size on the review instance to match prod (2.5gb memory). but still maybe something to think about/test!

workers/tasks/archive.ts
kalilsn: i'm realizing now that the disk filling up may not be a real concern, since maybe heroku runs the worker containers on demand? i don't actually remember how that works

tefkah: great suggestions, i will do that! maybe i can directly pipe it to s3, not sure what the memory implications of that are

3mcd (approved):

PR #3315, opened 2025-07-10
feat(archive): fix static url rewrites on pages other than home

3mcd:
## Issue(s) Resolved

N/A

## Test Plan

- Create a community archive.
- `npx serve` it.
- Confirm that links in JS-rendered contexts (collection dropdown, releases dropdown) direct you to the locally served page rather than the production pubpub community.
- Confirm that links in server-rendered (static HTML) contexts (pub links, page links) do the same.


tefkah (approved):
works great!

PR #3325, opened 2025-07-21
feat: add latest html in export.json

### LINKED ISSUE #3292: Add latest HTML to JSON export
See https://github.com/pubpub/pubpub/pull/3291 for a bad version.

Having the latest HTML, whether from a draft or the latest releases, makes it easy to write scripts that convert the json to importable JSON or XML for e.g. ghost and wordpress.

Other things that would be good in addition to just taking the exporting the straight HTML doc:
- Include linked footnotes/citations, similar to how we do it for pagedjs exports
- Strategy for downloading/rewriting figure, file, and other asset links -- probably as part of a separate CLI tool

tefkah:
## Issue(s) Resolved

Resolves #3292 

## Test Plan

Run export on not too large community (maybe demo)

See that `html` and `abstract` is included in the `export.json` for pubs with a release. the html should correspond to the latest release.

### Notes/Context/Gotchas

This also adds support for downloading image files in `style="background-image:...` `div`s, and the `ogImage`. This does slow down things a bit, but it's necessary. If needed i can spin it off in a separate pr.

Originally i was worried that that adding manual html exports would slow down the archiver a ton, especially when creating the citations and such. The citations do slow down the archiver but it's not that bad.

I thought initially this may lead to errors with S3, but it's something else related to this (https://github.com/nodejs/undici/issues/3492). Not exactly sure why adding the notes triggers this, may be some resolving of dois issue?

In any case, this might be fixed by this commit (https://github.com/nodejs/node/commit/05ad94789916563664acb666a7ac367fd8b2cc8b), which landed in node 24, so i set that as the new default node version.

This did cause some issues with the dependency overrides for the ts-rest/packages (as a reminder, i forked ts-rest to add support for file uploads using node 18, which we were using at the time). I think i sort of fixed that, but it's admittedly a bit jank. It might be worth just using stable ts-rest again as we don't need to use node 18 anymore, but that also requires changing the sdk which im not feeling rn.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -326,7 +326,7 @@
 				"webpack-manifest-plugin": "^2.0.4"
 			},
 			"engines": {
-				"node": "22.14.0"
+				"node": "24.x"
 			}
 		},
 		"eslint": {
@@ -56853,8 +56853,7 @@
 			}
 		},
 		"domhandler": {
-			"version": "5.0.3",
-			"resolved": "https://registry.npmjs.org/domhandler/-/domhandler-5.0.3.tgz",
+			"version": "https://registry.npmjs.org/domhandler/-/domhandler-5.0.3.tgz",
 			"integrity": "sha512-cgwlv/1iFQiFnU96XXgROh8xTeetsnJiDsTc7TYCLFd9+/WNkIqPTxiM/8pSd8VIrhXGTf1Ny1q1hquVqDJB5w==",
 			"requires": {
 				"domelementtype": "^2.3.0"
--- a/package.json
+++ b/package.json
@@ -365,10 +365,10 @@
 	},
 	"overrides": {
 		"@ts-rest/express": {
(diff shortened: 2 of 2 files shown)


workers/tasks/archive/siteDownloaderTransform.ts
```diff
@@ -111,6 +144,34 @@ export class SiteDownloaderTransform extends Transform {
...
+					/\./g,
+					'_',
+				)}`;
+				const assetPath = `${assetDir}${backgroundImageUrl.pathname}`;
+
+				console.log(`Pushing background image: ${backgroundImageUrl.href}`);
+				this.#pushAsset(backgroundImageUrl, assetPath);
+
+				break;
+			}
```
tefkah, line 174: for the `background-image:...` divs

3mcd (approved):

PR #3341, opened 2025-07-30
fix: rewrite urls with proper string delimiters

3mcd:


3mcd:
@tefkah I'm going to go ahead and merge this. The gist of it is the URLs were being rewritten to `/assets/...` from `"https://resize-v3.pubpub.org/..."`, so the quotes were removed, rendering the JSON invalid. The pages still loaded correctly because of the initial HTML snapshot, but all JS functionality was broken.

PR #3392, opened 2025-11-18
feat: allow users to change their own email (and password)

### LINKED ISSUE #433: Allow users to change email from their profile
**Who requested**
Catherine/everyone

**Stories**
- As a USER, I want to be able to change my email from my profile, so that I don't lose my account or need to create a new one if my email changes.

**Implementation Notes & Context**
We should probably force users to re-authenticate before doing this, and we'll need to send a new confirmation email.

**Technical Specification**

_To be completed by the Engineering team/broken into sub-cards as needed."_

**Mockups or reference images**

tefkah:
## Issue(s) Resolved

Resolves #433 

## High level overview

Adds a new section to `/legal/settings` to change your email and password.
Renamed that from "Privacy Settings" to "Privacy & Account" bc it was already doing more than privacy.

You can change your email by submitting your password and new email. Then an email will be sent with a link. As soon as you click that link it will have changed your email.
Easy as.

You can also change your password there. Technically this was already possible by clicking "Forgot password", but that is annoying UX.

The new API routes are maybe annoyingly at `/api/account/...` instead of `/api/user/...`, bc i ran into some trouble using `ts-rest` on paths that are similiar to "legacy" routes, which are defineed as just straight up Express routes.

Note: we can't really verify whether the user's new password is >8 chars on the server bc we are sending up a hashed thingy. I think it's fine, we do so on the client.

## Test Plan

1. Try to change your email address. Should work
2. Try to reuse that link to change it again. Should not work.

Look through the tests, is there an obvious case that's missed?

### DIFF
--- a/client/components/AccountSecuritySettings/AccountSecuritySettings.tsx
+++ b/client/components/AccountSecuritySettings/AccountSecuritySettings.tsx
@@ -7,7 +7,7 @@ import SHA3 from 'crypto-js/sha3';
 import { apiFetch } from 'client/utils/apiFetch';
 import { InputField } from 'components';
 
-require('./accountSecuritySettings.scss');
+import './accountSecuritySettings.scss';
 
 const stripHTTPError = (error: string) => {
 	return error.replace(/^HTTP Error \d+: /, '');
--- a/client/containers/EmailChange/EmailChange.tsx
+++ b/client/containers/EmailChange/EmailChange.tsx
@@ -5,7 +5,7 @@ import { AnchorButton, Spinner } from '@blueprintjs/core';
 import { apiFetch } from 'client/utils/apiFetch';
 import { GridWrapper } from 'components';
 
-require('./emailChange.scss');
+import './emailChange.scss';
 
 type Props = {
 	emailChangeData: {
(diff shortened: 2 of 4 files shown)


server/emailChangeToken/model.ts
tefkah: I thought creating a new table would be the best way to do this. Kinda annoying to have to add another table, but I couldn't really figure out another way to do this, other than reusing the (kinda bad) sending the user something like this, where `User.resetHash` is what we use for the password reset, and `encodedEmail` is the string of the new email encoded with a secret env var.

`https://<domain>/<${User.resetHash}>?email=${encodedEmail}`

server/routes/legal.tsx
```diff
@@ -15,34 +15,52 @@ app.get('/privacy/settings', (_, res) => res.redirect('/legal/settings'));
...
+		return renderToNodeStream(
+			res,
+			// @ts-expect-error ts-migrate(2322) FIXME: Type '{ children: Element; chunkName: string; init... Remove this comment to see the full error message
+			<Html
+				chunkName="Legal"
+				initialData={initialData}
+				viewData={{
+					integrations,
+					userNotificationPreferences,
+					userEmail: isSettingsTab ? req.user?.email : undefined,
```
tefkah: I want to show what the user's current email is on the page, hence why im passing it through like this

3mcd (approved):
Nice job. Confirmation email works well. The token approach is sound too!

PR #3413, opened 2026-01-15
fix: add validation for depositing crossref reviews

tefkah:
## Issue(s) Resolved

- for reviews
	- checks whether isReviewOf target resolves to a DOI
	- checks whether that DOI resolves at all
	- checks wheher that DOI is a crossref DOI
	- does not do the double deposit pattern
	- if any of those fail it will not allow you to deposit, and will tell you whats wrong
- allows you to manually add DOIs to external connections, which can fix the first check above
- fixes the problem where it would not treat https://doi.org/10.xxx/xxx external pubs the same as 10.xxx/xxxx dois
- checks when depositing whether all the dois on Pubs resolve properly, and will tell you when that’s not the case. this  makes it easier to fix deposits in a bad state

all of these are kinda not great fixes, i think a better solution would be to actually detect whether a deposit has succeeded or not, and not assign a DOI to a Pub until it has, but that’s maybe too big of a lift

### DIFF
--- a/server/doi/api.ts
+++ b/server/doi/api.ts
@@ -79,7 +79,6 @@ router.post(
 			// validate relationships before depositing to prevent cascading failures
 			if (pubId) {
 				const issues = await validatePubRelationshipsForDeposit(pubId);
-				console.log('issues', issues);
 				// all issue types are blocking for deposits
 				if (issues.length > 0) {
 					const errorMessages = issues.map((issue) => {


client/components/PubEdge/PubEdgeEditor.tsx
```diff
@@ -64,6 +73,45 @@ const PubEdgeEditor = (props: PubEdgeEditorProps) => {
...
+				// @ts-expect-error ts-migrate(2769) FIXME: No overload matches this call.
+				tabIndex="0"
+				onKeyDown={(evt) => evt.key === 'Enter' && setDoiOpen(true)}
+				onClick={() => setDoiOpen(true)}
+			>
+				Add DOI
+			</RKButton>
+		);
+	};
+
```
tefkah: this adds a dedicated DOI button to the connection configuration thingy. that way you can manually fix links that do not have dois

PR #3471, opened 2026-02-10
feat: only allow iframes to render html

tefkah:
## Issue(s) Resolved

Preventing scammers from using PubPub to host scams

## Test Plan

1. Try and upload a HTML document.
2. Go to the link. See that it forces a download
3. Try and put the link in an iframe
4. See that the iframe renders the HTML properly
5. Replace `https://assets.pubpub.org` with `https://s3.amazonaws.com/assets.pubpub.org`
6. Go to link, see that it still forces a download. Linking this version in an iframe doesn't work, as it's dependent on a Cloudfront function

### Notes/Context/Gotchas

this is the relevant cloudfront function

```js
function handler(event) {
	var response = event.response;
	var headers = response.headers;
	var request = event.request;

	var isHtml = headers['content-type'] && headers['content-type'].value === 'text/html' || request.uri.split('?')[0].endsWith('.html');

	if(!isHtml){
		return response;
	}

	var qs = request.querystring 
	var embedParam = qs.embed;

	var isEmbed = embedParam && embedParam.value === "1";
	var reqHeader = request.headers['sec-fetch-dest'];
	var isFrameRequest = reqHeader ? reqHeader.value === 'iframe' : false

	if (isHtml && isEmbed && isFrameRequest) {
		headers["content-disposition"] = { value: "inline" };
	} else {
		headers["content-disposition"] = { value: "attachment" };
	}

	return response;
}
```

### DIFF
--- a/server/spamTag/uploadScamKeywords.ts
+++ b/server/spamTag/uploadScamKeywords.ts
@@ -13,7 +13,6 @@ export const uploadScamKeywords = [
 ] as const;
 
 export const isSuspiciousUploadKey = (keyOrFilename: string): boolean => {
-	console.log('keyOrFilename', keyOrFilename);
 	const lower = keyOrFilename.toLowerCase();
 	return uploadScamKeywords.some((kw) => lower.includes(kw));
 };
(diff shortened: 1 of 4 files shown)


scripts/remove-scam-asset.sh
tefkah: this is just a useful script to quickly remove a scam asset and move it to a dedicated bucket

PR #3515, opened 2026-03-03
fix: improve dev setup time and container size

tefkah:
## Issue(s) Resolved

I keep running out of space on my Mac bc of the anonymous `/app/node_modules` volume.

I simplified the dev setup to just mount the main folder, that way we don't need to run a `pnpm install` in the dev container itself.

This takes the dev startup time from a couple of minutes to like 10 seconds for me, as well as saving a lot of space.

## Test Plan

1. Run `pnpm dev`
2. Observe normal dev env

### Notes/Context/Gotchas

I added the `@esbuuild/linux-arm64` devDep bc otherwise `esbuild-loader` complains that you have the MacOS version of `esbuild` installed when running in the container.
This assumes all devs run an ARM machine. We could also add the `@esbuild/linux-amd64` devDep if we a have a dev that doesn't have an ARM device.

We could also _only_ add `@esbuild/linux-amd64` and add `platform: linux/amd64` to the `docker-compose.dev.yml`, to run the dev container in a more similar way to the prod one, but that would probably be more resource intensive/slower to run for us Mac using devs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ ENV PORT=3000
 
 EXPOSE 3000
 
-CMD [ "pnpm", "start"]
+CMD ["pnpm", "start"]
 
 # ---- Production runtime stage ----
 FROM base AS runner


tefkah:
> Am I understanding the diff correctly? It seems like we already were just mounting the main app folder. So the dev pnpm install was redundant and unnecessary? Looks like everything is a removal on that front, rather than any new mounts. 
> 
> What a time saver!

yes that's correct! I think that was added previously bc of this specific incompatibility between some node_modules, but this should work!

PR #3518, opened 2026-03-04
feat: disallow links in comments

tefkah:
## Issue(s) Resolved

To be revisited after running #3521 

s p a m

## What

This PR disables the ability to add links to comments, by removing the Link button from the formatting bar, disabling the Cmd+K shortcut, and disabling the https:// regex.

It also makes it so links in previous comments are now no longer present in the DOM on load: they only appear once you click on them in a popover, where they are escaped.

image

The latter is maybe a bit dramatic, we could just show the actual link there as well.

### DIFF
--- a/client/components/Editor/plugins/inputRules.ts
+++ b/client/components/Editor/plugins/inputRules.ts
@@ -155,7 +155,8 @@ function linkRule(markType: MarkType) {
 // code blocks, and heading.
 export default (schema, options?) => {
 	const rules = smartQuotes.concat(ellipsis, emDash);
-	if (schema.marks.link && !options?.disableLinkCreation) rules.unshift(linkRule(schema.marks.link));
+	if (schema.marks.link && !options?.disableLinkCreation)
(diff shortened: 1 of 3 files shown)


gabestein:
Yeah, we're not going to eliminate all spam. You've put in place tools that should both drastically reduce it going forward and buy us time to help us identify and eliminate past spam, too. Let's at least try letting that work before. POSIWID is a good guide to identifying issues but if, in trying to solve the problems, we allow the descriptive to override the normative, we are ceding our roadmap to spammers.

On the specifics: I think we should run the script in #3521. If we have some false positives, that is imo preferable to turning off functionality.

tefkah:
> On the specifics: I think we should run the script in https://github.com/knowledgefutures/pubpub/pull/3521. If we have some false positives, that is imo preferable to turning off functionality.

alright! ill turn this PR into a draft then in case we want to try out a more limited version of this in the future, lets see in a month or so (and close it if not needed)?

PR #3590, opened 2026-04-18
chore: review BundleAnalyzer output and clean

isTravis:
Reduces initial page load by ~958 KB through dead code removal and removing an accidental storybook data inclusion. No runtime logic changes — just a dead import removal and a build-time guard fix (which also fixes a bug in DOI validation links).

### 1. Remove storybook mock data from production bundle (-640 KB from main)

`Doi.tsx` had a stale `import { pubData } from 'utils/storybook/data'` that pulled 640 KB of mock fixtures (fake community, pub, discussion, overview data) into the production `main.js` chunk. The import was dead except for two `<a href>` links in `renderValidationWarnings()` that were accidentally using the storybook mock's `pubData.slug` instead of `this.props.pubData.slug` — a bug causing incorrect links on the DOI validation warnings UI. Fixed by removing the import and destructuring `pubData` from `this.props`.

### 2. Exclude prosemirror-dev-tools from production build (-166 KB from vendor, -152 KB jsondiffpatch)

`useEditorView.ts` guarded the `import('prosemirror-dev-tools')` call with `isDevelopment()`, a runtime hostname check. Webpack cannot dead-code-eliminate runtime conditions, so the dynamic import chunk (and all of prosemirror-dev-tools + jsondiffpatch) was still bundled in production. Changed to `process.env.NODE_ENV !== 'production'`, which the existing `DefinePlugin` resolves to `false` at build time, allowing Terser to strip the entire block and its dependencies.

## Results (dev build, unminified)

| Metric | Before | After | Change |
|---|---|---|---|
| vendor.js | 17,211 KB | ~16,893 KB | -318 KB |
| main.js | 5,047 KB | ~4,407 KB | -640 KB |
| main.css | 899 KB | 867 KB | -32 KB |
| **Initial page load total** | **23,157 KB** | **~22,167 KB** | **~-990 KB (-4.3%)** |

### Breakdown of savings by source

| Source | Savings | Type |
|---|---|---|
| Storybook mock data removed | 640 KB | Eliminated from main |
| prosemirror-dev-tools + jsondiffpatch | 318 KB | Eliminated from vendor (prod only) |

## Files changed

- `client/containers/DashboardSettings/PubSettings/Doi.tsx` — remove storybook import, fix pubData bug
- `client/components/Editor/hooks/useEditorView.ts` — `isDevelopment()` → `process.env.NODE_ENV !== 'production'`

## Risk

**Very low.** Both changes only remove dead code paths:
- The storybook import was unused except for a bug (wrong slug in validation links, now fixed)
- The prosemirror-dev-tools guard change is semantically identical — it was already only loaded in dev; this just makes it eliminable at build time


tefkah (approved):
gah, how silly of me! great catch!

tefkah:
i also have a branch that had a go at minifying the bundle at some point, mostly wrt lazy loading more of the languages in the editor, can see if i can still clean that up!

tefkah:
i think the eventual solution should ofc be that we have route based bundles tho. I'm not sure how we could get there without reworking our entire tooling, but it's silly that stuff only happening on the doi page is affecting every other page

PR #3603, opened 2026-04-23
feat: Show communities in user's account dropdown and other header improvements

isTravis:
Three changes to the global header controls:

1. **"Your Communities" in user menu** — The user dropdown now has a "Your Communities" section lazily fetches and lists all communities the user is affiliated with. Adds a new `GET /api/users/communities` endpoint that queries across direct memberships, pub/collection memberships, and attributions. Finally...

1. **Supadmin link** — Superadmins now see a "Supadmin" link on the far left of the header controls, linking to `/superadmin`.

2. **Dev community switcher repositioned** — On localhost/dev, the community switcher now appears in a consistent position as we moved the honeypot/captcha visuals to be absolutely positioned below Create Pub button (which they are a part of). This will prevent the community switcher from jumping positions soon afer page load when the captcha pops in.

Screenshot 2026-04-22 at 11 03 26 PM


tefkah:
haha great, thanks for moving the captcha!

